Issue #70, opened 2014-09-12
S_Cache is a terrible name for the collection of Nova servers


glyph:
We should call it `ServerCollection` or something, and we should have a `Server` instance so that logic around manipulating states is centralized somewhere.

glyph:
Fixed via #148

PR #79, opened 2014-09-19
WIP: starting maas plugin

glyph:


mimic/canned_responses/maas_json_home.py
```diff
@@ -0,0 +1,1147 @@
+"""
+Canned response for monitoring json home
+"""
+
+
+def json_home(url):
+    """
+    Canned response for the jsno_home call
```
glyph: jsno

mimic/rest/maas_api.py
```diff
@@ -0,0 +1,960 @@
...
+                                                                  code (E.164 format)'}]}]
+        self.suppressions_list = []
+
+
+def createEntity(params):
+    """
+    Returns a dictionary representing an entity
+    """
+    params = collections.defaultdict(lambda: '', params)
+    newentity = {}
```
glyph: Minor style note: when you are building a dictionary like this, it's best to build it as a single value (as you do above) rather than one value at a time.  If you want to copy most of the stuff from `params` and make a few small tweaks, you can use `dict.update` to copy multiple values in one shot.

mimic/rest/maas_api.py
```diff
@@ -0,0 +1,960 @@
...
+    newentity['label'] = params[u'label'].encode("ascii")
+    newentity['id'] = 'e' + random_hex_generator(4)
+    newentity['agent_id'] = params['agent_id'] or random_hex_generator(12)
+    newentity['created_at'] = time.time()
+    newentity['updated_at'] = time.time()
```
glyph: All time calculations within Mimic ought to be using the `Clock` object associated with the `MimicCore`.  If you use `time.time()`, MaaS's idea of time will be desynchronized from the rest of Mimic's idea of time.

mimic/rest/maas_api.py
```diff
@@ -0,0 +1,960 @@
...
+        if 'encode' in dir(params[k]):
+            params[k] = params[k].encode('ascii')
+    params['id'] = 'c' + random_hex_generator(4)
+    params['collectors'] = []
+    for q in range(3):
+        params['collectors'].append('co' + random_hex_generator(3))
+    params['confd_hash'] = None
+    params['confd_name'] = None
+    params['created_at'] = time.time()
+    params['updated_at'] = time.time()
```
glyph: See above about `time.time`

mimic/rest/maas_api.py
glyph: Similarly, any time you want two times to be the same, call `clock.seconds()` *once*; if you call it twice in rapid succession like this, 99.999% of the time the values will be the same, and then that one very very rare case where they're not exactly the same will cause confusion.

mimic/rest/maas_api.py
```diff
@@ -0,0 +1,960 @@
...
+        for a in alarms:
+            if a['entity_id'] == entity_id:
+                del alarms[alarms.index(a)]
+        request.setResponseCode(204)
+        request.setHeader('content-type', 'text/plain')
+
+    @app.route('/v1.0/<string:tenant_id>/entities/<string:entity_id>/checks', methods=['POST'])
+    def create_check(self, request, tenant_id, entity_id):
+        """
+        Create a check
```
glyph: any docstring which is just the name of the method with an article inserted probably needs a little work :).  Perhaps a link to the docs, or a description of the `:param:`s, or an explanation of what a "check" represents, or how mimic stores it... something that gives an explanation of the method for future maintainers.

mimic/rest/maas_api.py
```diff
@@ -0,0 +1,960 @@
...
+    newentity['agent_id'] = params['agent_id'] or random_hex_generator(12)
+    newentity['created_at'] = time.time()
+    newentity['updated_at'] = time.time()
+    newentity['managed'] = params['managed']
+    newentity['metadata'] = params['metadata']
+    newentity['ip_addresses'] = params['ip_addresses'] or {}
+    return newentity
+
+
+def createCheck(params):
```
glyph: These new functions should all follow the coding standard, but I think we can do that as a separate issue.

mimic/test/test_maas.py
```diff
@@ -0,0 +1,677 @@
...
+                'nt_id': nt_id, 'np_id': np_id, 'sp_id': sp_id}
+
+    def setUp(self):
+        """
+        Setup MaasApi helper object & uri 'n stuff
+        """
+        helper = APIMockHelper(self, [MaasApi(["ORD"])])
+        self.root = helper.root
+        self.uri = helper.uri
+        entity_id = None
```
glyph: All these `None`s are unnecessary - you're immediately re-assigning all these variables, so you can delete all these lines and everything will be the same :).

mimic/test/test_maas.py
```diff
@@ -0,0 +1,677 @@
...
+        nt_id = None
+        np_id = None
+        sp_id = None
+        entity_id = self.getXobjectIDfromResponse(self.createEntity('ItsAnEntity'))
+        check_id = self.getXobjectIDfromResponse(self.createCheck('ItsAcheck', entity_id))
+        alarm_id = self.getXobjectIDfromResponse(self.createAlarm('ItsAnAlarm', entity_id, check_id))
+        nt_id = self.getXobjectIDfromResponse(self.createNotification('ItsANotificationTarget'))
+        np_id = self.getXobjectIDfromResponse(self.createNotificationPlan('ItsANotificationPlan'))
+        sp_id = self.getXobjectIDfromResponse(self.createSuppression('ItsASuppression'))
+        self.assertNotEquals(None, entity_id)
```
glyph: When comparing to `None`, "is" is the preferred comparison operator, so `assertIsNot` or `assertNotIdentical` would be better.  ("==" gives the object being compared to the opportunity to run some code in its `__eq__` method, whereas `is` just asks the python runtime, so it's faster and has fewer types of errors which can occur).

Given the types of objects being compared here, it doesn't really matter semantically, but it's better style that way.

glyph: I'd also assert more about the returned value here: "obviously" it shouldn't be `None`, but what *should* it be?  If it's randomly generated, what format should it be in?  If it's supposed to be an instance of a particular type, you can use `assertIsInstance`?  If it's supposed to be a particular length, you can assert about its `len`.

mimic/rest/maas_api.py
```diff
@@ -0,0 +1,960 @@
...
+
+    def resource_for_region(self, region, uri_prefix, session_store):
+        """
+        Get an : obj: `twisted.web.iweb.IResource` for the given URI prefix;
+        implement : obj: `IAPIMock`.
+        """
+        return MaasMock(self, uri_prefix, session_store, region).app.resource()
+
+
+class M_Cache(dict):
```
glyph: This is a copy of a really unfortunate anti-pattern.  Please file an issue in the vein of #70.

glyph:
The build failed because when you resolved the conflict, you just went with the list that included the rcv3 plugin; please merge the lists together so the build succeeds :)

glyph:
Although this branch is really large, it's already in use by multiple teams.  That means that it is, for Mimic's description of "production", already in production.  There's not much point to keeping changes off `master` that have already been deployed, because the point of keeping `master` in good shape is that you should be able to deploy it at any time.

Once the tests are passing, we should just land it in more or less the shape it's in, rather than splitting it up.  I have given the code a quick skim of a review, the tests are passing, the lint is passing, and the problems I was able to spot are good opportunities for learning but not hard blockers for integrating this.  They should all be fixed but they can all be fixed in subsequent issues.

However, we should have a discussion at some point soon about how to make it so that other PRs in this genre - adding plugins - can be done incrementally and not get to this "big bang" state where they need to be landed all at once.  Perhaps this branch can serve as a "good" example in that some other PRs were merged into it - the code review can happen at the time that individual small features are being added to a nascent plugin, and the plugin branch itself can be committed without review (because all its pieces were reviewed as they were being created).  Perhaps we should just be more comfortable with incomplete plugins on `master`, and we should add a feature flag so that users won't see them if they're known to be too functionally incomplete to actually get used.

Issue #53, opened 2014-08-08
request failures from compute (nova error injection)


glyph:
One of the benefits of Mimic is that it can let you do integration testing against an API that can deterministically produce specific types of failures.

As a first step towards this, we should have a way to ask the compute endpoint to fail in a particular way.

The API endpoint should allow the caller to express the following:

“When a request comes in to create a server who matches a given matcher, use a reaction from the following list of reactions.”

“use” can mean one of two things; either dequeue the reaction from the matched list of reactions, or cycle that list, popping the reaction off the front and then pushing it onto the back.

A matcher is:

  - a regular expression that matches a name
  - a regular expression matches a given server metadata field

A reaction is:

  - (optionally) a response for the create request, defaulting to “normal successful response” if unspecified
  - (optionally) a list of responses for all subsequent “get” requests for that server, defaulting to “normal get response” if unspecified.  This list also includes a flag similar to the “use” flag above - this list of get responses can either be exhausted as if it is a queue or cycled repeatedly by appending each response to the end of the response list each time it's returned.
  - (optionally) whether to create the state for the server at all (i.e. did the “compute backend” succeed, should a server even show up on subsequent “get” requests), defaulting to “yes”

A response is:

  - all of these elements are optional
  - an HTTP status code
  - a set of HTTP headers to add or remove
  - an HTTP connection behavior, including:
    - leave the connection open for a timeout, then close it
    - leave the connection open for a timeout before responding
  - an HTTP response body, specified either as
    - a set of JSON fields to add and remove (dict.update perhaps? maybe something more expressive so we can add to a list or remove from a dictionary that is nested?)
    - a JSON object that replaces the full response
    - literal base64-encoded bytes, to simulate really wacky errors, like an HTML error page due to a reverse proxy being down somewhere, or gzipped content where you don't expect it

The mimic control plane should show up in the service catalog as a separate resource so that test clients can properly detect it (and not issue test-control-plane requests to any real application resources).

The schema described above should also be used to enhance the metadata-based individual responses to servers, which are currently limited in the forms of anomalous behavior that they can express.  Rather than including a list of reactions, however, an individual create-server request should include a “mimic” field in its metadata pointing at a single reaction object.  That reaction object should include all the same JSON objects described above in the “reaction” section.

glyph:
How about this, instead.

We have a list, identified by name, of "things which might be reacted to".  Those names could be things like "create server", "delete server", "list servers", "get server details".

[...]

PR #92, opened 2014-10-17
WIP: Nova error injection

ddunlop:


glyph:
I think that this branch is probably a dead end for this feature.  #148 is a step in a better direction for #53 .

PR #103, opened 2014-11-09
Py2app 88

derwolfe:
Pushing the first stages of the commit. As of right now, py2app will build the application, but when run, it will fail with the following error:

> 11/9/14 8:36:46.890 AM mimic[39596]: ImportError: No module named syslog
> 11/9/14 8:45:41.470 AM mimic[40056]:   File "twisted/python/syslog.pyc", line 11, in <module>

I'm pretty sure this is related to how my start-app.py script is working. 

Also this has **no tests**, which will obviously need to change.


lekhajee:
Reply from travis support:

Hi Lekha,

Thanks for getting in touch. While I would love to enable multi-os for you, we're currently unable to accommodate projects with "language: python" (see: https://github.com/travis-ci/travis-ci/issues/2312) If this is something that interests you, I would strongly recommend reaching out to dstufft, not least because you're both Rackers ;-) We look forward to offering multi-os to Python projects!

Have a great day,

--
Daniel Buch
support@travis-ci.com

derwolfe:
@lekhajee Thanks for getting in touch with them. I'll see how the pyca/cryptography project is handling this, as it looks like they have multi OS support. I believe  that they are doing a bit of the virtualenv setup themselves and setting the "travis language" parameter to C.

lekhajee:
@derwolfe: ah ok, interesting. Thank you for looking at it.

derwolfe:
I've emailed travis about setting up multi-os support for my fork, with the intention of setting up the project as a C project. 

If they agree, I'll start experimenting with build configurations.

derwolfe:
I've made a bit of headway getting travis to build the application in my fork. The issue I'm running into is that the build for pyobjc is taking longer than 50 minutes. Travis times out builds that take longer than 50 minutes.

derwolfe:
Stripping the pyobjc dependencies down has gotten the application to build on travis. But, there are still some errors in how I've configured travis. The build also seems to be getting tripped up by this, I believe this issue is popping up as well: https://twistedmatrix.com/trac/ticket/3985

glyph:
Always glad to have an excuse to get a Twisted bug fixed :).  This implies that you didn't build the dropin.cache file before creating a zip file with twisted plugins in it, though.  If you do `list(getPlugins(...))` at the relevant point in the build process (maybe py2app has a hook for this when it builds the zip file?) that should address it for you.

derwolfe:
@glyph - thanks for the tip! I'll try to get this fixed up later today.

derwolfe:
It may sound odd, but the bundle job has passed on travis once, earlier today. But, I'm having difficulty getting it to rebuild again. However, the tox script is passing when run locally.

The `dropin.cache` file is being built by setup-mac.py. If this step is removed from the build process, the error throw on travis (shown below) is produced.

```python-traceback
[ERROR]
Traceback (most recent call last):
  File "twisted/plugin.pyc", line 181, in getCache
    
exceptions.AttributeError: 'ZipPath' object has no attribute 'setContent'
mimic.test.test_core.CoreBuildingTests.test_from_plugin_includes_all_plugins
mimic.test.test_tap.TapTests.test_makeService
mimic.test.test_tap.TapTests.test_realtime
```

derwolfe:
I have rebased in current changes on master and have broken several things.

derwolfe:
Since I have set travis up for multi-os builds, you may see the current build status here: https://travis-ci.org/derwolfe/mimic.

If anyone would like to try to pull the branch down and build, you may do so by running `tox -e bundle`. The build process should take about 10 to 15 minutes as I've pared down the amount of pyobjc code that needs to be built.

derwolfe:
@glyph - I've not been able to reproduce the travis failure on my laptop, either running make or just `tox -e bundle`. If you have the time, would you try to build? 

It is possible that I'm somehow not seeing a dropin.cache file that my system is producing, but, I'm deleting each of these before the build occurs. In short, I'm feeling stumped.

glyph:
@derwolfe - the problematic line in the travis build is this:

`gcc (Ubuntu/Linaro 4.6.3-1ubuntu5) 4.6.3`

Thanks @dreid for that diagnosis :)

derwolfe:
@glyph - thanks both to you and @dreid for taking a look. But, I'm not sure we're talking about the same error. The issue that I'm seemingly unable to get solved is still the build for the application bundle: https://travis-ci.org/derwolfe/mimic/jobs/46923835.

derwolfe:
Please disregard my previous question. I have a had build pass on travis. Once, I confirm that all is well, i'll clean it up merge it into this PR.

glyph:
@derwolfe - so you've figured out the dropin.cache issue then?

derwolfe:
Yes. It looks like it was tox doing things in a different Dir, so the cache
file wasn't being found. Once I switched back to doing a few tasks in
run.sh, things started working.
Am 14.01.2015 19:38 schrieb "Glyph" <notifications@github.com>:

glyph:
Fantastic. I *really* look forward to approving this when that little red :x:  turns into a little green :white_check_mark:

derwolfe:
The tests are now passing for https://travis-ci.org/derwolfe/mimic. 

In order to get travis to run the tests correctly in the https://travis-ci.org/rackerlabs/mimic/builds repo, travis will have to enable multi-os support for you.

Also - my apologies for the enormously ugly commit log. Thanks!

glyph:
Oh! I thought that @lekhajee already did this.  I think maybe I misunderstood when you enabled multi-OS for your own repo and thought it was done?

lekhajee:
oops sorry. I thought so too. Sent the travis support team an email.

derwolfe:
@glyph  and @lekhajee - No worries. If I remember correctly travis told you last time that multi-os support wasn't available if the repo was configured using python as a language configuration option.

Travis enabled it for me in derwolfe/mimic only after I told them that I was going to change the .travis.yml language option to C in future commits.

lekhajee:
@derwolfe @glyph: Multi-OS has now been enabled, email thread below:

Hey Lekha,

Looks good, thank you for sending that around.

I've enabled multi-os for your project.

Cheers, Mathias

--
Mathias Meyer
support@travis-ci.com

How would you rate my reply?
Great    Okay    Not Good


On Thu, Jan 15, 2015 at 7:31:40 UTC, Lekha Jeevan <lekha.jeevan@rackspace.com> wrote:
Hi,

We have a branch (https://github.com/rackerlabs/mimic/pull/103) that is going to make a change to the travis file to be of the language C. Would you please be able to enable the multi-OS feature for rackerlabs/mimic.

Thanks,
Lekha

derwolfe:
Would it be helpful if I split this into two PRs - one for the changes needed to run multi-os travis, and one containing the code necessary for the mac app?

glyph:
@derwolfe - smaller changes are always better!

derwolfe:
I've added PR #150 that handles multi-os support for travis.

glyph:
oh my goodness it's passing

.travis/install.sh
```diff
@@ -36,6 +36,9 @@ if [[ "$DARWIN" = true ]]; then
             curl -O https://bootstrap.pypa.io/get-pip.py
             sudo python get-pip.py
             ;;
+        bundle)
```
glyph: Why are you building the mac app with a brew-installed python?  Doesn't this just make the app bundle bigger?  Is the system python somehow inadequate?

setup-mac.py
```diff
@@ -0,0 +1,60 @@
...
+VERSION = '0.1'
+ID = 'mimic'
+SCRIPT='bundle/start-app.py'
+TEST_SCRIPT='bundle/run-tests.py'
+PLIST = dict(
+    CFBundleName                = NAME,
+    CFBundleShortVersionString  = ' '.join([NAME, VERSION]),
+    CFBundleGetInfoString       = NAME,
+    CFBundleExecutable          = NAME,
+    CFBundleIdentifier          = 'com.yourdn.%s' % ID,
```
glyph: Do you own `yourdn.com`?

setup-mac.py
```diff
@@ -0,0 +1,60 @@
...
+    LSUIElement                 = '1',
+    LSMultipleInstancesProhibited = '1',
```
glyph: Interesting. I guess we have to prohibit multiple instances because we have no mechanism for customizing the port number or logfile location?

bundle/start-app.py
```diff
@@ -0,0 +1,115 @@
...
+        self.statusItem = NSStatusBar\
+            .systemStatusBar()\
+            .statusItemWithLength_(NSVariableStatusItemLength)
+
+        self.statusItem.setTitle_(u"M")
+        self.statusItem.setHighlightMode_(TRUE)
+        self.statusItem.setEnabled_(TRUE)
+
+        self.quit = NSMenuItem.alloc().initWithTitle_action_keyEquivalent_(
+            "Quit", "terminate:", "")
```
glyph: :heart: nibless applications

docs/development/getting-started.rst
```diff
@@ -84,6 +84,41 @@ Alternately, you can use our ``tox`` job:
...
+
+``mimic`` can be built using `py2app`_ and `pyobjc`_.
+Due to several quirks in the current version of `py2app`_, specific versions of
+libraries are needed in order for `py2app`_ to build correctly.
+These requirements are specified in ``py2app-requirements.txt``.
+
+`py2app`_ requires that a non-system python be installed for bundling.
+To get around this, I've found it easiest to install a brewed 2.7 python.
+This is preferable to using ``pyenv`` because ``brew`` installs a framework
+python.
```
glyph: Oops, I guess this covers my earlier comment :).

This is confusing, though.  I've built bundles with the system python, and I don't think I set up anything special.  Maybe it just worked because I installed a new py2app in a virtualenv?  What exactly do you mean by "bundling"?

.travis/install.sh
derwolfe: I was having difficulty getting the application to build when using my system's python. I then found an SO answer from the py2app creator saying that a non-system python should be used: http://stackoverflow.com/questions/17597616/py2app-isnt-creating-a-standalone-version-of-my-app

But, it is definitely possible that the build mechanism was failing because of something on my system.

setup-mac.py
derwolfe: I don't have a reason for having this - I believe I cargo-culted it form a pyobjc tutorial.

setup-mac.py
derwolfe: Exactly. Since the port cannot be customized in the current version, I thought only one instance could be running at a time.

bundle/start-app.py
derwolfe: Same here! This was much simpler than trying to build with xcode.

docs/development/getting-started.rst
derwolfe: Hmm. That is odd, I have been building the application by installing all of the requirements in a virtualenv and then building the application too. 

It is also possible that I've somehow misconfigured pyenv with my fish shell. I've had issues like this in the past.

Bundling is poor word choice on my part - I really just mean building the mac application bundle.

derwolfe: I'm going to try to rebuild it in a fresh env specifying the system version.

derwolfe: You were completely correct! It builds fine just using the system python. I'll change change the docs and the travis code to reflect this.

.travis/install.sh
derwolfe: I've addressed this below as well - but for clarity - the application builds fine using the system python.

derwolfe:
This build failed in an unexpected way.

Travis has the build erroring out with  ```exceptions.AttributeError: 'ZipPath' object has no attribute 'setContent'```. This build was run from a virtualenv directly, using make.

I have run the build locally, in a virtualenv using the system python. The build passed using ```make``` but failed when run with ```tox -e bundle```.

derwolfe:
Building the application using `make` directly or `tox -e bundle` is causing odd, unexpected behavior when I use a virtualenv created using the system python.

The steps taken; 
1. have virtualenv installed
2. build a virtualenv using /usr/bin/python2.7.
3. pip install -r requirements.txt -r dev-requirements.txt -r py2app-requirements.txt
4. run `make`

I've had the build error out or succeed with no discernible pattern. The error returned is the same `dropin.cache` error as before.

Using `tox -e bundle`, the results seem to be the same. I'll investigate a bit more tomorrow.

.travis/install.sh
derwolfe: @glyph - I may have misunderstood what you were going for in this comment. 

I've been using the brew python to create a standalone version of the application. Yes, this does make the bundle bigger, but, it completely insulates it from any system python changes that could be introduced. I'm basically advocating for bundling the necessary python interpreter code with mimic, instead of having mimic require that python exist somewhere in the user's path.

setup-mac.py
glyph: `CFBundleIdentifier` should be, roughly, a domain name that you own (or at least in a TLD that you own).  I guess we could register a domain for Mimic :) but I think com.rackspace.mimic would be an OK starting point.

glyph:
OK, I suppose I'm not entirely opposed to using homebrewed python if it gets better results for building an app.  This is basically *backwards* from my previous experience, but I'm glad to hear that it works well now.  The one potential issue here is that random changes to Homebrew happen a lot more frequently than random changes to the OS, so we may want to update /usr/local/Cellar to a specific git hash before `brew install`ing.

derwolfe:
The more I look at this, the more I think the successes gleaned by using a brewed python are a red herring. 

If nothing else, the build steps aren't completing correctly a decent percentage of the time. I'm going to do a few more experiments in a branch outside of the PR.

derwolfe:
So, after having done some experiments on my own, I'm able to get the application to build 100% of the time (~20) builds using the brew python, and about 50% of the time using the system. 

The error I receive is the same dropin.cache error described above. The only things I can think of could
1. A race condition somewhere in how the dropin.cache file gets created. I don't see how the python version would affect this.
2. A permissions issue in using the sytem python. Again, I'm not sure why this would be an issue with the different python versions.

I really wish I had a better explanation for this behavior. 

I will add some revert commits to get back to the previous successful build at 22d1df6.

derwolfe:
@glyph I've added a travis build that uses the system python, pyobjc, and py2app. It takes a lot less time to build since it doesn't really have to install much.

I can't explain why, but using tox to run the builds has seemed, at least locally, to result in more builds failing.

.travis/install.sh
derwolfe: I've add a travis build step that builds using the system python. This way both options are available. 
I do agree that building use the system python is actually easier and still presents a stable target.

I've added steps to the getting_started docs detailing how to build the application using either the system python (semi-standalone) or using a brew python (standalone).

derwolfe:
@glyph - I'm pretty sure I've gotten this patch to stabilize. If you'd like to review or comment on it, please feel free to do so without fear of the code/docs changing out from under you :grinning:.

glyph:
@derwolfe -0.01% coverage though! :)

glyph:
@derwolfe - I guess the metadata on this PR is actually wrong, and it is ready for review?  I forget how the intersection of labels / contributors goes (perhaps we should just make you a contributor though…)

glyph:
(Optimistically changing the metadata anyway)

derwolfe:
@glyph - You are right about the label change. I think I tried to change, but github wouldn't allow it. Thanks!

derwolfe:
merged forward to keep up with master.

glyph:
I want to just merge this, but…

I'm a bit uncomfortable with the somewhat convoluted build process.  There are now 6 paths to building an artifact:

  1. `setup.py`
  2. `setup-mac.py`
  3. `Makefile`
  4. `tox.ini`
  5. `.travis.yml`
  6. a narrative story outlined in the documentation

These all interact, and some are platform-specific.

The fact that `.travis.yml` sets `TOX_ENV` to something that `tox` is not actually going to process is very confusing; I expected `tox bundle` to have a `system-bundle` environment.  Why not make the environment be `TRAVIS_MAC_BUILD` if that's the only thing it's going to process?  Also: why bother with a `Makefile` when this shell script could simply call another script?

Also: `setup.py` needs to remain synchronized with `setup-mac.py`.  IMHO there should be a conditional `py2app` import in `setup.py` itself, so that the build stuff remains integrated and we don't accidentally build conflicting versions.

I'd like it if our one-and-only way of building stuff was `tox`, but now that I'm looking over the history it seems as though you were having some issues with that.  I can see a `exceptions.AttributeError: 'ZipPath' object has no attribute 'setContent'` locally when I try via tox, and I have no clue why; perhaps I can look at that later.

But, for now, if we're building with a shell script let's make the shell script the *only* way to build the mac app, promote it out of `.travis.yml`, simplify the documentation to say "run this one script", make it get invoked by travis by setting a different environment variable from `TOX_ENV`, and remove the `Makefile`.  This will at least give us just one thing to update when we figure out what the issue is and move it to get built by tox instead.

Sorry for the long delay here, this contribution is obviously incredibly valuable both for Mimic itself and for the Twisted ecosystem, so thanks again!

derwolfe:
@glyph - Thanks for your feedback! I agree with all of your comments.

I continue to vacillate on this, but, do you think we should build the application bundle using only the system python? 

My reasons in favor are:
- anyone using the mac app would definitely have the system python installed
- it greatly shortens the build time since none of the pyobjc dependencies need to be built.
- the build matrix would be smaller as there would be *only* one way to build the application.

Against:
- there might be a feature missing in the 2.7.6 interpreter that a brew python might offer (is there any functionality, e.g. a security update found in 2.7.9 not in 2.7.6 that mimic *should* have?)

glyph:
I think that we should do the official build with the system python because build performance is an increasingly significant issue, and having travis have to churn through building all of pyobjc would be really unpleasant.

glyph:
However, I feel like we should easily be *able* to build against whichever Python the user has active :).

derwolfe:
@glyph - I've pushed the following changes to the PR:

1. Use a single build command to build the application. This is done via ``build-app.sh``. It creates a new virtualenv, installs the dependencies, and builds the application. This is meant to emulate how tox should work.
2. Change the documentation to reflect the new build command and the fact that the official build uses the system python.
3. Merge ``setup-mac.py`` into ``setup.py`` and use a conditional import to check if the ``py2app`` is available. The logic I've added checks if ``py2app`` is available, if it is, the command only allows the user to build the application (this might not be ideal). If ``py2app`` is not available, then the normal setup command, with install_requires, etc. is used.
4. The ``tox_env`` bundle has been removed. Travis now has a single ``MACAPP_ENV`` variable that will be set to ``system`` to build the application using ``build-app.sh``.

Thanks!

setup.py
```diff
@@ -4,20 +4,87 @@
...
+    )
+
+    class BuildWithCache(py2app):
+        """
+        Before building the application rebuild the `dropin.cache` files.
+        """
+
+        def run(self):
+            """
+            This generates `dropin.cache` files for mimic"s plugins.
```
derwolfe: should be '

glyph:
@derwolfe - can you merge master back into this so it can get an automatic merge again?

derwolfe:
It looks like there is an issue with the mimic.test.test_tap and the py2app builder. At least locally, it is returning an import error.

derwolfe:
It looks like it isn't pulling in the mimic/twisted/plugins/mimic.py file.

glyph:
Could this be because we added the `--develop` option to `tox` in b4f61dd07abf83c55eed4640a1bf5b65061910f8 in order to get the coverage report to accurately reflect the plugin file being imported?

derwolfe:
I'm pretty sure it isn't - it is happening locally as well.

glyph:
Hrm. But only after merging master?

glyph:
Now I'm getting this: https://gist.github.com/glyph/fcc354d38222dd9d4ae0

derwolfe:
It looks this is the commit that introduced the change: b71c72fc018b6f2c388f187117aa3af3c622f724

derwolfe:
Do you get the same error if you run ./build-app.sh?

glyph:
No, then I get `exceptions.ImportError: No module named mimic`, same as on travis…

glyph:
Looking inside `site_packages.zip`, I don't even see twisted's *own* plugins:

```
★ unzip -l dist/mimic.app/Contents/Resources/lib/python2.7/site-packages.zip | grep twisted/plugins
      499  03-03-15 17:58   twisted/plugins/__init__.pyc
     9751  03-03-15 17:58   twisted/plugins/dropin.cache
````

**Comment by derwolfe** [comment]
Yes - I just found that as well! It looks like it generates a dropin.cache, but nothing else..

**Comment by glyph** [comment]
It's possible that `--system-site-packages` is causing some confusion.  Have you had any better luck with your own pyobjc?

**Comment by glyph** [comment]
Without `--system-site-packages`, I'm back to the `scan_code` error.

**Comment by derwolfe** [comment]
With the scan_code error, which version of py2app are you running? If I remember correctly, newer versions of py2app have an issue. I need to pin the py2app dependencies, instead of just minimum requirements.

**Comment by glyph** [comment]
Yeah, I was just messing around with py2app versions…

**Comment by glyph** [comment]
I think the version with the error was 0.9?

**Comment by glyph** [comment]
When pinning it to 0.8.1, I now get this at the end of the test run:

```
+ python setup.py py2app
Traceback (most recent call last):
  File "setup.py", line 8, in <module>
    from py2app.build_app import py2app
  File "/Users/glyph/Projects/Mimic/venv-app/lib/python2.7/site-packages/py2app/__init__.py", line 31, in <module>
    __version__ = pkg_resources.require('py2app')[0].version
  File "/Users/glyph/Projects/Mimic/venv-app/lib/python2.7/site-packages/pkg_resources/__init__.py", line 918, in require
    needed = self.resolve(parse_requirements(requirements))
  File "/Users/glyph/Projects/Mimic/venv-app/lib/python2.7/site-packages/pkg_resources/__init__.py", line 810, in resolve
    raise VersionConflict(dist, req).with_context(dependent_req)
pkg_resources.ContextualVersionConflict: (modulegraph 0.10.4 (/Users/glyph/Projects/Mimic/venv-app/lib/python2.7/site-packages), Requirement.parse('modulegraph>=0.11'), set(['py2app']))
+ ./dist/mimic.app/Contents/MacOS/run-tests
build-app.sh: line 33: ./dist/mimic.app/Contents/MacOS/run-tests: No such file or directory
```

derwolfe:
I'm now getting odd framework python build errors on my machine.

glyph:
I think one thing that might be screwing it up is the fact that if py2app is available at all, the `setup.py` no longer provides any of the *other* necessary build metadata to the build process, such as `package_dir` or `packages`.

glyph:
I am deeply confused.  py2app appears to have regressed in some way since the last time I was looking at this; and not just regressed in its current version, but older versions are broken too.  I *really* don't understand what is going on, but clearly that `AttributeError` is a py2app bug and we should report it.

derwolfe:
I'm quite confused myself. One way or another, I'm now getting the following error, using the same steps as before.

2015-03-03 21:46:09.077 run-tests[33926:88303] A Python runtime not could be located.  You may need to install a framework build of Python, or edit the PyRuntimeLocations array in this application's Info.plist file.

The env is still being built using the system python too.

glyph:
*one* of the many bugs blocking us seems to be https://bitbucket.org/ronaldoussoren/py2app/issue/156/virtualenvpy-recipe-calls-methods-renamed so I've commented there...

glyph:
Since that bug is in the `virtualenv` recipe, I have now been messing around with using `--user` installation to get things set up instead of a virtualenv.  I'm getting a little further, but still having trouble generating the plugin.cache and getting it into the zip file.

lekhajee:
Merged #191, which includes all the commits from this PR. Closing this. Thanks!

PR #131, opened 2014-12-19
maasplugin monitoring zones

tokunbo:


ksheedlo:
@tokunbo it looks like I originally authored this into `maas-plugin` (see #126), is there any way you could keep me as the author when you merge into master?

Issue #143, opened 2015-01-14
The MAAS api returns entity ids prefixed by 'en' and check ids prefixed by 'ch' as against 'e' and 'c'. 


lekhajee:
The MAAS api returns entity ids prefixed by 'en' and check ids prefixed by 'ch' as against 'e' and 'c'. 

@tokunbo : any reason you had them start with 'e' and 'c'

glyph:
Fixed by #144.

PR #144, opened 2015-01-14
fix entity and check id to start with 'en' and 'ch' respectively

lekhajee:
Fixes issue #143

### DIFF
--- a/mimic/test/test_maas.py
+++ b/mimic/test/test_maas.py
@@ -154,20 +154,20 @@ class MaasAPITests(SynchronousTestCase):
         self.uri = helper.uri
         self.entity_id = self.getXobjectIDfromResponse(self.createEntity('ItsAnEntity'))
         self.check_id = self.getXobjectIDfromResponse(self.createCheck('ItsAcheck',
-                                                      self.entity_id))
+                                                                       self.entity_id))
         self.alarm_id = self.getXobjectIDfromResponse(self.createAlarm('ItsAnAlarm',
-                                                      self.entity_id,
-                                                      self.check_id))
+                                                                       self.entity_id,
+                                                                       self.check_id))
         self.nt_id = self.getXobjectIDfromResponse(self.createNotification('ItsANotificationTarget'))
         self.np_id = self.getXobjectIDfromResponse(self.createNotificationPlan('ItsANotificationPlan'))
         self.sp_id = self.getXobjectIDfromResponse(self.createSuppression('ItsASuppression'))
 
     def test_resource_ids(self):
         """
-        MAAS sets IDs for reources it created by prefixing it with
-        the first two characters of the resource type.
-        For example an entity id is prefixed with 'en'.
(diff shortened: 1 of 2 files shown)


cyli:
LGTM!

mimic/test/test_maas.py
```diff
@@ -165,11 +165,17 @@ def setUp(self):
         np_id = self.getXobjectIDfromResponse(self.createNotificationPlan('ItsANotificationPlan'))
         sp_id = self.getXobjectIDfromResponse(self.createSuppression('ItsASuppression'))
         self.assertNotEquals(None, entity_id)
+        self.assertTrue(entity_id.startswith('en'))
```
glyph: Asserts in a `setUp` are generally poor form; when a test fails, an individual test should fail and not a whole suite.  Can you move this setup logic into a thing that can be called from a test, and move the asserts to only *one* test, rather than repeating them in everyone's `setUp`?

glyph:
I think the main comment I have about this particular change is that it adds a bunch of asserts without adding a docstring explaining what those asserts actually mean.  This means I have the same problem as @cyli: I would +1 it except for the fact that I don't really know what the significance of the change is.  Which means it's a change that could use more accompanying maintenance documentation :).  Can you find any documentation reference in the monitoring API and link to it from a test docstring?  If not, I guess just moving the asserts to a test which explains the significance of the prefixes would be helpful.

lekhajee:
@glyph: oops, moved the test from the setUp and included docstring. Looks like the other tests within `test_maas` are not using the created resources from the setUp but doing a GET to be able to use them. Will fix that in a different PR.

mimic/test/test_maas.py
```diff
@@ -152,24 +152,35 @@ def setUp(self):
...
+        self.alarm_id = self.getXobjectIDfromResponse(self.createAlarm('ItsAnAlarm',
+                                                      self.entity_id,
+                                                      self.check_id))
+        self.nt_id = self.getXobjectIDfromResponse(self.createNotification('ItsANotificationTarget'))
+        self.np_id = self.getXobjectIDfromResponse(self.createNotificationPlan('ItsANotificationPlan'))
+        self.sp_id = self.getXobjectIDfromResponse(self.createSuppression('ItsASuppression'))
+
+    def test_resource_ids(self):
+        """
+        MAAS sets IDs for reources it created by prefixing it with
```
glyph: "resources"

mimic/test/test_maas.py
```diff
@@ -152,24 +152,35 @@ def setUp(self):
...
+        the first two characters of the resource type.
+        For example an entity id is prefixed with 'en'.
```
glyph: This whole paragraph looks like it's wrapped to a way-too-narrow margin, too.  Maybe tell sublime to re-wrap it to 79?

glyph:
If you could just address those two minor things about the docstring I think it's ready to land (please do so yourself once you've fixed them up).

PR #148, opened 2015-01-16
Object model for Nova

glyph:
This should serve as a better example for new plugins, and sets up the infrastructure for error injection.


mimic/model/nova_objects.py
```diff
@@ -0,0 +1,392 @@
...
+            ],
+            public_ips=[
+                IPv4Address(address="198.101.241.{0}".format(ipsegment())),
+                IPv6Address(address="2001:4800:780e:0510:d87b:9cbc:ff04:513a")
+            ],
+            creation_request_json=creation_json,
+            flavor_ref=server_json['flavorRef'],
+            image_ref=server_json['imageRef'] or '',
+            disk_config="AUTO",
+            # ^ TODO: should be: server_json['OS-DCF:diskConfig'],
```
lekhajee: Do we have issues for the TODOs, can we associate them

mimic/model/nova_objects.py
```diff
@@ -0,0 +1,392 @@
...
+            # ^ TODO: should be random
+        )
+        collection.servers.append(self)
+        return self
+
+
+@attributes(["address"])
+class IPv4Address(object):
+    """
+    An IPv6 address for a server.
```
lekhajee: The docstring reads `IPv6`

mimic/model/nova_objects.py
```diff
@@ -0,0 +1,392 @@
...
+
+    def json(self):
+        """
+        A JSON-serializable representation of this address.
+        """
+        return {"addr": self.address, "version": 4}
+
+
+@attributes(["address"])
+class IPv6Address(object):
```
lekhajee: Can we have one `IPAddress`class  that takes the version as well, instead of one for IPv4 and one for IPv6

mimic/model/nova_objects.py
```diff
@@ -0,0 +1,392 @@
...
+            'server': {
+                "OS-DCF:diskConfig": self.disk_config,
+                "id": self.server_id,
+                "links": self.links_json(absolutize_url),
+                "adminPass": self.admin_password,
+            }
+        }
+
+    @classmethod
+    def from_creation_request_json(cls, collection, creation_json,
```
lekhajee: Could we maybe say `create_server_from_request_json`, the current name is not very self explanatory of what the method does

mimic/model/nova_objects.py
glyph: I'd actually prefer to keep two separate classes, because we could add validation to the format for the address string itself at some point, and that differs between the two classes.

mimic/model/nova_objects.py
glyph: `from_<x>` is the naming convention for a class-method constructor, so I'm following that.

The problem with `create_...` is that, as a verb, it implies that it's going through the full process of "creating" the server, which also involves *adding* it to the collection; whereas, this method is just instantiating the Server object, which may or may not be fully "created" as the caller sees fit.

This is really hair-splitting of course, but the existing convention of `from_` seems to avoid the issue of whether "create" and "instantiate" are really synonyms in this context.

mimic/rest/nova_api.py
```diff
@@ -213,19 +195,16 @@ def get_limit(self, request, tenant_id):
...
-        )
+        return json.dumps(get_limit())
 
     @app.route('/v2/<string:tenant_id>/servers/<string:server_id>/ips', methods=['GET'])
     def get_ips(self, request, tenant_id, server_id):
         """
-        Returns a get flavor response, for any given flavorid.
-        (currently the GET ips works only after a GET server after the server is created)
+        Returns a get flavor response, for any given flavorid.  (currently the
+        GET ips works only after a GET server after the server is created)
```
lekhajee: this is fixed by the object model. Could we remove the comment

lekhajee:
+1

mimic/model/nova_objects.py
```diff
@@ -0,0 +1,392 @@
...
+        access Mimic under different hostnames and it will give you URLs
+        appropriate to how you accessed it every time.  This is intentionally
+        to support the use-case of running tests against your local dev machine
+        as 'localhost' and then showing someone else the state that things are
+        in when they will have to access your machine under a different
+        hostname and therefore a different URI.
+
+    :param ipsegment: A hook provided for IP generation so the IP addresses in
+        tests are deterministic; normally a random number between 0 and 255.
+    """
```
lekhajee: include a param for the hook

mimic/model/nova_objects.py
```diff
@@ -0,0 +1,392 @@
...
+    if hook is not None:
+        hook(new_server)
+    response = new_server.creation_response_json(absolutize_url)
+    http.setResponseCode(ACCEPTED)
+    return dumps(response)
+
+
+def default_with_hook(function):
+    """
+    Convert a function that takes a function.
```
lekhajee: better docstring ?

PR #150, opened 2015-01-18
Multi-os support setup files for travis

derwolfe:
This patch enables multi-os support for testing using Travis. It basically just makes it so each of the tox environments can be tested on both OSX and Linux.

My thought is that this patch be merged prior to the mac-app bundle PR #103.

Thanks!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,29 +1,44 @@
-language: python
+language: c
+
+os:
+  - osx
+  - linux
+
+compiler:
+    - clang
+    - gcc
 
 env:
-  - TOX_ENV=lint
-  - TOX_ENV=py26
-  - TOX_ENV=py27
-  - TOX_ENV=pypy
-  - TOX_ENV=docs
-  - TOX_ENV=docs-spellcheck
-  - TOX_ENV=docs-linkcheck
+  matrix:
+     - TOX_ENV=lint
+     - TOX_ENV=py26
+     - TOX_ENV=py27
+     - TOX_ENV=pypy
+     - TOX_ENV=docs
+     - TOX_ENV=docs-spellcheck
+     - TOX_ENV=docs-linkcheck
 
 matrix:
   allow_failures:
-    - env: "TOX_ENV=docs-linkcheck"
+    - env: TOX_ENV=docs-linkcheck
+  exclude:
+    - os: osx
+      compiler: gcc
+    - os: linux
+      compiler: clang
 
 install:
   - ./.travis/install.sh
 
-script: tox -e $TOX_ENV
+script:
+  - ./.travis/run.sh
 
 after_success:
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -1,29 +1,44 @@
...
+     - TOX_ENV=docs
+     - TOX_ENV=docs-spellcheck
+     - TOX_ENV=docs-linkcheck
 
 matrix:
   allow_failures:
-    - env: "TOX_ENV=docs-linkcheck"
+    - env: TOX_ENV=docs-linkcheck
+  exclude:
+    - os: osx
```
cyli: Apologies, my travis-fu is pretty weak.  Do these excludes this mean we would like to run osx/clang and linux/gcc, and thus not run osx/gcc and linux/clang?

Just wondering because travis seems to have run linux/gcc and linux/clang, and skipped osx entirely, if I'm reading the build message right?

glyph: https://github.com/rackerlabs/mimic/pull/103#issuecomment-70138476 indicates to me that multi-OS is in fact enabled on this repo so we should in fact be seeing OS X builds in the matrix, and I don't see any, so … I think you're reading it right.  The output from the tests seemed consistent with the labels in the matrix, and it was run 2 days after that comment indicating that the multi-OS support is in fact enabled.

Just to make sure, I re-ran the build to see what happened.  Maybe we need to get in touch with support again?

glyph: I just shot support an email.  I'm pretty sure this is our mistake, but I just can't see what it is.

glyph:
Taking ownership until support gets back to me, but if somebody else figures out what's going on feel free to grab it :).

.travis.yml
derwolfe: I'm not sure why this isn't being picked up either, as I have the same configuration in my PR. I think this is an error with travis as well.

glyph:
Despite [the comment which appeared to indicate it's been enabled](https://github.com/rackerlabs/mimic/pull/103#issuecomment-70138476), I just got this reply from support:

> The feature needs to be enabled manually on our end. This has not happened to rackerlabs/mimic. I've just done this for you.

So perhaps it just wasn't enabled.

glyph:
@derwolfe - I've done a rebuild, but the matrix on that build says "linux" everywhere so I'm not sure if it effectively predates multi-OS support.  Can you try pushing another revision, to see if that will trigger a build?

derwolfe:
@glyph it looks like that did it! The commit I pushed added a clang build to linux. It looks to me like it is building for OSX now. :sparkles: Thanks! :sparkles:

In reading through the docs for Travis, I also realized I don't have a strong reason for excluding gcc builds from OSX or clang builds from linux.  My thought was that on OSX 10.9 through 10.10 I believe clang is the default and that ubuntu by default uses gcc.

glyph:
Right on! Happy to merge this at last ;).

Issue #146, opened 2015-01-15
Warn users if username, tenant_id pair changes during authentication


lekhajee:
If you authenticate once using a tenantId / username pair in your Identity request, and then you authenticate again later with the same username but a different tenantId, you'll get the first tenantId back, forever.  If you authenticate without using a tenantId at all on your first authentication request, you'll get one randomly generated, and subsequent requests will give back that randomly generated one.

The Identity mimic should at least log an error if the request includes an incorrect tenantId that doesn't match its existing idea of what the tenant should be.

glyph:
Huh, I edited the description but it's still @lekhajee's picture next to it.  Github probably shouldn't let me put words in someone else's mouth like that, project owner or not :).

cyli:
I was originally going to suggest that maybe we should not make usernames unique across all tenants, but apparently at least Rackspace does not seem to allow duplicate usernames across different tenants.

I believe if the wrong `tenantName`/`tenantId` is passed in the user auth section, identity returns a 401 Unauthorized?  Mimic can probably do the same.

glyph:
Fixed by #154

glyph:
> Rackspace does not seem to allow duplicate usernames across different tenants

Just for future reference... although lots of examples show identity authentication including a tenant ID, in fact, all you need (as on mycloud.rackspace.com) is a username and password, which means the username needs to be a globally unique lookup key.

Personally at least I have no idea why the tenant ID is ever included, I've never seen an explanation of what including it in auth does :).

ksheedlo:
@glyph Including the tenant ID is necessary in the case of multiple clients. For instance, I have Cloud Intelligence and an API client that runs from the command line. I can't control the tenant ID that Intelligence uses - it's my Rackspace tenant ID and is not easy to change. The only way for both clients to see the same data is for me to configure the API client to use my Rackspace tenant ID.

Issue #153, opened 2015-01-21
New role should not be added every time the user auths


cyli:
After authing against mimic identity once, the JSON has:

```
"user": {
            "id": ...
            "name": ...
            "roles": [
                {
                    "description": "User Admin Role.", 
                    "id": "3", 
                    "name": "identity:user-admin"
                }, 
                {
                    "description": "User Admin Role.", 
                    "id": "3", 
                    "name": "identity:user-admin"
                } 
            ]
       }
}
```

After authing twice...

```
"user": {
            "id": ...
            "name": ...
            "roles": [
                {
                    "description": "User Admin Role.", 
                    "id": "3", 
                    "name": "identity:user-admin"
                }, 
                {
                    "description": "User Admin Role.", 
                    "id": "3", 
                    "name": "identity:user-admin"
                }, 
                {
                    "description": "User Admin Role.", 
                    "id": "3", 
                    "name": "identity:user-admin"
                }
            ]
       }
}
```

And so forth...

PR #154, opened 2015-01-21
RFR - Return 401 if authing with the wrong tenant ID

cyli:
This should fix issue #146.

Basically if you auth once, and auth again but specify a different tenant ID, identity should return 401.
I unfortunately also had to sneak a change to actually validate the token, because I couldn't trigger the failure otherwise.  Luckily this doesn't break any tests.

Not super happy about the `get_token_and_service_catalog` refactor, but the error message for username/password and username/api-key differs very slightly from the token/tenant_id auth.  Suggestions welcome.

### DIFF
--- a/mimic/rest/auth_api.py
+++ b/mimic/rest/auth_api.py
@@ -46,58 +46,78 @@ class AuthApi(object):
         tenant_id = (content['auth'].get('tenantName', None) or
                      content['auth'].get('tenantId', None))
 
-        try:
-            if content['auth'].get('passwordCredentials'):
-                username = content['auth']['passwordCredentials']['username']
-                password = content['auth']['passwordCredentials']['password']
-                session = self.core.sessions.session_for_username_password(
-                    username, password, tenant_id)
-            elif content['auth'].get('RAX-KSKEY:apiKeyCredentials'):
-                username = content['auth']['RAX-KSKEY:apiKeyCredentials'][
-                    'username']
-                api_key = content['auth']['RAX-KSKEY:apiKeyCredentials'][
-                    'apiKey']
-                session = self.core.sessions.session_for_api_key(
-                    username, api_key, tenant_id)
-            elif content['auth'].get('token') and tenant_id:
-                session = self.core.sessions.session_for_tenant_id(tenant_id)
+        def format_response(callable_returning_session,
+                            nonmatching_tenant_message_generator):
+            try:
+                session = callable_returning_session()
+            except NonMatchingTenantError as e:
+                request.setResponseCode(401)
+                return json.dumps({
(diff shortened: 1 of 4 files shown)


mimic/rest/auth_api.py
```diff
@@ -44,41 +45,79 @@ def get_token_and_service_catalog(self, request):
 
         tenant_id = (content['auth'].get('tenantName', None) or
                      content['auth'].get('tenantId', None))
+
+        def format_response(callable_returning_session,
```
glyph: This function is starting to push the boundaries of being too long.  Looking at it, I don't see any obvious way to break it into reasonable smaller pieces, but it's something to think about for future changes here.

mimic/rest/auth_api.py
```diff
@@ -44,41 +45,79 @@ def get_token_and_service_catalog(self, request):
...
+                    session.tenant_id,
+                    entry_generator=lambda tenant_id:
+                    list(self.core.entries_for_tenant(
+                         tenant_id, prefix_map, base_uri_from_request(request))),
+                    prefix_for_endpoint=lookup,
+                    response_token=session.token,
+                    response_user_id=session.user_id,
+                    response_user_name=session.username,
+                )
+                result['access']['user']['roles'].append({
```
glyph: This right here is the cause of #153 I am pretty sure :)

cyli: Yep.  I was going to delete it, but figured separate PR.  >:D Unless I can delete it here?

mimic/rest/auth_api.py
```diff
@@ -44,41 +45,79 @@ def get_token_and_service_catalog(self, request):
...
+
         elif content['auth'].get('RAX-KSKEY:apiKeyCredentials'):
-            username = content['auth']['RAX-KSKEY:apiKeyCredentials']['username']
-            api_key = content['auth']['RAX-KSKEY:apiKeyCredentials']['apiKey']
-            session = self.core.sessions.session_for_api_key(
-                username, api_key, tenant_id)
+            username = content['auth']['RAX-KSKEY:apiKeyCredentials'][
+                'username']
+            api_key = content['auth']['RAX-KSKEY:apiKeyCredentials'][
+                'apiKey']
```
glyph: Come to think of it, maybe this is a place that we can split things up - now that this is indented so far that the repetition is causing every line to be wrapped, perhaps we could split up

mimic/rest/auth_api.py
glyph: Separate PR sounds good.

glyph:
I'll figure out how to split up the function in a separate PR.  LGTM.

PR #157, opened 2015-01-22
WIP - Add change-user-info endpoint to identity

cyli:
Which allows users to change their email, etc., but most importantly, change their passwords.

Planned features and limitations:
- only going to allow email+password change, not whether the user is enabled, default region, etc.
- not actually going to save the password or passwords (??? - could be convinced to go the other way on this one)
- on password change the tokens and api key are going to be reset (need to verify that the api key does indeed change)

### DIFF
--- a/mimic/session.py
+++ b/mimic/session.py
@@ -43,6 +43,13 @@ class NonMatchingTenantError(Exception):
     """
 
 
+@attributes(["user_id"])
+class NoSuchUserIdError(Exception):
+    """
+    No such user with the given user ID exists.
+    """
+
+
 class SessionStore(object):
     """
     A collection of sessions addressable by multiple different keys.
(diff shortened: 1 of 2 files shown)


mimic/session.py
```diff
@@ -102,6 +112,7 @@ def _new_session(self, username_key=None, **attributes):
         self._username_to_token[username_key] = session.token
         self._token_to_session[session.token] = session
         self._tenant_to_token[session.tenant_id] = session.token
+        self._user_id_to_sesion[session.user_id] = session
```
glyph: `sesion`→`session`

The pattern for most of the other things here though is that we have one thing that points at the actual `session` object, `_token_to_session`, and then lots of other dictionaries which point at the token, thereby indirectly referencing the session.  I'd kinda like to keep it to one direct reference, so that we can clear out the session by popping it out of one place instead of N places.

glyph:
Poke - this needs a master merge to build again.

cyli:
Yeah, sorry, haven't had a chance to work on this at all.  :|

glyph:
This branch has conflicts, has been idle for months, and no "needed by" labels so I'm calling it.  Anyone with an interest in this may always feel free to reopen!

PR #159, opened 2015-01-23
Nodes should by default be added as ACTIVE, like servers in Nova.

cyli:
Intermediary states should probably be introduced via a control plane.


glyph:
Seems consistent with the Nova plugin, nice and short change, thanks.

PR #160, opened 2015-01-28
fail when request can't be generated

glyph:
Add a test for our test helper to make sure that it is implemented in such a way that you can see when you give request the wrong arguments.


lvh:
LGTM!

PR #165, opened 2015-02-02
Nova expects all metadata values to be strings, so nova building/error injection has to handle strings

cyli:


lvh:
LGTM assuming CI likes it :)

PR #166, opened 2015-02-02
Normalize links in nova

cyli:
The flavor links blob and image links blob were missing the "rel" attribute, or it was in the wrong place.

Actually add a test to attempt to validate the json responses returned by nova APIs.


glyph:
Jeez, how did our existing tests even pass? Good catch, thank you.

cyli:
We didn't validate the blob previously  :|

Issue #167, opened 2015-02-03
exclude osx/gcc and linux/clang builds from travis


derwolfe:
The build matrix for travis is running a bit slowly and failing on builds that aren't really necessary.

A fix for this would be to setup travis to run builds only for osx/clang and gcc/linux.

glyph:
Indeed, addressed.

PR #168, opened 2015-02-03
exlude clang on linux

derwolfe:
Exclude osx/gcc and linux/clang.


derwolfe:
If passes, this would close issue #167

glyph:
Looks good!  (Also, I notice that coveralls seems to have just upgraded to be a multi-status plugin, so that's awesome.)

PR #171, opened 2015-02-06
Adding RAX-KSKEY:apiKeyCredentials

tokunbo:


mimic/rest/auth_api.py
```diff
@@ -124,6 +124,17 @@ def get_username(self, request, tenant_id):
...
         return json.dumps(dict(user=dict(id=session.username)))
 
+    @app.route('/v2.0/users/<string:userid>/OS-KSADM/credentials/RAX-KSKEY:apiKeyCredentials',
+               methods=['GET'])
+    def rax_kskey_apikeycredentials(self, request, userid):
+        """
+        Support, such as it is, for the apiKeysCredentials call.
+        """
+        apikey = '1337133713371337133713371337'
+        username = 'mimicuser'
```
glyph: This state should really be stored on the `Session`, since that is already keyed by username, using the same idiom of "create it if there's nothing that matches already".  Hopefully it won't be much harder to do this?

tokunbo: How do I get the username?

glyph: You have to get the username from the session object. The session can be retrieved from `self.core.session_for_<something>`.  I think `userid` here is the same as `username`, so you could use the existing `session_for_username_password` method with a bogus password, or you could create a new `session_for_username` that explicitly doesn't authenticate anything.

mimic/rest/auth_api.py
```diff
@@ -124,6 +124,19 @@ def get_username(self, request, tenant_id):
         session = self.core.sessions.session_for_tenant_id(tenant_id)
         return json.dumps(dict(user=dict(id=session.username)))
 
+    @app.route('/v2.0/users/<string:user_id>/OS-KSADM/credentials/RAX-KSKEY:apiKeyCredentials',
+               methods=['GET'])
+    def rax_kskey_apikeycredentials(self, request, user_id):
+        """
+        Support, such as it is, for the apiKeysCredentials call.
+        """
+        username = "mimic_userid_"+user_id
```
lekhajee: nit: space around operator

mimic/rest/auth_api.py
```diff
@@ -124,6 +124,19 @@ def get_username(self, request, tenant_id):
...
+        if user_id in self.core.sessions._userid_to_session:
+            username = self.core.sessions._userid_to_session[user_id].username.decode('ascii')
+        apikey = '7fc56270e7a70fa81a5935b72eacbe29'  # echo -n A | md5sum
```
lekhajee: Do we want to create a session for the user if one does not exist.

mimic/rest/auth_api.py
```diff
@@ -124,6 +124,23 @@ def get_username(self, request, tenant_id):
...
+        username = None
+        if user_id in self.core.sessions._userid_to_session:
+            username = self.core.sessions._userid_to_session[user_id].username.decode('ascii')
+            apikey = '7fc56270e7a70fa81a5935b72eacbe29'  # echo -n A | md5sum
+            return json.dumps({'RAX-KSKEY:apiKeyCredentials': {'username': username,
+                               'apiKey': apikey}})
+        else:
+            request.setResponseCode(404)
+            return json.dumps({'itemNotFound':
+                              {'code': 404, 'message': 'User '+user_id+' not found'}})
```
lekhajee: nit : space around operator. I wonder why lint is not catching this

mimic/rest/auth_api.py
```diff
@@ -124,6 +124,23 @@ def get_username(self, request, tenant_id):
         session = self.core.sessions.session_for_tenant_id(tenant_id)
         return json.dumps(dict(user=dict(id=session.username)))
 
+    @app.route('/v2.0/users/<string:user_id>/OS-KSADM/credentials/RAX-KSKEY:apiKeyCredentials',
+               methods=['GET'])
+    def rax_kskey_apikeycredentials(self, request, user_id):
+        """
+        Support, such as it is, for the apiKeysCredentials call.
+        """
+        username = None
```
lekhajee: why set `username` to `None`

mimic/test/test_auth.py
```diff
@@ -625,6 +625,40 @@ def test_token_and_catalog_for_password_credentials_wrong_tenant(self):
...
+            }
+        }
+        (response, json_body) = self.successResultOf(json_request(
+            self, root, "POST", "/identity/v2.0/tokens", creds))
+        self.assertEqual(response.code, 200)
+        user_id = json_body['access']['user']['id']
+        username = json_body['access']['user']['name']
+        (response, json_body) = self.successResultOf(json_request(
+            self, root, "GET",
+            "/identity/v2.0/users/"+user_id+"/OS-KSADM/credentials/RAX-KSKEY:apiKeyCredentials"
```
lekhajee: nit : space around operator

lekhajee:
+1 except for some non trivial comments

tokunbo:
retest this pleaseeeeeeeeeeeeeeeeeeeeeeeeeeeee

mimic/rest/auth_api.py
```diff
@@ -124,6 +124,22 @@ def get_username(self, request, tenant_id):
...
+        if user_id in self.core.sessions._userid_to_session:
+            username = self.core.sessions._userid_to_session[user_id].username.decode('ascii')
+            apikey = '7fc56270e7a70fa81a5935b72eacbe29'  # echo -n A | md5sum
+            return json.dumps({'RAX-KSKEY:apiKeyCredentials': {'username': username,
+                               'apiKey': apikey}})
+        else:
+            request.setResponseCode(404)
+            return json.dumps({'itemNotFound':
+                              {'code': 404, 'message': 'User ' + user_id + ' not found'}})
+
```
ksheedlo: What if the user authenticated to mimic using their own API key credentials? Should they get the same key back?

mimic/test/test_auth.py
```diff
@@ -625,6 +625,40 @@ def test_token_and_catalog_for_password_credentials_wrong_tenant(self):
...
+        self.assertEqual(response.code, 200)
+        user_id = json_body['access']['user']['id']
+        username = json_body['access']['user']['name']
+        (response, json_body) = self.successResultOf(json_request(
+            self, root, "GET",
+            "/identity/v2.0/users/" + user_id + "/OS-KSADM/credentials/RAX-KSKEY:apiKeyCredentials"
+        ))
+        self.assertEqual(response.code, 200)
+        self.assertEqual(json_body['RAX-KSKEY:apiKeyCredentials']['username'],
+                         username)
```
ksheedlo: This test should probably assert something about the API key returned in the response, although it's probably inappropriate to check for an exact match against the source code. Maybe use a regex to assert that it's a 32 chars hex string?

mimic/rest/auth_api.py
tokunbo: Meh, this is just a stub code so that Agent-Install page doesn't get a 404. I'm actually not trying to make this work 100% like production. I say it's good enough for now because my real goal is just getting the e2e test working.

mimic/test/test_auth.py
tokunbo: meh! ....I'm just going to test that string is 32 in length.

mimic/rest/auth_api.py
glyph: Perhaps we could file a follow-up issue?

mimic/test/test_auth.py
glyph: IMHO, tests should always be as simple as possible.  Regexes are … not.  It's really easy to write a false-positive regex or modify it so it matches something counter-intuitive.  So let's try to keep regexes out of the tests as much as possible :).

mimic/rest/auth_api.py
tokunbo: maybe mimic could have a simple function that just returns the md5 of the username as the APIkey.  Then it'd always been the same and you could make the auth fail if the md5(username) isn't equal to the apikey you got... 


....I guess. idunno

mimic/test/test_auth.py
ksheedlo: You can assert that it's a 32 character string, or just any truthy string. My issue was I don't like the idea of testing an API key credentials API without testing that anything resembling the actual API key.

mimic/rest/auth_api.py
ksheedlo: @tokunbo if you're using md5, you're probably doing it wrong. Anyway, I don't know if people care that they get their same API key back, so I'm happy to just let this merge and see if anyone needs it enough to file their own issue.

ksheedlo:
LGTM to me :+1:

PR #172, opened 2015-02-10
RFR: Include validate token api call

lekhajee:
Includes the Identity admin api call that validates the token. For the purposes of the ele testing, such a validation always passes and also creates a new session for the given tenant_id and token_id.

### DIFF
--- a/mimic/canned_responses/auth.py
+++ b/mimic/canned_responses/auth.py
@@ -113,7 +113,7 @@ def get_token(tenant_id,
         }
     }
 
-    if (entry_generator and prefix_for_endpoint):
+    if entry_generator is not None and prefix_for_endpoint is not None:
         response["access"]["serviceCatalog"] = list(entry_json())
     return response
 
--- a/mimic/canned_responses/auth.py
+++ b/mimic/canned_responses/auth.py
@@ -11,7 +11,8 @@ GLOBAL_MUTABLE_TOKEN_STORE = {}
 HARD_CODED_TOKEN = "fff73937db5047b8b12fc9691ea5b9e8"
 HARD_CODED_USER_ID = "10002"
 HARD_CODED_USER_NAME = "mimictestuser"
-HARD_CODED_ROLES = [{"id": "3", "description": "User Admin Role.",
+HARD_CODED_ROLES = [{"id": "3",
+                     "description": "User Admin Role.",
                      "name": "identity:user-admin"}]
 
 
@@ -22,6 +23,38 @@ def format_timestamp(dt):
     :param datetime.datetime dt: A datetime.datetime object to be formatted.
     """
     return dt.strftime('%Y-%m-%dT%H:%M:%S.999-05:00')
+
+
+def impersonator_user_role(id, name):
+    """
+    Canned response for validate token if the token being validated
+    is an impersonated token.
+    """
+    return {
+        "id": id,
+        "roles": [
+            {"id": "10000001",
+             "serviceId": "test-eabb70a0e702a4626977c331d5c4",
+             "description": "Service admin role for cloud feeds access. Assign only to service users",
+             "name": "cloudfeeds:service-admin"},
+            {"id": "10000002",
+             "serviceId": "test-eabb70a0e702a4626977c331d5c4",
+             "description": "Checkmate Access role",
+             "name": "checkmate"},
+            {"id": "100000003",
+             "serviceId": "test-eabb70a0e702a4626977c331d5c4",
+             "description": "Service admin role for Monitoring access. Assign only to service users",
+             "name": "monitoring:service-admin"},
+            {"id": "100000004",
+             "serviceId": "test-eabb70a0e702a4626977c331d5c4",
+             "description": "Admin Role.",
+             "name": "identity:admin"},
+            {"id": "10000005",
+             "serviceId": "test-d4614b87411e141fe8109099bc4f",
+             "description": "Role to access Customer service as an Admin",
+             "name": "customer:admin"}],
+        "name": name
+    }
 
 
 def get_token(tenant_id,
--- a/mimic/canned_responses/auth.py
+++ b/mimic/canned_responses/auth.py
@@ -25,8 +25,8 @@ def format_timestamp(dt):
 
 
 def get_token(tenant_id,
-              entry_generator,
-              prefix_for_endpoint,
+              entry_generator=None,
+              prefix_for_endpoint=None,
               timestamp=format_timestamp,
               response_token=HARD_CODED_TOKEN,
               response_user_id=HARD_CODED_USER_ID,
@@ -61,7 +61,7 @@ def get_token(tenant_id,
                 "endpoints": list(endpoint_json())
             }
 
-    return {
+    response = {
         "access": {
             "token": {
                 # TODO: This token should be synthesized and stored in an
@@ -72,7 +72,6 @@ def get_token(tenant_id,
                     "id": tenant_id,
                     "name": tenant_id},
                 "RAX-AUTH:authenticatedBy": ["PASSWORD"]},
-            "serviceCatalog": list(entry_json()),
             "user": {
                 "id": response_user_id,
                 "name": response_user_name,
@@ -80,6 +79,10 @@ def get_token(tenant_id,
             }
         }
     }
+
+    if (entry_generator and prefix_for_endpoint):
+        response["access"]["serviceCatalog"] = list(entry_json())
+    return response
 
 
 def get_endpoints(tenant_id, entry_generator, prefix_for_endpoint):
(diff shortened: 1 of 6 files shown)


glyph:
(Restarted the Travis build due to intermittent connection timeout on docs build.)

mimic/rest/auth_api.py
```diff
@@ -135,17 +135,42 @@ def get_impersonation_token(self, request):
         except ValueError:
             request.setResponseCode(400)
             return json.dumps(invalid_resource("Invalid JSON request body"))
-
+        impersonator_token = request.getHeader("X-Auth-Token")
```
jjbuchan: Is `getHeader`case sensitive?

mimic/test/test_auth.py
```diff
@@ -774,3 +788,121 @@ def test_impersonation_request_with_invalid_body_causes_http_bad_request(self):
...
+            }
+        ))
+        self.assertEqual(200, response1.code)
+        # Impersonate user test1
+        (response2, json_body2) = self.successResultOf(json_request(
+            self, root, "POST",
+            "http://mybase/identity/v2.0/RAX-AUTH/impersonation-tokens",
+            {"RAX-AUTH:impersonation": {"expire-in-seconds": 1,
+                                        "user": {"username": "test1"}}},
+            headers={'X-Auth-Token': [str(impersonator_token)]}
```
jjbuchan: Why would `x-auth-token` be an array?

lekhajee: `request` which is a wrapped by `json_request` in the `helpers.py` expects the headers to be an array.

mimic/rest/auth_api.py
lekhajee: yes, it is not case sensitive. I will change that to be lower case.

mimic/test/test_auth.py
jjbuchan: Oh yeah, I see it ends up using this - http://twistedmatrix.com/documents/10.1.0/api/twisted.web.http_headers.Headers.html#setRawHeaders

Much easier to track down after I cloned this repo.

mimic/session.py
```diff
@@ -120,9 +124,9 @@ def session_for_token(self, token, tenant_id=None):
             if tenant_id is not None and s.tenant_id != tenant_id:
                 raise NonMatchingTenantError(session=s,
                                              desired_tenant=tenant_id)
-            return s
-
-        return self._new_session(token=token, tenant_id=tenant_id)
+        else:
+            s = self._new_session(token=token, tenant_id=tenant_id)
+        return s, self._impersonated_token_to_impersonator_session.get(token)
```
glyph: I don't really like changing the signature of `_new_session` (and thus all of its callers) to a 2-tuple; tuples are much less self-describing than (characteristic) structures with attributes, so unless it's really obvious which is the first thing and which is the second thing in the return value, it's much better to have attributes with names that at least imply what they are.

mimic/session.py
```diff
@@ -84,7 +88,7 @@ def _new_session(self, username_key=None, **attributes):
                  with this :obj:`MimicCore`).
         :rtype: :obj:`Session`
         """
-        for key in ['username', 'token', 'tenant_id']:
+        for key in ['username', 'token', 'tenant_id', 'impersonated_token']:
```
glyph: By having `impersonated_token` be a mappable attribute of the session like this, that means that a session has *one and only one* impersonated token.  This is wrong, because multiple admins might impersonate a given user; each impersonation is its own session, with its own token.

That's why the "subsession" was there before.  An impersonation session should share the `api_objects` attribute (i.e. "the tenant's actual stuff"), but should not share the metadata about the session (the token, the expiration time).

mimic/test/test_auth.py
```diff
@@ -14,6 +14,7 @@
 
 
 class ExampleCatalogEndpoint(object):
+
```
glyph: Please fix your autopep8 to not introduce this blank-line noise :).

(Also, since you apparently did do this already, can you share how you configured it?)

mimic/test/test_auth.py
glyph: Glad you were able to find that object :-). Although we're using 14.0+ so more properly it would be http://twistedmatrix.com/documents/14.0.0/api/twisted.web.http_headers.Headers.html#setRawHeaders … :)

glyph: Conceptually, it's expected to be a list because a header might occur multiple times and have multiple values, so using a list allows us to be unambiguous that it only occurs once.

mimic/test/test_auth.py
lekhajee: I added the E309 to the ignore list.

lekhajee:
This has been tested against the api-client tests on the https://github.com/racker/ele/pull/2775.

ynachiket:
This looks good to me 
+1

mimic/canned_responses/auth.py
```diff
@@ -24,9 +25,41 @@ def format_timestamp(dt):
     return dt.strftime('%Y-%m-%dT%H:%M:%S.999-05:00')
 
 
+def impersonator_user_role(id, name):
```
glyph: There should be a ticket for getting rid of this and replacing it with a more "real" role collection in the identity model.

mimic/canned_responses/auth.py
```diff
@@ -81,6 +113,10 @@ def endpoint_json():
         }
     }
 
+    if (entry_generator and prefix_for_endpoint):
```
glyph: Anywhere that you use a raw object in an `if` statement, you're calling its `__nonzero__` method, which can have side-effects.  If the only thing you want to test about an object is identity with `None`, it should just be `is None`, since that's a tighter contract; it's more specific.  So I think this should be `if entry_generator is not None and prefix_for_endpoint is not None`.

mimic/rest/auth_api.py
```diff
@@ -151,17 +153,45 @@ def get_impersonation_token(self, request):
...
+                                                               impersonator_token,
+                                                               impersonated_token)
+
         return json.dumps({"access": {
-            "token": {"id": session.token,
+            "token": {"id": impersonated_token,
                       "expires": format_timestamp(session.expires)}
         }})
 
+    @app.route('/v2.0/tokens/<string:token_id>', methods=['GET'])
```
glyph: I think we might want to consider starting to include links in each docstring to the openstack API being emulated.  In this case, http://developer.openstack.org/api-ref-identity-v2.html#admin-tokens

mimic/rest/auth_api.py
```diff
@@ -151,17 +153,45 @@ def get_impersonation_token(self, request):
...
+    def validate_token(self, request, token_id):
+        """
+        Creates a new session for the given tenant_id and token_id
+        and always returns response code 200.
+        """
+        request.setResponseCode(200)
+        tenant_id = request.args.get('belongsTo')
+        if tenant_id:
```
glyph: This should also be `is None` I think; unless it's valid to have an empty `&belongsTo=&...` in the URL?

mimic/rest/auth_api.py
```diff
@@ -151,17 +153,45 @@ def get_impersonation_token(self, request):
...
+            tenant_id = tenant_id[0]
+        session = self.core.sessions.session_for_tenant_id(tenant_id, token_id)
+        response = get_token(
+            session.tenant_id,
+            response_token=session.token,
+            response_user_id=session.user_id,
+            response_user_name=session.username,
+        )
+        for each_session in session.impersonator_session_list:
```
glyph: `session_impersonator_list` should be a straight-up dictionary, not a list of dictionaries; each `(impersonator, impersonatee)` tuple only gets a single token (so there should only ever be one entry in each one of the dictionaries in the list) and the token must be globally unique (not even just within the tenant, but within all of identity, because the lookup on the token starts at the top of identity with no other context).

Also, the lookup here ought to be a method on the session; `session.impersonator_session_for_token(impersonated_token_id)`?

mimic/test/test_auth.py
```diff
@@ -809,3 +819,218 @@ def test_impersonation_request_with_invalid_body_causes_http_bad_request(self):
...
+            "http://mybase/identity/v2.0/tokens/{0}?belongsTo=12345".format(
+                impersonated_token)
+        ))
+        self.assertEqual(200, response3.code)
+        self.assertTrue(json_body3["access"]["RAX-AUTH:impersonator"])
+
+    def test_response_for_validate_impersonated_token_multiple_users(self):
+        """
+        Test to verify :func: `validate_token` and then authenticate
+        """
```
glyph: This test method is reaaaaaaaally long; I notice that there are a lot of lines which are basically just repeated with different values.  Could you put some of this stuff into a function, maybe even just an inner function to start with, to cut down on the length?

glyph:
Please address the few comments above and then merge; they're small enough that I don't think this needs another review pass.

PR #173, opened 2015-02-10
Fix osx pypy on travis

derwolfe:
Use pyenv instead of brew to install pypy. The reason pyenv's pypy was not working is that pyenv wasn't being reinitialized in the run.sh.

### DIFF
--- a/.travis/install.sh
+++ b/.travis/install.sh
@@ -29,8 +29,9 @@ if [[ "$DARWIN" = true ]]; then
             pyenv global 2.7.8
             ;;
         pypy)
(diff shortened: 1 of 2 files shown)


derwolfe:
This patch should also make the build process a bit simpler. Instead of relying on brew for some of the python installs and pyenv for the others; all of the osx builds that need to install python would do so through pyenv. 

This also has the benefit of removing any upstream problems that brew might be having with  sourceforge.

glyph:
Looks good to me; nice that we can unify this.

PR #175, opened 2015-02-12
initial error injection for nova

glyph:
This establishes a general-purpose URI hierarchy for manipulating behaviors, and a single implementation of one behavior - "fail to create server" - which can be registered ahead of time instead of using metadata.

We will need follow-on work to refactor this to be more general, but this should be good enough to get started with.

### DIFF
--- a/mimic/model/nova_behaviors.py
+++ b/mimic/model/nova_behaviors.py
@@ -97,14 +97,6 @@ def regexp_predicate(value):
     return re.compile(value).match
 
 
-def tenant_id_criterion(value):
-    """
-    Return a Criterion which matches the given regular expression string
-    against the ``"tenant_id"`` attribute.
-    """
-    return Criterion(name='tenant_id', predicate=regexp_predicate(value))
-
-
 def server_name_criterion(value):
     """
     Return a Criterion which matches the given regular expression string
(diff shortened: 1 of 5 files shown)


glyph:
@tokunbo - if you could review this when the builds finish that would be great.

glyph:
(Wow, travis is taking its sweet time. Half an hour to even start the build?)

tokunbo:
Wow advanced code happening here. I've never seen the @attributes thing and I see you're returning "function pointers" too. Gimme some time to digest what's happening here ^_^;

tokunbo:
Is this like a closure? https://github.com/rackerlabs/mimic/pull/175/files#diff-08737134bad520aee53b307e41669a38R44

glyph:
Yup! They're closures :).

tokunbo:
+1 ....I'm not claiming I understand everything here 100%, but I get the idea. ^^;

glyph:
Ugh, 3 failed builds. There must be some way to tell travis to automatically retry when certain steps fail…

derwolfe:
@glyph - I've not tried it yet, but it looks like travis supports a retry command: 

http://docs.travis-ci.com/user/build-timeouts/

The syntax seems to be like 

```
install: travis_retry bundle install
```

glyph:
Now that coverage is back up and the tests are passing, I'm going to take @tokunbo's review as authoritative, so it's his fault if something goes wrong :smirk_cat:

PR #180, opened 2015-02-14
retry builds on travis

derwolfe:
use travis's ```retry_build``` command. It is meant to try to rerun a step a maximum of three times if any error condition in the step is met.

The functionality is detailed here: http://docs.travis-ci.com/user/build-timeouts/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,7 +31,7 @@ install:
   - travis_retry ./.travis/install.sh
 
 script:
-  - ./.travis/run.sh
+  - travis_retry ./.travis/run.sh
 
 after_success:
   - source ~/.venv/bin/activate && coveralls
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,7 +28,7 @@ matrix:
       compiler: clang
 
 install:
-  - ./.travis/install.sh
+  - travis_retry ./.travis/install.sh
 
 script:
   - ./.travis/run.sh


derwolfe:
While the build for this *is* passing, I didn't see it needing to retry the build.

glyph:
So, the problem is that I don't actually want to re-try the whole build; if the *tests themselves* fail, even due to an intermittent issue, I really want to see that as a failure; we shouldn't be dependent on external network resources during the tests.  It's the installation of the dependencies I want to retry.

`tox` can *just* create virtual environments via `tox --recreate --notest`. I haven't figured out how to tell it to skip dependency installation *entirely* to explicitly separate those steps, but hopefully the pip download cache will prevent that (IIRC @dstufft said something about it only trying every request once every 24h).

dstufft:
It'll only re-fetch index pages (assuming ``CacheControl`` headers are set up) every 10 minutes (we cap index pages to 10 minutes cached). For file downloads we it'll cache it for however long the file's ``CacheControl`` headers says to do, on PyPI that's a very long time. In both cases it'll attempt to do conditional GETs to refresh the cache instead of re-download.

glyph:
#181 makes the changes I'm proposing; hopefully that will work nicely.

dstufft:
Should work.

glyph:
I'm going to close this in favor of #181; hopefully that will be mergeable after the latest blast of builds have some results to report.  Thanks for getting this started though @derwolfe !

derwolfe:
@glyph You're welcome!

PR #181, opened 2015-02-16
Retry just install

glyph:
Create the virtual environments for tox in a separate explicit step.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,7 +31,7 @@ install:
   - travis_retry ./.travis/install.sh
 
 script:
-  - ./.travis/run.sh
+  - travis_retry ./.travis/run.sh
 
 after_success:
   - source ~/.venv/bin/activate && coveralls
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,7 +31,7 @@ install:
   - travis_retry ./.travis/install.sh
 
 script:
-  - travis_retry ./.travis/run.sh
+  - ./.travis/run.sh
 
 after_success:
   - source ~/.venv/bin/activate && coveralls
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,7 +28,7 @@ matrix:
       compiler: clang
 
 install:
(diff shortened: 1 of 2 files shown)


glyph:
I will push a bunch of dummy commits so that we can visually see if retrying makes it more reliable in the commit list.

glyph:
So, https://travis-ci.org/rackerlabs/mimic/jobs/50889597 is confusing me. It appears that `tox --recreate --notest` *succeeds* (from what the summary says), so there's no re-trying.  When I break my network locally though, I see the script properly exit with an error.

glyph:
This depends on #182 which should probably be merged promptly.

glyph:
That travis job's issue appears to be that `installdeps` errors are correctly reported but `inst` errors aren't.

glyph:
Got a little tired of coveralls being so chatty.  Now that it has status-API support,

coveralls-comments

glyph:
I looked at a whole lot of builds in those green checkmarks, and I wasn't able to find anywhere that a retry had obviously happened. Maybe `travis_retry` obscures the failures? Anyway, 8 consecutive successes on all builds seems to be an indication that this doesn't do any harm…

derwolfe:
@glyph - When I was testing ```travis_retry``` it would show each successive retry of the step that failed. If it failed more than three times it would error out completely. I'll look through the old travis builds for #180 as it had an failing ```run.sh``` step that was retried.

derwolfe:
After looking through the logs for #180, I can't find the build log displaying the behavior I've described. I'm betting that I restarted the build, and therein destroyed the log history.

glyph:
Augh: https://travis-ci.org/rackerlabs/mimic/jobs/50992835

glyph:
Why won't this fail when we want it to :)

glyph:
Oh right, that link will shortly be meaningless, because travis deletes history when you rebuild…

lekhajee:
+1

PR #182, opened 2015-02-16
Pin development dependencies

glyph:
We just started failing builds due to a new release of pep8. Deal with that release and ignore the warning it introduced, and lock all transitive dependencies in tox and dev-requirements.txt to avoid this kind of surprise failure in the future.

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -1,7 +1,20 @@
+cffi==0.8.6
+coverage==3.7.1
+cryptography==0.7.2
 doc8==0.3.4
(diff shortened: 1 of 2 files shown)


glyph:
This depends on #181 (the branch is merged in here) so even upon a successful review we might not want to merge it.

glyph:
@lekhajee @derwolfe - a quick review on this would be good, pretty much everything is going to fail travis until this lands :).

derwolfe:
@glyph - this looks good to me; please merge it in.

PR #183, opened 2015-02-20
Randomize administrative password for new server creation.

### LINKED ISSUE #164: randomize admin passwords
Right now the root password for servers is always hard-coded to `"testpassword"`, this should instead be random per server.

sam-falvo:
Fixes #164

### DIFF
--- a/mimic/test/test_nova.py
+++ b/mimic/test/test_nova.py
@@ -78,6 +78,27 @@ class NovaAPITests(SynchronousTestCase):
         )
         validate_link_json(self, self.create_server_response_body['server'])
 
+    def test_created_servers_have_dissimilar_admin_passwords(self):
+        """
+        Two (or more) servers created should not share passwords.
+        """
+        create_server = request(
+            self, self.root, "POST", self.uri + '/servers',
+            json.dumps({
+                "server": {
+                    "name": self.server_name,
+                    "imageRef": "test-image",
+                    "flavorRef": "test-flavor"
+                }
+            }))
+        other_response = self.successResultOf(create_server)
+        other_response_body = self.successResultOf(
+            treq.json_content(other_response))
+        self.assertNotEqual(
+            self.create_server_response_body['server']['adminPass'],
+            other_response_body['server']['adminPass']
+        )
+
     def test_list_servers(self):
         """
         Test to verify :func:`list_servers` on ``GET /v2.0/<tenant_id>/servers``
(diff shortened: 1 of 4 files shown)


mimic/test/test_nova.py
```diff
@@ -131,6 +135,8 @@ def test_get_server(self):
                          self.server_id)
         self.assertEqual(
             get_server_response_body['server']['status'], 'ACTIVE')
+        admin_password = get_server_response_body['server'].get('adminPass', None)
```
cyli: Nice test!  Definitely shouldn't be giving out admin passwords in getting info.

cyli:
Two nitpicks:

1.  Can we add a test to make sure that two created servers do not have the same admin password?
2.  Can we add one or more tests for the randomized string generation function?

I think the helper tests reside in https://github.com/rackerlabs/mimic/blob/master/mimic/test/test_util.py.

Other than that, this PR looks pretty great.  :)

mimic/util/helper.py
```diff
@@ -15,6 +15,25 @@
...
+
+    :param int length: How long the string must be.
+    :param str selectable: If left unspecified, the random character selection
+        will be taken from uppercase and lowercase letters, digits, and a few
+        punctuation marks.  Otherwise, the characters will be taken from the
+        string provided.
+    :returns: A string of length `length`.
+    """
+    selectable = selectable or (
+        "abcdefghijklmnopqrstuvwxyz"
```
cyli: Nonblocking - this is totally fine, and in fact may be more clear in some ways, but you can also use `strings.ascii_lowercase + strings.ascii_uppercase` (or just `strings.ascii_letters`), plus `strings.digits` :)

I'm not entirely sure about the symbols - I see that there exists a `strings.punctuation`, but not sure what that means - probably listing all the symbols is a better idea.

Anyway, totally nonblocking.  :)

mimic/test/test_util.py
```diff
@@ -83,3 +83,27 @@ def test_unicode_body(self):
...
+    def test_selectable(self):
+        """
+        When passing a custom selectable, the results should derive only from
+        the characters you provide.
+        """
+        desired_chars = "02468"
+        for iteration in xrange(100):
+            a_string = helper.random_string(1024, selectable=desired_chars)
+            for char in a_string:
+                self.assertFalse(char not in desired_chars)
```
cyli: Non-blocking style nitpick:  the double negative may be a little harder to read than just a positive assertion :)

cyli:
Up to you what you want to do about the non-blocking nitpicks :)  Otherwise LGTM if travis approves.

mimic/util/helper.py
sam-falvo: Cool, wasn't aware of this.

sam-falvo:
I take your advice seriously, even the non-blockers.  Thanks!

cyli:
Thanks!  Again, LGTM if travis +1's :)

PR #184, opened 2015-02-20
Respect manually provided OS-DCF:diskConfig.

### LINKED ISSUE #163: honor OS-DCF:diskConfig in server creation
Right now you always get a disk config of `"AUTO"` no matter what you specified.

sam-falvo:
Fixes #163


cyli:
This LGTM as soon as travis passes.  

I was also wondering if we wanted to check that the value is either "AUTO" or "MANUAL" and return a 400 if it isn't, as per http://docs.rackspace.com/servers/api/v2/cs-devguide/content/CreateServers.html.

I don't think it'd be much of an addition, but it can also be a different PR entirely if you like - if you don't want to do it in this one, either you or I could file a new issue and it can be handled later.

mimic/test/test_nova.py
```diff
@@ -39,6 +39,27 @@ def setUp(self):
...
+                    "imageRef": "test-image",
+                    "flavorRef": "test-flavor",
+                    "OS-DCF:diskConfig": "MANUAL",
+                }
+            }))
+        create_server_response = self.successResultOf(create_server)
+        response_body = self.successResultOf(
+            treq.json_content(create_server_response))
+        self.assertEqual(
+            response_body['server']['OS-DCF:diskConfig'], 'MANUAL')
```
glyph: It would be good to also assert something about the `disk_config` attribute itself, or, failing that, the results of a subsequent `GET` request; this test could be made to pass by simply returning a different static value for the create response and *not* properly updating Mimic's internal state to reflect that server's disk configuration.

mimic/rest/nova_api.py
```diff
@@ -219,8 +219,16 @@ def create_server(self, request, tenant_id):
...
             return json.dumps(invalid_resource("Invalid JSON request body"))
 
-        return (self._region_collection_for_tenant(tenant_id)
-                .request_creation(request, content, self.url))
+        try:
+            creation = (self._region_collection_for_tenant(tenant_id)
+                        .request_creation(request, content, self.url))
+        except ValueError:
+            request.setResponseCode(400)
+            return json.dumps(
```
cyli: In this case, the nova error when I tried it looks like it's:

```
{
    "badRequest": {
        "message": "OS-DCF:diskConfig must be either 'MANUAL' or 'AUTO'.",
        "code": 400
    }
}
```

Looks like the `invalid_resource` for the invalid JSON is also wrong.  Would you mind just changing the message to reflect the actual nova message, and I'll file another ticket to fix the use of invalid_resource for both this and the JSON message?

cyli: Filed:  https://github.com/rackerlabs/mimic/issues/186

Sorry, realized I wasn't super clear, I just meant change "OS-DCF:diskConfig is neither AUTO nor MANUAL" to "OS-DCF:diskConfig must be either 'MANUAL' or 'AUTO'.".

Someone else can come along and swap out the function later.  :)

mimic/test/test_nova.py
sam-falvo: Hi Glyph; can you go into more detail about what you have in mind here?  I'm asking because the production code is affecting internal state to the best of my knowledge.

mimic/rest/nova_api.py
sam-falvo: These are addressed in patches #c683a9e and #88ac3f7 .

cyli: Awesome, thanks!  :D

cyli:
This LGTM modulo handling of @glyph's comment and a +1 by travis.

mimic/test/test_nova.py
glyph: Sure @sam-falvo.

I'm just talking about the thoroughness of the test, not the correctness of the implementation.

This test is asserting only that the HTTP response to the creation `POST` looks a certain way; it doesn't assert anything about subsequent requests.  So an implementation which simply passed the input through to the output but *didn't* store the disk configuration anywhere internally would still pass the test.  There doesn't seem to be any other test verifying this either, so an implementation change which breaks in this way wouldn't be caught.

mimic/test/test_nova.py
```diff
@@ -39,6 +39,45 @@ def setUp(self):
...
+
+    def test_create_server_with_bad_diskConfig(self):
+        """
+        Given a diskConfig not in ["AUTO", "MANUAL"], we expect a 400
+        Bad Request error.
+        """
```
glyph: The test docstring idiom I tend to prefer would phrase this as an action verb where the system under test is the subject, such as:

```python
"""
When ``create_server`` is passed an invalid ``OS-DCF:diskConfig`` that is not `"AUTO"` or `"MANUAL"`,
it returns an HTTP status code of 400 (bad request).
"""
```

mimic/test/test_util.py
```diff
@@ -83,3 +83,9 @@ def test_unicode_body(self):
         self.failureResultOf(
             request(self, Resource(), b"POST", b"", u"not bytes")
         )
+
+    def test_bad_request(self):
```
glyph: This test should have a docstring explaining what it's testing.  I am not quite clear on the purpose of `bad_request` - is this the proper format of the actual response returned by Nova when selecting an invalid disk config?  (You wouldn't happen to have a documentation reference there, would you?  It would be great to link that from the test.)

sam-falvo: No documentation reference exists, because the error format is completely undocumented in the Nova docs.  All we know is that a 400 result happens.  Thus, Mimic is mirroring Rackspace's implementation of Nova, but it's not guaranteed to match, e.g., Nebula's implementation.

mimic/test/test_nova.py
sam-falvo: Fixed in 271b07ec8225e03c630cffa58642891310c3aad9

mimic/test/test_util.py
sam-falvo: Fixed in 271b07ec8225e03c630cffa58642891310c3aad9

cyli:
Apologies, one more merge forward, and afterwards please merge?

PR #107, opened 2014-11-11
WIP - Add Support to mimic CDN providers

malini-kamalambal:
This PR adds support to mimic CDN provider API calls. While this might not fit in the scope of Mimic (since this does not mock Openstack endpoints, but is aimed at mocking Fastly API calls), the intent of this PR is to get feedback on if this is the right approach. If this turns out to fit into the Mimic context & can be merged, I can clean up this PR & add tests.

PR #189, opened 2015-03-01
[RFR] Adds Fastly Mock

malini-kamalambal:
This PR adds Mimic support for some Fastly endpoints & corresponding tests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,8 +115,13 @@ node, for 20 seconds:
 
 ## Fastly CDN ##
 
+Fastly is one of the leading CDN providers with great API support.
+Fastly is also one of the CDN providers, supported by the Openstack Poppy
+project. Mimic supports all Fastly API calls needed by Openstack Poppy in its
+Fastly implementation. See below for the complete list.
+
 #### Calls supported: ####
-https://github.com/rackerlabs/mimic/blob/master/mimic/rest/fastly_api.py
+https://github.com/rackerlabs/m imic/blob/master/mimic/rest/fastly_api.py
 
  1. GET /current_customer
  2. POST /service
--- a/README.md
+++ b/README.md
@@ -129,6 +129,7 @@ https://github.com/rackerlabs/mimic/blob/master/mimic/rest/fastly_api.py
  9. PUT /service/{service_id}/version/{version_number}/activate
  10. DELETE /service/{service_id}
  11. GET /service/{service_id}/details
+ 12. GET / (health check)
 
 ## Mimic Control APIs ##
 
--- a/README.md
+++ b/README.md
@@ -113,6 +113,23 @@ node, for 20 seconds:
 
 `{"loadBalancer": {"name": "a-new-loadbalancer2", "protocol": "HTTP", "virtualIps": [{"type": "PUBLIC"}], "metadata": [{"key": "lb_pending_update", "value": 20}], "nodes": []}}`
 
+## Fastly CDN ##
+
+#### Calls supported: ####
+https://github.com/rackerlabs/mimic/blob/master/mimic/rest/fastly_api.py
+
+ 1. GET /current_customer
+ 2. POST /service
+ 3. POST /service/{service_id}/version
+ 4. GET /service/search
+ 5. POST /service/{service_id}/version/{version_id}/domain
+ 6. GET /service/{service_id}/version/{version_id}/domain/check_all
+ 7. POST /service/{service_id}/version/{version_id}/backend
+ 8. GET /service/{service_id}/version
+ 9. PUT /service/{service_id}/version/{version_number}/activate
+ 10. DELETE /service/{service_id}
+ 11. GET /service/{service_id}/details
+
 ## Mimic Control APIs ##
 
 When any of Mimic's included plugins schedule a timeout, you will need to cause
--- a/mimic/canned_responses/fastly.py
+++ b/mimic/canned_responses/fastly.py
@@ -1,0 +1,273 @@
+"""
+Canned response for fastly
+"""
+import uuid
+
+
+class FastlyResponse(object):
+
+    """
+    Canned response for fastly.
+    See docs here https://docs.fastly.com/api/config
+    """
+
+    fastly_cache = {}
+
+    def get_current_customer(self, request):
+        """
+        Returns the current customer with response code 200.
+
+        :return: a JSON-serializable dictionary matching the format of the JSON
+                 response for fastly_client.get_current_customer()
+                 ("/current_customer") request.
+        """
+        current_customer = {
+            u'can_edit_matches': u'0',
+            u'can_read_public_ip_list': u'0',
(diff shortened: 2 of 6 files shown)


glyph:
Since it's ready for review it should probably be looked at by someone other than @malini-kamalambal :)

README.md
```diff
@@ -113,6 +113,24 @@ node, for 20 seconds:
 
 `{"loadBalancer": {"name": "a-new-loadbalancer2", "protocol": "HTTP", "virtualIps": [{"type": "PUBLIC"}], "metadata": [{"key": "lb_pending_update", "value": 20}], "nodes": []}}`
 
+## Fastly CDN ##
```
glyph: A reference to what Fastly is would be helpful here, I think.

mimic/canned_responses/fastly.py
```diff
@@ -0,0 +1,278 @@
...
+
+    def get_current_customer(self):
+        """
+        Returns the current customer with response code 200.
+
+        :return: a JSON-serializable dictionary matching the format of the JSON
+                 response for fastly_client.get_current_customer()
+                 ("/current_customer") request.
+        """
+        current_customer = {
```
glyph: This whole dictionary is built without any reference to `self`, so defining it outside - as a class variable, say - would make it clearer that we are not actually building anything based on the current object just yet.  As it was I had to carefully look through to see if `self` was being used.

mimic/canned_responses/fastly.py
```diff
@@ -0,0 +1,278 @@
...
+            u'has_config_panel': u'1',
+            u'has_improved_ssl_config': False,
+            u'id': u'3bKRHISECARTtoUMxW6',
+            u'has_historical_stats': u'1',
+            u'has_openstack_logging': u'0',
+            u'can_configure_wordpress': u'0',
+            u'has_improved_logging': u'1',
+            u'readonly': '',
+            u'ip_whitelist': u'0.0.0.0/0',
+            u'owner_id': u'asewQWEwdRnLnCKyMztZqkQvy',
```
glyph: This looks … suspiciously specific, as does the `'id'` above.  I think it might be a good idea to generate some random IDs here?  Normally I would say this could follow on in a separate change, but if we don't do that in the first revision where this is added, then tests may start to depend on these hard-coded IDs.

mimic/canned_responses/fastly.py
```diff
@@ -0,0 +1,278 @@
...
+                 response for fastly_client.create_service()
+                 ("/service") request.
+        """
+        data = dict((key, value[0]) for key, value in url_data)
+
+        publish_key = uuid.uuid4().hex
+        service_id = uuid.uuid4().hex
+        service_name = data['name']
+
+        self.fastly_cache[service_name] = {
```
glyph: This looks like a copy of the recently-removed bad pattern for Nova, where everything was a "cache" and stored in a global variable.  The thing stored here should probably evolve to be some kind of class instance, which can do JSON serialization as *one* of its tasks.

It's fine to put this off for a future revision, though.

mimic/canned_responses/fastly.py
```diff
@@ -0,0 +1,278 @@
...
+                           u'deployed': None,
+                           u'backend': 0}],
+            u'created_at': u'2014-11-12T18:43:21+00:00',
+            u'updated_at': u'2014-11-12T18:43:21+00:00',
+            u'customer_id': data['customer_id'],
+            u'id': service_id}
+        return create_service
+
+    def get_service_by_name(self, service_name):
+        """Returns service details json.
```
glyph: Could you write slightly more descriptive docstrings here?  I don't know what "service details json" means.

mimic/test/test_fastly.py
```diff
@@ -0,0 +1,240 @@
...
+            self, self.root, "POST",
+            self.uri + '/service?customer_id={0}&name={1}'.format(
+                self.customer_id, self.service_name)))
+
+        self.service_id = self.service_json['versions'][0]['service_id']
+        self.version_id = self.service_json['versions'][0]['number']
+
+    def test_get_customer(self):
+        """
+        Test to verify : ``GET /current_customer``
```
glyph: For test docstrings, I really prefer a style where it's a declarative statement about the functionality being tested.  For example,

```
``GET /current_customer`` against the fastly API mock returns JSON-serialized customer details.
```

This says not only that it's verifying `/current_customer`, but *what* about `/current_customer` is being verified.  This is helpful as APIs grow more behaviors over time, because if an endpoint has 5 or 6 different behaviors depending on its state, a lot of them just end up being `verify GET /current_customer when stuff is different`, which makes a maintainer's job difficult because they have to decipher what is expected to happen when stuff is "normal" and when it is "different" and how those differences arise.

glyph:
@malini-kamalambal - thanks for your contribution to mimic :).

The main comment that is preventing me from hitting the "merge" button right away is really the hard-coded IDs in the mock.  I think improving the documentation just a little bit to make it clear what is and isn't being mocked about Fastly so far would also be helpful, but for the other stuff, you can either address it directly or just file an issue.

lekhajee:
Just noting that this branch was once - https://github.com/rackerlabs/mimic/pull/107

mimic/canned_responses/fastly.py
malini-kamalambal: @glyph, In the context of Openstack Poppy (which is going to be the main/only(?) consumer of Fastly Mimic), there will be only a single owner. Poppy enables provisioning on CDN providers using a single Fastly account. That being said it is really easy for me to update this to generate a unique ID & tht  will still align with Poppy's use of Mimic . Let me know what you think.

mimic/canned_responses/fastly.py
malini-kamalambal: @glyph, can I open an issue for this after this gets merged?

mimic/canned_responses/fastly.py
malini-kamalambal: will do :)

mimic/canned_responses/fastly.py
glyph: Generating a unique ID per mimic run would be fine.

mimic/canned_responses/fastly.py
glyph: @malini-kamalambal Absolutely.

mimic/canned_responses/fastly.py
glyph: Thanks!

malini-kamalambal:
@glyph, @lekhajee, I think I have all of the review comments addressed now. Can you please take a look when you get a chance?

glyph:
I think this is good enough to get started - hopefully you can file some of those follow-up issues now though :-D.

malini-kamalambal:
woohooo..thanks @glyph I will file the follow up issues to address some of ur comments.

PR #190, opened 2015-03-03
RFR: add .coverage and _trial_temp to .gitignore

ksheedlo:
These files are automatically generated by the build and cause the git
state to appear dirty. We should explicitly ignore them.


lekhajee:
+1

PR #191, opened 2015-03-04
Mac app redux

glyph:
Generate a mac application automatically on travis, using `--user` installs to work around virtualenv recipe bugs.


glyph:
OK I think this fixes all the issues with #103 , at least functionally.

To build this locally, I do `PIP_USER=yes bash -x build-app.sh` and it seems to work OK.  Tests run, they pass, the app launches.  On Travis we have an uglier situation where it has to use homebrew python simply because they've installed a bunch of packages into `/Library` on the system python and thereby broken it somehow, but the build is still basically functional. When Travis gets a yosemite builder we might be able to be a bit cleaner about it.

@derwolfe - can you have a look?

bundle/start-app.py
```diff
@@ -0,0 +1,115 @@
...
+        self.menubarMenu = NSMenu.alloc().init()
+        self.menubarMenu.addItem_(self.port)
+        self.menubarMenu.addItem_(self.quit)
+
+        # add the menu to status bar item
+        self.statusItem.setMenu_(self.menubarMenu)
+        self.statusItem.setToolTip_(u"mimic - rackspace mock api")
+
+        AppHelper.callLater(1, startMimic)
+        # XXX I"m continuing to get an exception here
```
derwolfe: It seems like this comment of mine isn't necessary - should it be removed?

docs/development/getting-started.rst
```diff
@@ -84,6 +84,66 @@ Alternately, you can use our ``tox`` job:
...
+The officially supported method of building of the application depends on the
+system python, `pyobjc`_, and `py2app`_ libraries installed with Mac OS.
+`Travis-CI`_ is configured to build the mac application and run its tests.
+
+To build the application and run its tests locally use the following commands.
+
+.. code-block:: console
+
+   $ cd /dir/where/mimic/lives/
+   $ ./build-app.sh
```
derwolfe: The directions here could be changed  to 

```
$ cd /dir/where/mimic/lives
$ PIP_USER=yes bash -x build-app.sh
```

docs/development/getting-started.rst
```diff
@@ -84,6 +84,66 @@ Alternately, you can use our ``tox`` job:
...
+``Quit``. You can view the application logs by opening
+``Applications/Utilities/Console.app``.
+
+To run ``mimic.app``'s tests use
+
+.. code-block:: console
+
+   $ /path/to/mimic.app/Contents/MacOS/run-tests
+
+The application can also built as a standalone application
```
derwolfe: Should we remove the docs detailing how to build a standalone application? The reason I'd suggest we do so, is the directions aren't tested by travis.

setup.py
```diff
@@ -1,23 +1,126 @@
...
+        },
+        options={
+            "py2app": {
+                "includes": [
+                    "syslog",
+                    "mimic.test.*",
+                    "mimic.plugins.*",
+                    "twisted.plugins.*",
+                    "twisted.plugin",
+                ],
```
derwolfe: Should these keys be present in the `py2app options` dictionary?
```
                "semi_standalone": True,
                "site_packages": True
```
When I built the application last night, this solved the issue of the bundle not being able to runtime.

derwolfe:
@glyph Outside of the comments I've made, I don't see anything else that I feel needs to change. 

I was able to build the application locally, but it complained that it wasn't able to find a python runtime. This persisted after I made the changes to `py2app` options stating 

```
semi-standalone: True,
site-packages: True
```
I believe my errors are related to my build environment and not the application (my reason for thinking so is that Travis can build the app successfully).

derwolfe:
I did just notice that is using a brew python. If so, then using 
```
semi-standalone: True,
site-packages: True
```
wouldn't be necessary, as this isn't a semi-standalone; it's packaging the brewed interpreter with it.

glyph:
@derwolfe - I am trying to disregard your comment, but I also can't find 

```
semi-standalone: True,
site-packages: True
```

anywhere in the code. or a mention of semi-standalone anymore anywhere.  also that's invalid Python syntax *and* invalid shell syntax, so I don't know where to look for it :).

setup.py
glyph: When I upgrade to py2app 0.9 - which meant I had to stop using virtualenvs because of the bug I referenced on #103 - I stopped having issues with the semi-standalone application not being able to find the system Python.  I also tried setting the `PyInterpreterLocations` key, which *did* work, but struck me as a more potentially fragile solution.

glyph:
There are two remaining issues here:

  1. how do we get builds *out* of Travis?
  2. (spooky voice) codesigning

I think we might have to have someone (and by someone I probably mean me) get an apple developer account and start signing the application so people can run it on machines with default gatekeeper security.

docs/development/getting-started.rst
glyph: The *only* thing tested by travis is building a standalone application now, isn't it? :smile:

setup.py
```diff
@@ -1,23 +1,126 @@
...
+                    path_on_fs = (
+                        os.path.abspath(
+                            os.path.join(
+                                os.path.dirname(
+                                    __import__(what + ".plugins",
+                                               fromlist=["nonempty"])
+                                    .__file__),
+                                "dropin.cache")
+                        ))
+                    os.utime(path_on_fs, (time.time() + 86400,) * 2)
```
glyph: This probably deserves a comment.  Surprised nobody else called me out on it yet :)

docs/development/getting-started.rst
derwolfe: Good point! If the application being distributed *is* always built by travis using the brewed python, should we remove the documentation detailing how to build using the system python?

setup.py
derwolfe: I think I glossed over this piece of code instead of making a comment - sorry about that. I do agree that a comment describing its purpose would be helpful.

derwolfe:
To get builds out of travis, it looks like travis can push build artifacts to S3. The docs detailing how to do so are at http://docs.travis-ci.com/user/deployment/s3/

docs/development/getting-started.rst
glyph: Building with the system python will produce a much smaller app (3.3M vs. 11M) so it seems like it might be worth keeping around, especially given that one of the selling points of Mimic is its small footprint :).

bundle/start-app.py
glyph: Yeah it should be.  And I'm also not entirely clear why the `AppHelper.callLater` is necessary, come to mention it.

glyph:
I guess we can do codesigning last and tell our first batch of users to just download it over https and dequarantine it

ksheedlo:
Holy commit log Batman! Can we squash some of these down?

glyph:
@ksheedlo - No. That's what merge commits are for :).  The history here is valuable, because someone working on it can see what previous attempts were tried.  Just the epic length of it shows you that there is something fairly nuanced going on :).

Squashing history is a bad habit that pretty much only exists to work around deficiencies in various log viewers (chiefly Github's, because it doesn't let you fold or elide merge commits easily).  One day those log viewers will get better, and if we've squashed out all the history in the meanwhile, we won't have useful data for them to inspect.  In the meanwhile, there are good log viewers like SourceTree which let you untangle long branch development histories.

lekhajee:
+1 . Works like a charm!!!

PR #192, opened 2015-03-04
make sure to require twisted >= 14.0.2 for security especially now that ...

glyph:
...the mac app might be doing --user installs sometimes


glyph:
Um… did I do something wrong on this PR? Travis doesn't seem to see it.

lekhajee:
+1

PR #193, opened 2015-03-05
Default to empty lists for any missing objects in createSuppression

tokunbo:


lekhajee:
Can you add a test to verify this.

mimic/rest/maas_api.py
```diff
@@ -210,6 +210,14 @@ def createSuppression(params):
         if 'encode' in dir(params[k]):
             params[k] = params[k].encode('ascii')
     params['id'] = 'sp' + random_hex_generator(4)
+    if 'notification_plans' not in params:
```
glyph: Coveralls is passing this because it doesn't test line coverage, but it's worth noting that we need to have tests for every invisible "`else`" clause here :).

lekhajee:
+1

PR #197, opened 2015-03-10
RFR: Noit mimic

lekhajee:
Implements the following reconnoiter api calls. (http://labs.omniti.com/labs/reconnoiter). The documentation for reconnoiter is poor and most of this derived from Ele's implementation and some testing against a running instance of noit. (Readme on https://github.com/virgo-agent-toolkit/noit_client attempts to point at some documentation related to the Noit api calls Ele uses)

Used the old cache model to store api data as wanted to spend most of my time on trying to render xml. The object model change will be done as part of a different PR.


mimic/canned_responses/noit.py
```diff
@@ -0,0 +1,123 @@
+"""
+Canned response for Noit
+"""
+import time
+from mimic.canned_responses.noit_metrics_fixture import (metrics_common_template,
+                                                         metrics)
+
+
+noit_cache = {}
```
glyph: nope nope nope

We don't have a session to store per-user information on.  That's... weird, and I don't know how noit manages to do what it does without authentication. However, there's a relatively simple process for de-`_cache`-ifying this and making it so we don't depend on global state everywhere:

  1. take every function in this module that touches `noit_cache` and put them as methods on an object
  2. instantiate that object from within `NoitApi` and store it as an object on `self`

We can figure out how noit does authentication and sessions later, but when we do, we can then just take that object and stick it in a session, rather than having to change all the code that touches this global.

mimic/rest/noit_api.py
```diff
@@ -0,0 +1,114 @@
...
+import xmltodict
+from uuid import UUID
+from twisted.web.server import Request
+from mimic.rest.mimicapp import MimicApp
+from mimic.canned_responses.noit import (create_check, get_check,
+                                         get_all_checks, delete_check,
+                                         test_check)
+
+
+# TO DO:
```
glyph: please file issues for these and remove the comment, the bugtracker should have all the stuff we want to do, not the code :). If you think that this would be useful for someone reading the code later, just include links to the github issues.

mimic/rest/noit_api.py
```diff
@@ -0,0 +1,114 @@
...
+# Move away from using dict for the templates
+# Include all check type to the metrics in the fixtures
+# Metrics Fixture to populate as per test check request
+# Include more error cases based on ele's tests
+
+
+Request.defaultContentType = 'application/xml'
```
glyph: This changes the default content type of *all* HTTP responses in this entire process.  Instead, just set the `application/xml` content type explicitly in each place you're returning a response.

glyph: See #202

glyph:
`xmltodict` is a little bit of a weird choice, not what I would have picked, but it meshes nicely with the way that JSON responses are being synthesized elsewhere in the code.  I think that's it for my comments, so please address them and resubmit.

mimic/canned_responses/noit.py
lekhajee: sorry, I will do this as a separate PR.

lekhajee:
yes, thanks to @cyli for bringing `xmltodict` to my attention. I reviewed it for our purposes and it seemed to suit noit's needs.

mimic/rest/noit_api.py
lekhajee: done in f7c9e66

mimic/canned_responses/noit.py
glyph: the issue is #203

PR #200, opened 2015-03-13
new plugins and altered responses to existing plugins for reach

tawalton:
This PR serves as a record of *all* the code changes necessary to get the cloud control panel tests passing, and should be closed those changes have been split out into smaller units and merged.


glyph:
Since I think we've discussed this in various in-person discussions, just to make it clear to onlookers; this PR is probably not going to be merged as-is; we are keeping it open as a record of additional work that needs to be split out into smaller PRs and merged.

glyph:
I guess we've even got a label for that now.

glyph:
I will update the description to record its purpose as well. Which is weird because it puts my words into @tawalton's comment box but so it goes :)

mimic/rest/maas_api.py
```diff
@@ -573,35 +568,41 @@ def overview(self, request, tenant_id):
...
-            v['latest_alarm_states'] = []
-            values.append(v)
-        request.setResponseCode(200)
-        return json.dumps({'metadata': metadata, 'values': values})
+        # This is a hack to make cloud control panel server details page finish loading the server
+        # details
+        if 'uri' in request.args:
+            request.setResponseCode(404)
+            return json.dumps({})
+        else:
```
ksheedlo: I have a couple of concerns here.

1. Why do we need to 404 in this case at all? Successfully getting the overview response shouldn't cause a page to hang. Is this a bug in Reach? and if so, can you post a gist or similar explaining it?
2. Since you return at the end of the if block, you don't need the else block and re-indentation. I'd prefer you not indent everything over again if you don't have to.

ksheedlo:
Is this PR still being worked on?

tawalton:
@ksheedlo This PR is not intended to be reviewed or merged.  This PR serves as a record of the code changes necessary to get the cloud control panel to create servers through the ui and run some of the servers tests. It should be closed once these  changes have been split out into smaller units and merged.

glyph:
I think this is too out of date to be of use for tracking purposes any more.  I'm going to close for now.  Reopen whenever this effort starts up again…

PR #201, opened 2015-03-17
Add an integrationy test to make sure impersonation works for autoscale

cyli:
Autoscale tests recently failed due to impersonation bugs.  Added a test to ensure that after impersonating a user, the tenant ID returned for the impersonatee is the same as the tenant impersonated.

Test currently fails.

Also, 'authenticate' is mispelled as 'autheticate' in those awesome helper functions :)

### DIFF
--- a/mimic/test/test_auth.py
+++ b/mimic/test/test_auth.py
@@ -963,6 +963,6 @@ class AuthIntegrationTests(SynchronousTestCase):
 
         # authenticate with this token and see what the tenant is
         response, json_body = autheticate_with_token(
-            self, core, token_id=token)
+            self, core, token_id=token, tenant_id=tenant_id)
         self.assertEqual(tenant_id,
                          json_body['access']['token']['tenant']['id'])
--- a/mimic/test/test_auth.py
+++ b/mimic/test/test_auth.py
@@ -931,9 +931,12 @@ class AuthIntegrationTests(SynchronousTestCase):
         root = MimicRoot(core).app.resource()
         tenant_id = "111111"
 
-        # authenticate as that user
+        # authenticate as that user - this is not strictly necessary, since
+        # getting a user for a tenant should work regardless of whether a user
+        # was previously in the system, but this will ensure that we can check
+        # the username
         response, json_body = autheticate_with_username_password(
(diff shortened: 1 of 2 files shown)


mimic/test/test_auth.py
```diff
@@ -914,3 +914,55 @@ def test_response_for_validate_impersonated_token_multiple_users(self):
...
+        response, json_body = self.successResultOf(json_request(
+            self, root, "GET",
+            "/identity/v2.0/tokens/{0}/endpoints".format(token)))
+        self.assertEqual(200, response.code)
+        self.assertEqual(tenant_id,
+                         json_body["endpoints"][0]["tenantId"])
+
+        # authenticate with this token and see what the tenant is
+        response, json_body = autheticate_with_token(
+            self, core, token_id=token)
```
lekhajee: `autheticate_with_token` allows for `token_id` and `tenant` to be optional and defaults it to a random token and tenant. For the purpose of this test we want to pass the tenant_id else it will auth with the token of tenant '111111' with the tenant as '12345' resulting in the error `{u'unauthorized': {u'message': u"Token doesn't belong to Tenant with Id/Name: '12345'", u'code': 401}}`

lekhajee:
Sorry, fixing impersonate to include the session associated with the impersonation_token to `_token_to_session`. And the misspelt `autheticate`

mimic/test/test_auth.py
cyli: Ah, ok, thank you!  :)  This one isn't actually used by autoscale, I just thought I'd include it for completeness.

lekhajee:
@cyli : I fixed them in this same PR. Hope that is ok :)

cyli:
@lekhajee That's awesome, thank you!  I hadn't had time to look into why it was failing yet, just wanted to get a test demonstrating the error.  :)

Thanks for fixing so fast!

lekhajee:
@cyli: ah, ok. Sorry I was out yesterday and missed a the slack notifications about this bug. And welcome :)

glyph:
LGTM, thanks for moving so quickly everyone.

PR #206, opened 2015-03-20
fix to correspond create entity to be as requested in the request payloa...

lekhajee:
...d + lint

Earlier create_entity just passed along the entity name and no other attributes from the request payload when creating a new entity. This resulted in the entity label same as the request but all other attributes to be defaults.

### DIFF
--- a/mimic/rest/maas_api.py
+++ b/mimic/rest/maas_api.py
@@ -65,7 +65,7 @@ class MaasApi(object):
         return MaasMock(self, uri_prefix, session_store, region).app.resource()
 
 
-class M_Cache(dict):
+class MCache(dict):
 
     """
     M(onitoring) Cache Object to hold dictionaries of all entities, checks and alarms.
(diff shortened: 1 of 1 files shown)


mimic/rest/maas_api.py
```diff
@@ -337,7 +338,7 @@ def create_entity(self, request, tenant_id):
         """
         postdata = json.loads(request.content.read())
         myhostname_and_port = 'http://' + request.getRequestHostname() + ':' + self.endpoint_port
-        newentity = createEntity({'label': postdata[u'label'].encode('ascii')})
+        newentity = create_entity(postdata)
```
lekhajee: This is the major change. Earlier `create_entity` just passed along the entity name and no other attributes from the request payload.

tokunbo:
aaaad

lekhajee:
@tokunbo : thanks :)

PR #209, opened 2015-03-26
RFR: fix(session): use existing session for same tenant

### LINKED ISSUE #205: MAAS API fails to persist resources across sessions for the same tenant
When I create resources against my tenant on Mimic, I expect that these resources will be available to other clients logged in as the same tenant. Running against the tip on Mimic, I can see this is not the case.

```
2015-03-19 15:34:00-0700 [HTTPChannel,52,127.0.0.1] 127.0.0.1 - - [19/Mar/2015:22:34:00 +0000] "POST /mimicking/MaasApi-d8e5f0/ORD/v1.0/921648/entities HTTP/1.1" 201 - "-" "-"
// 998 more create entity calls from a custom API client...
2015-03-19 15:34:00-0700 [HTTPChannel,55,127.0.0.1] 127.0.0.1 - - [19/Mar/2015:22:34:00 +0000] "POST /mimicking/MaasApi-d8e5f0/ORD/v1.0/921648/entities HTTP/1.1" 201 - "-" "-"
2015-03-19 15:34:17-0700 [HTTPChannel,57,127.0.0.1] 127.0.0.1 - - [19/Mar/2015:22:34:16 +0000] "POST /identity/v2.0/tokens HTTP/1.1" 200 1859 "-" "-"
2015-03-19 15:34:17-0700 [HTTPChannel,58,127.0.0.1] 127.0.0.1 - - [19/Mar/2015:22:34:16 +0000] "GET /mimicking/MaasApi-d8e5f0/ORD/v1.0/921648/__experiments/json_home HTTP/1.1" 200 30431 "-" "-"
// No entities are returned to Cloud Intelligence in either /entities or /views/overview
2015-03-19 15:34:17-0700 [HTTPChannel,59,127.0.0.1] 127.0.0.1 - - [19/Mar/2015:22:34:17 +0000] "GET /mimicking/MaasApi-d8e5f0/ORD/v1.0/921648/entities?limit=1000 HTTP/1.1" 200 111 "http://localhost:3000/" "Rackspace Cloud Intelligence; username_939f5f10-617a-4390-9739-ed7165a716b5; 921648"
2015-03-19 15:34:17-0700 [HTTPChannel,60,127.0.0.1] 127.0.0.1 - - [19/Mar/2015:22:34:17 +0000] "GET /mimicking/MaasApi-d8e5f0/ORD/v1.0/921648/views/overview?limit=1000 HTTP/1.1" 200 111 "http://localhost:3000/" "Rackspace Cloud Intelligence; username_939f5f10-617a-4390-9739-ed7165a716b5; 921648"
```

I consider this a severe bug that makes it impossible for me to use Mimic at the current version. We should fix this as soon as possible.

ksheedlo:
Fixes #205.

### DIFF
--- a/mimic/session.py
+++ b/mimic/session.py
@@ -74,8 +74,8 @@ class SessionStore(object):
         self._userid_to_session = {
             # mapping of userid (ascii) to session (Session)
         }
-        self._tenant_to_token = {
-            # mapping of tenant_id (unicode) to token (unicode)
+        self._tenant_to_session = {
+            # mapping of tenant_id (unicode) to session (Session)
         }
         self._username_to_token = {
             # mapping of token (unicode) to username (unicode: key in
@@ -115,8 +115,12 @@ class SessionStore(object):
         self._username_to_token[username_key] = session.token
         self._token_to_session[session.token] = session
         self._userid_to_session[session.user_id] = session
-        self._tenant_to_token[session.tenant_id] = session.token
+        self._tenant_to_session[session.tenant_id] = session
         return session
+
+    def _assert_tenant_matches(self, session, tenant_id):
+        if tenant_id is not None and session.tenant_id != tenant_id:
+            raise NonMatchingTenantError(session=session, desired_tenant=tenant_id)
 
     def session_for_token(self, token, tenant_id=None):
         """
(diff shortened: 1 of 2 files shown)


glyph:
Can you add a test case for this?  I don't think all of the changes are necessary to get the desired behavior (particularly I still think there is a 1-to-1 "token" to "session" mapping, so the change to make it `tenant_to_session` instead of `tenant_to_token` is unnecessary) but I want to make sure to validate that against a test.

glyph:
(Also, you know, coveralls complaining makes it so we can't land it without the test. :smile: )

ksheedlo:
I have all green locally and full coverage, so this should be good to go soon.

ksheedlo:
> I still think there is a 1-to-1 "token" to "session" mapping

This is actually not the case. When using Cloud Intelligence and authenticating through Reach, I don't have control over the token (it comes from real Identity) and can't use the same token that an external API client used. We need `tenant_to_session` because of the scenario where the tenant is common, but the token is unpredictable due to indirection.

lekhajee:
+1 lgtm. It would be great to also add an integration test like https://github.com/rackerlabs/mimic/blob/master/mimic/test/test_auth.py#l925.

ksheedlo:
Should I address that in a separate PR?

lekhajee:
sure. will do.

PR #211, opened 2015-04-01
creation behavior to have a server go into error state after some time

### LINKED ISSUE #210: behavior that adds a timed failure to make a server go from active to error
Sometimes (very rarely!) an active server goes into error without warning.  This can be caused by certain manual operations (password reset) or may happen spontaneously, presumably due to hypervisor events we have not fully classified yet.

We should have a behavior which uses the virtual clock to set a timeout and cause a server go into error.

glyph:
Fixes #210


lekhajee:
+1 LGTM

PR #212, opened 2015-04-06
add nodes returns 202, not 200

radix:
In production I've only seen CLB returning 202 from the add nodes call, not 200. It seems that at some point in time the documentation was erroneously describing a response of 200, but it was corrected to 202. 

http://docs.rackspace.com/loadbalancers/api/v1.0/clb-devguide/content/POST_createLoadBalancer_v1.0__account__loadbalancers_load-balancers.html#comment-1677370185


lekhajee:
+1

radix:
Thanks Lekha :)

PR #214, opened 2015-04-16
RFR: fix(maas_api): agent_host_info 404 for missing entity

ksheedlo:
Mimic currently throws a 500 when calling `/views/agent_host_info` with a nonexistent entity ID. This change fixes Mimic to return the 404, matching the result from Cloud Monitoring API.

### DIFF
--- a/mimic/rest/maas_api.py
+++ b/mimic/rest/maas_api.py
@@ -642,10 +642,21 @@ class MaasMock(object):
                                'details': 'You must specify an agentId, entityId, or an entity URI.',
                                'mimicNotes': 'But mimic will only accept entityId right now',
                                'txnId': '.fake.mimic.transaction.id.c-1111111.ts-123444444.v-12344frf'})
+
         entity_id = request.args['entityId'][0].strip()
         for e in self._entity_cache_for_tenant(tenant_id).entities_list:
             if e['id'] == entity_id:
                 agent_id = e['agent_id']
+                break
+        else:
+            request.setResponseCode(404)
+            return json.dumps({'type': 'notFoundError',
+                               'code': 404,
+                               'message': 'Object does not exist',
+                               'details': 'Object "Entity" with key "{0}" does not exist'.format(
+                                   entity_id),
+                               'txnId': '.fake.mimic.transaction.id.c-1111111.ts-123444444.v-12344frf'})
+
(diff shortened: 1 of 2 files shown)


glyph:
Hi @ksheedlo,

Thanks for the PR.

It looks like the added unit test is failing on travis, and there's a minor lint error (line too long).

Also, it seems like the implementation here could be a lot simpler and still make the given unit test pass - I think it would be worthwhile to assert about some of the included keys (if they're important) or drop them (if they aren't).

I'll try to re-review if you can address this stuff.

ksheedlo:
Oopsie. Wasn't able to test locally because of dev env problems. Will try
again.

ksheedlo:
Taking a second look at this this morning.

> I think it would be worthwhile to assert about some of the included keys (if they're important) or drop them (if they aren't).

I'm not sure what you mean by this. Can you go into more detail, or give a code example?

ksheedlo:
The most recent revision should pass tests.

mimic/rest/maas_api.py
```diff
@@ -642,10 +642,21 @@ def view_agent_host_info(self, request, tenant_id):
                                'details': 'You must specify an agentId, entityId, or an entity URI.',
                                'mimicNotes': 'But mimic will only accept entityId right now',
                                'txnId': '.fake.mimic.transaction.id.c-1111111.ts-123444444.v-12344frf'})
+
         entity_id = request.args['entityId'][0].strip()
         for e in self._entity_cache_for_tenant(tenant_id).entities_list:
             if e['id'] == entity_id:
                 agent_id = e['agent_id']
+                break
+        else:
```
lekhajee: the `if` block then the code outside the block and then the `else` is making this code not easy to read and comprehend.

ksheedlo: Can you suggest an alternative way of laying it out?

lekhajee: On second thought it is absolutely fine as is. The alternative way would be,
```
entities = [e for e in self._entity_cache_for_tenant(tenant_id).entities_list if e['id'] == entity_id]
if entities:
    agent_id = entities[0]['agent_id']
else:
    request.setResponseCode(404)
    ....
```
I would even consider having a function for the list comprehension as it seems like something that could be reused.

lekhajee:
+1 lgtm modulo the comment, which may or may not be addressed :)

PR #215, opened 2015-04-22
Make delete LB return actually empty responses, not `""`

radix:
This enables API + backend code to represent truly empty responses with an `EMPTY_RESPONSE` constant and a new utility function `json_dump`, which honors `EMPTY_RESPONSE` to return an empty string instead of a json-serialized empty string.

It also makes the delete-LB call use this new facility. There are lots of other cases where it can be used, but this is just a first pass.

### DIFF
--- a/mimic/canned_responses/loadbalancer.py
+++ b/mimic/canned_responses/loadbalancer.py
@@ -5,7 +5,8 @@ add/get/delete/list nodes
 """
 from random import randrange
 from copy import deepcopy
-from mimic.util.helper import (not_found_response, invalid_resource,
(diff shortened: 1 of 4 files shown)


lekhajee:
+1 thanks

radix:
Should I just ignore the two travis jobs? They were both timeouts, and I've run 'lint' locally and it's passed.

radix:
never mind, I reran them and they passed :)

radix:
Should I merge this? (not sure how the usual review/merge process goes in the mimic project)

lekhajee:
yes, once you have a +1 and the builds pass, it is fine to merge.

PR #216, opened 2015-04-23
bulk-delete nodes from CLB

### LINKED ISSUE #213: Support CLB Bulk Delete
http://docs.rackspace.com/loadbalancers/api/v1.0/clb-devguide/content/DELETE_bulkDeleteNodes_v1.0__account__loadbalancers__loadBalancerId__nodes_Nodes-d1e2173.html

Basically, DELETE /.../nodes?id=a&id=b&id=c

radix:
Fixes #213.

This introduces support for the `DELETE .../loadbalancers/{lb_id}/nodes?id=1&id=2` endpoint for deleting multiple nodes at a time.

### DIFF
--- a/mimic/canned_responses/loadbalancer.py
+++ b/mimic/canned_responses/loadbalancer.py
@@ -220,10 +220,24 @@ def get_nodes(store, lb_id, node_id, current_timestamp):
     return not_found_response("loadbalancer"), 404
 
 
+def _delete_node(store, lb_id, node_id):
+    """Delete a node by ID."""
+    if store.lbs[lb_id].get("nodes"):
+        for each in store.lbs[lb_id]["nodes"]:
+            if each["id"] == node_id:
+                index = store.lbs[lb_id]["nodes"].index(each)
+                del store.lbs[lb_id]["nodes"][index]
+                if not store.lbs[lb_id]["nodes"]:
+                    del store.lbs[lb_id]["nodes"]
+                store.lbs[lb_id].update({"nodeCount": len(store.lbs[lb_id].get("nodes", []))})
+                return True
+    return False
+
+
 def delete_node(store, lb_id, node_id, current_timestamp):
     """
-    Determines whether the node to be deleted exists in mimic store and
-    returns the response code.
+    Determines whether the node to be deleted exists in mimic store, deletes
+    the node, and returns the response code.
     """
     if lb_id in store.lbs:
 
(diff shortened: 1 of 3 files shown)


lekhajee:
The coverage seems to have decreased by .05%. https://coveralls.io/builds/2402623/source?filename=mimic%2Fcanned_responses%2Floadbalancer.py

radix:
Oh, sorry about that. Thanks for the note @lekhajee, I'll add a test to cover that case :)

radix:
Okay, I've added a test for the uncovered case and also merged from master so the diff should not have the EMPTY_RESPONSE changes any more.

mimic/canned_responses/loadbalancer.py
```diff
@@ -231,28 +245,68 @@ def delete_node(store, lb_id, node_id, current_timestamp):
...
+            "code": 400}
+        return resp, 400
+
+    if lb_id not in store.lbs:
+        return not_found_response("loadbalancer"), 404
+
+    _verify_and_update_lb_state(store, lb_id, False, current_timestamp)
+
+    if store.lbs[lb_id]["status"] != "ACTIVE":
+        resp = {"message": "LoadBalancer is not ACTIVE",
```
cyli: Just double-checking - is the not active message for bulk actions different than the not active message for singular actions?  You are probably correct since you've also corrected the error message for `delete_node`, but just wanted to make sure.

radix: Yep, this is exactly the message that I got from the real CLB. Thanks for double checking :)

mimic/test/test_loadbalancer.py
```diff
@@ -517,10 +555,97 @@ def test_delete_non_existant_node_on_loadbalancer(self):
...
+        """
+        response, body = _bulk_delete(
+            self, self.root, self.uri, self.lb_id, [])
+        self.assertEqual(response.code, 400)
+        self.assertEqual(
+            body,
+            {'code': 400,
+             'message': "Must supply one or more id's to process this request."})
+
+    def test_bulk_delete_no_nodes_invalid_lb(self):
```
cyli: +1 nice test case!

mimic/canned_responses/loadbalancer.py
cyli: :|  wish their messages were more standardized - do you want to leave a comment here in case someone else comes through and decides to make the two messages consistent?

mimic/test/test_loadbalancer.py
```diff
@@ -587,6 +712,15 @@ def _delete_loadbalancer(self, lb_id):
                             str(lb_id))
         return self.successResultOf(delete_lb)
 
+    def _create_loadbalancer(self, metadata):
```
cyli: +1 thank you for the refactor :)

mimic/test/test_loadbalancer.py
```diff
@@ -733,3 +851,49 @@ def test_delete_load_balancer_and_pending_delete_state(self):
...
+        self.assertEqual(updated_lb["loadBalancer"]["status"], "PENDING-UPDATE")
+
+        # Now, trying to bulk-delete nodes (including invalid ones) will cause
+        # it to return the special error
+        response, body = _bulk_delete(
+            self, self.root, self.uri, lb['id'], [100, 200])
+        self.assertEqual(response.code, 422)
+        self.assertEqual(
+            body,
+            {u'message': u'LoadBalancer is not ACTIVE', u'code': 422})
```
cyli: Non-blocking - would it be overkill to add an assertion that the one valid node is not deleted?

cyli:
This LGTM modulo a comment (which you may or may not choose to add :))

mimic/test/test_loadbalancer.py
radix: This case doesn't even pass a valid node ID at all to the deletion request, so I don't think it's useful. Also there's already a test case that ensures that passing some valid and some invalid IDs to a bulk deletion doesn't delete the valid ones.

mimic/canned_responses/loadbalancer.py
radix: Sure thing, I've added it :) thanks

PR #220, opened 2015-04-25
RFR: test(auth): multiple tokens per session

ksheedlo:
This change is a follow up to #209. It adds an integration test that was
requested in code review.


ksheedlo:
I don't know if I can verify that both tokens are accessing the same session given the current dummy `ExampleAPI`. See the comment included in the diff.

@lekhajee do you have any ideas on what we should do here?

lekhajee:
@ksheedlo : We can verify that the `username` in the service catalog match.

lekhajee:
+1, good to merge once the build on travis is successful :)

ksheedlo:
cool, thanks!

PR #221, opened 2015-04-25
RFR: Fix the 404 not found response from Nova to include a body

cyli:
As opposed to None.

### DIFF
--- a/mimic/model/nova_objects.py
+++ b/mimic/model/nova_objects.py
@@ -13,6 +13,7 @@ from six import string_types
 from mimic.util.helper import (
     seconds_to_timestamp,
     invalid_resource,
+    not_found_response,
     random_string,
 )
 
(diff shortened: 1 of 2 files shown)


mimic/model/nova_objects.py
```diff
@@ -499,7 +500,7 @@ def request_read(self, http_get_request, server_id, absolutize_url):
         server = self.server_by_id(server_id)
         if server is None:
             http_get_request.setResponseCode(404)
-            return None
+            return dumps(not_found_response('servers'))
```
manishtomar: Non-blocking and not as part of this PR: Since `not_found_response` is always going to be accompanied by 404, is it possible to set response code and returning response in one function call?

manishtomar:
LGTM

mimic/model/nova_objects.py
cyli: Good point, and maybe this should be split up by service anyway.  I'll file an issue.  :)

PR #227, opened 2015-04-28
Start glance plugin - GET images call

tawalton:


mimic/rest/glance_api.py
```diff
@@ -0,0 +1,76 @@
...
+    app = MimicApp()
+
+    @app.route('/v2/<string:tenant_id>/images', methods=['GET'])
+    def get_images(self, request, tenant_id):
+        """
+        Returns a list of glance images. Currently there is no provision
+        for shared versus unshared images in the response
+        """
+        request.setResponseCode(200)
+        return json.dumps(get_images())
```
lekhajee: shouldn't this return a list ?

tawalton: Yes, it should, but after talking with @glyph we decided to start with one image for the sake of landing this PR. We will need to add additional images in the future.  There is an earlier PR that had all the current images included: https://github.com/rackerlabs/mimic/blob/plugins_for_reach/mimic/canned_responses/json/glance/glance_images_json.py

lekhajee: right, even with the one image, I would think it would exist with the `images` list object.

tawalton: I am not following what you are saying. Can you provide an example or we can discuss off line?

glyph: I *think* what Lekha is saying refers to the structure of the data which `get_images()` returns, but this line of code is a bit removed from the problem, so I'll put the comment there :)

mimic/canned_responses/images_json.py
```diff
@@ -0,0 +1,51 @@
+"""
+Glance images
+"""
+
+images = {
+    "images":
+        {
```
glyph: This should be a list; i.e.

`"images": {...}`

should be

`"images": [{...}]`.

lekhajee:
+1

lekhajee:
@tawalton : Needs a rebase to be merged.

glyph:
In the future, I'd appreciate it if we could merge master rather than rebasing pushed branches.

PR #234, opened 2015-04-29
Change server status

### LINKED ISSUE #233: Support changing server status in Nova via the control plane
This would fall under the umbrella of #229.

This is needed by autoscale tests to do more specific error injection - to ensure that if some servers goe into error status, that convergence will notice and self-heal.

glyph:
Fixes #233


tawalton:
:+1:

lekhajee:
+1, lgtm

PR #235, opened 2015-04-29
Make builds faster by building fewer things on multiple platforms.

glyph:


glyph:
" ran for 35 min 45 sec" says travis, which suggests we have cut the build time approximately in half… this is still kinda slow, but unfortunately we do have to keep mac app and mac commandline test coverage running if we want to support those environments :-\

lekhajee:
+1

PR #236, opened 2015-04-30
Add nova control plane api to the plugins and update test

lekhajee:
Addresses #237


cyli:
LGTM, thanks @lekhajee!

PR #238, opened 2015-05-01
Support pagination in Nova when listing servers

### LINKED ISSUE #231: Support pagination when listing servers
This is required by Autoscale.

cyli:
Fixes #231.

### DIFF
--- a/mimic/model/nova_objects.py
+++ b/mimic/model/nova_objects.py
@@ -37,6 +37,41 @@ class BadRequestError(Exception):
     """
 
 
+def _nova_error_message(msg_type, message, status_code, request):
+    """
+    Set the response code on the request, and return a JSON blob representing
+    a Nova error body, in the format Nova returns error messages.
+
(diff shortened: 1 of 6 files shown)


mimic/model/nova_objects.py
```diff
@@ -515,21 +551,77 @@ def request_ips(self, http_get_ips_request, server_id):
...
         """
         Request the list JSON for all servers.
 
         Note: only supports filtering by name right now, but will need to
         support more going forward.
-        """
-        return dumps(
-            {"servers": [
+
+        Pagination behavior verified against Rackspace Nova as of 2015-04-29.
```
glyph: Minor nitpick; this should probably be "Rackspace Cloud Servers" rather than "Rackspace Nova"; "Nova", as an openstack thing, sort of implicitly means "not Rackspace specific".

glyph:
Amazing test coverage, every case looks clearly covered, all tests passing... I don't have a whole lot to add. LGTM.

PR #239, opened 2015-05-01
Session-based role API

### LINKED ISSUE #232: Role API based on sessions
Found in https://github.com/rackerlabs/mimic/issues/228#issuecomment-97248245

lvh:
Closes #232.

### DIFF
--- a/mimic/model/identity.py
+++ b/mimic/model/identity.py
@@ -1,0 +1,21 @@
+"""
+Models relating to identity.
+"""
+from attr import attributes, attr
+
+
+@attributes
+class IdentitySession(object):
+    """
+    An identity session for a tenant.
+    """
+    @classmethod
+    def from_store(cls, session_store, tenant_id):
+        """
+        Get the identity session for the given tenant.
+        """
+        return (session_store
+                .session_for_tenant_id(tenant_id)
+                .data_for_api(cls, cls))
+
+    identity_admin = attr(default=False)
--- a/mimic/model/identity.py
+++ b/mimic/model/identity.py
@@ -1,7 +1,7 @@
 """
 Models relating to identity.
 """
-from attr import attributes, attr
+from attr import attributes, attr, validators
 
 
 @attributes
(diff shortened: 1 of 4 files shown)


glyph:
If this is RFR, could you add the "ready for review" label?

mimic/model/identity.py
```diff
@@ -0,0 +1,22 @@
+"""
+Models relating to identity.
+"""
+from attr import attributes, attr, validators
```
glyph: We haven't had a great process for vetting new requirements, so I wouldn't reject this patch based on the addition of `attrs` alone, but it does seem like if you'd like to land this, we should at least have a plan (filed issues) to migrate to `attrs` rather than depend on both it and `characteristic` long-term.

requirements.txt
```diff
@@ -5,3 +5,4 @@ treq==15.0.0
 characteristic==14.3.0
 six==1.6.1
 xmltodict==0.9.1
+attrs==15.0.0
```
glyph: As the failed travis builds indicate, the real source of truth for the builders is `install_requires` in `setup.py` rather than `requirements.txt`.

mimic/model/identity.py
lvh: I'm still trying to figure out how this project's process works, but I think the appropriate thing is an umbrella issue, so I filed it: https://github.com/rackerlabs/mimic/issues/241

lvh: If you've looked at the ticket before now, I just added all of the places where it would need to get changed.

glyph: We're all still trying to figure out the process a little bit :-). Thanks for the umbrella issue.

glyph:
This needs a merge of master to merge cleanly again.

lvh:
This is ready for review, so I've added the label. Thanks for the pointer, @glyph.

requirements.txt
lvh: (Resolved, see setup.py below)

glyph:
Looking good.  Thanks for the patch!

PR #240, opened 2015-05-03
Run in Docker as non-privileged user

ksheedlo:
This change improves Docker support in particular with respect to the security limitations brought up in #95. We run as a non-privileged user in a basic Debian Jessie.

This change does not address the issue of leaving the buildchain lying around. I'd be afraid of breaking something :)


glyph:
> This change improves Docker support in particular with respect to the security limitations brought up in #95. We run as a non-privileged user in a basic Debian Jessie.

Docker already has a command-line flag to select a user; the `-u` option to `docker run`.  One issue I have with the way that user-ID management works in Docker is that there's no clear way to map between users on the hosts and users in the guest.  So, by running as uid 1000 we still have whatever privileges we have as that user on the host.

Might it make more sense to run as the `nobody` user, who might at least correspond to the local `nobody` on the host, and that doesn't require creating new users with `adduser`?

> This change does not address the issue of leaving the buildchain lying around. I'd be afraid of breaking something :)

One change at a time :)

ksheedlo:
I'm very new to Docker, so I was expecting to get a lot of things wrong.

> there's no clear way to map between users on the hosts and users in the guest. So, by running as uid 1000 we still have whatever privileges we have as that user on the host.

This is news to me. I thought uid's were fully virtualized.

> Might it make more sense to run as the nobody user

Agreed, but Twisted wants to create a pidfile. I'm exploring ways to get around that, as nobody is not allowed to create files.

ksheedlo:
Update: I can make Twisted drop privileges to nobody, but then it can't delete the pidfile. This doesn't seem to be a problem, though, because it won't commit changes to the filesystem.

glyph:
> I'm very new to Docker, so I was expecting to get a lot of things wrong.

I should not overstate my level of experience with Docker either.  As far as I know there is no "best practice" in this area yet.  If you use a `USER` line, you're hard-coding some configuration which might not match with the host.  If you don't, you risk the default execution user being root, which carries additional risks.

> > there's no clear way to map between users on the hosts and users in the guest. So, by running as uid 1000 we still have whatever privileges we have as that user on the host.
> This is news to me. I thought uid's were fully virtualized.

It's complicated. The contained user has all the privileges of the host user, except of course, they're in a container, so the container *process* doesn't have privileges on the host.  The reason to be concerned about user IDs is that if there's a container-escape bug, you escape with the privileges you currently hold.

For example, if a volume is mounted inside the container, and the contained process writes a file to it, the UID on the filesystem shows up as the numerical UID from inside the container.  When you think about it, there's no obvious way *to* "fully virtualize" that case; the filesystem is going to get a numerical UID from somewhere, that is going to map to permissions of some host user, so unless dockerd understands the user configuration of every possible host OS and the tools to add/remove/modify users (which are definitely different between distributions) it won't be able to keep things up to date.

> > Might it make more sense to run as the nobody user
> Agreed, but Twisted wants to create a pidfile. I'm exploring ways to get around that, as nobody is not allowed to create files.

Running as root and then dropping privileges is actually worse.  With no change, as the code is on master, you can just run it with the `-u` option and get it to run as an appropriately unprivileged user.  With the change in this PR, we now *require* root privileges in the container, just so we can drop privileges (the `setuid` call to drop privileges is itself privileged).  If we just specify a `USER` line, that can at least be overridden by the docker administrator.

ksheedlo:
well then I don't think this is any better that what we already have now ¯\\\_(ツ)\_/¯ Ops are hard.

If you don't have any better ideas @glyph I'm inclined to close the PR.

glyph:
@ksheedlo Perhaps let's close this PR since this doesn't seem to be an especially fruitful avenue of investigation - but I think maybe we should open an issue to address this so that we can attract some attention from someone more knowledgeable about docker ops than either of us and a future PR can close that.

glyph:
Actually yeah I'll go do that right now.

glyph:
Thanks for raising the question @ksheedlo. We can discuss further on #243 and hopefully get some more input.

Issue #222, opened 2015-04-27
Split up `helpers.not_found_response` by service


cyli:
Each service may just want a helper function that helps them form the JSON bodies for errors, since each service may have errors in a different format.  But they seem to all be something like:

```
{ 
   "somekey": {
       "message": ...,
       "code": <http status code>
  }
}
````

So the helper function can also take a request and set the status code too.

PR #242, opened 2015-05-04
RFR: Use nova specific error body generation functions

cyli:
Like nova specific not-found, bad request, forbidden, etc. functions that also set the HTTP code.  Part of #222.

### DIFF
--- a/mimic/model/nova_objects.py
+++ b/mimic/model/nova_objects.py
@@ -20,7 +20,7 @@ from mimic.util.helper import (
 from mimic.model.behaviors import (
     BehaviorRegistry, EventDescription, Criterion, regexp_predicate
 )
-from twisted.web.http import ACCEPTED
+from twisted.web.http import ACCEPTED, BAD_REQUEST, FORBIDDEN, NOT_FOUND
 
 
 @attributes(['nova_message'])
@@ -69,7 +69,7 @@ def bad_request(message, request):
 
     :return: dictionary representing the error body.
     """
-    return _nova_error_message("badRequest", message, 400, request)
+    return _nova_error_message("badRequest", message, BAD_REQUEST, request)
 
 
 def not_found(message, request):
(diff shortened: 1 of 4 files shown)


glyph:
Looks like you're missing one line of coverage: https://coveralls.io/builds/2483366/source?filename=mimic%2Fmodel%2Fnova_objects.py#L668

mimic/model/nova_objects.py
```diff
@@ -73,6 +72,32 @@ def bad_request(message, request):
...
+    """
+    Return a 404 error body associated with a Nova not found error.
+    Also sets the response code on the request.
+
+    :param str message: The message to include in the bad request body.
+    :param request: The request on which to set the response code.
+
+    :return: dictionary representing the error body.
+    """
+    return _nova_error_message("itemNotFound", message, 404, request)
```
glyph: Can you put this back to use the symbolic `NOT_FOUND` constant?  I've noticed Mimic's code tends to drift towards magic constants. Granted, "404" is a pretty well known one, but there's no particular reason not to be consistent.  (Plus, I would really like it if one day we could stop using integers for this, integers are a bad data-type for HTTP response codes for a number of reasons, and symbolic constants make it easier to change / update that interface.)

mimic/model/nova_objects.py
```diff
@@ -73,6 +72,32 @@ def bad_request(message, request):
...
+    """
+    Return a 403 error body associated with a Nova forbidden error.
+    Also sets the response code on the request.
+
+    :param str message: The message to include in the bad request body.
+    :param request: The request on which to set the response code.
+
+    :return: dictionary representing the error body.
+    """
+    return _nova_error_message("forbidden", message, 403, request)
```
glyph: Similarly, I think this should use `twisted.web.http.FORBIDDEN`.

glyph:
Since we seem to be reassigning stuff to folks after review, and not always having the reviewer land, I've added a label for that state.  Other than the minor issue I noted above, I think this is good to go.

glyph:
Finally, a passing build!  Those small changes are good enough for me, let's land this before another commit makes us wait 10 more hours for Travis :)

Issue #244, opened 2015-05-05
Support states in CLB


kivattik:
- [x] Give CLB and object model that includes states
- [x] Return a response to requests based on the state
  - [x] ex. Return 422 when in PENDING_DELETE
- [x] Provide an interface to set the state (a new endpoint?)

cyli:
This seems to have been finished since there is an endpoint now to change the CLB status, so closing,

PR #245, opened 2015-05-05
Start CLB refactor by making an object

kivattik:
part of #244


glyph:
Looks like the test failure is an issue with testtools; perhaps we shouldn't depend on it if it has a bad interaction with py2app.  But it's unrelated to this PR.

glyph:
Looks like a pretty straightforward change.

PR #246, opened 2015-05-06
Saf2 migrate first method

sam-falvo:
Depends on #245 landing first.

### DIFF
--- a/mimic/model/clb_objects.py
+++ b/mimic/model/clb_objects.py
@@ -2,7 +2,11 @@
 Model objects for the CLB mimic.
 """
 
+from copy import deepcopy
+from mimic.util.helper import seconds_to_timestamp
+from twisted.python import log
 from characteristic import attributes, Attribute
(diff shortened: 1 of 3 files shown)


mimic/model/clb_objects.py
```diff
@@ -16,6 +20,56 @@ def __init__(self):
         self.lbs = {}
         self.meta = {}
 
+    def add_load_balancer(self, tenant_id, lb_info, lb_id, current_timestamp):
+        """Returns response of a newly created load balancer with
```
glyph: I'm surprised the doc linter didn't complain about this; we typically put docstrings on the line following the opening quote. Given that it got past lint I won't block on it but just for future reference :).

glyph:
Thanks for adding the additional parameter documentation while you were doing it, @sam-falvo !

I do wonder why github's diff viewer showed me the full diff even though I had merged master.

PR #247, opened 2015-05-06
testtools 1.8 came out yesterday, and it's broken

glyph:
Something to do with `USER_BASE` not being importable?  It's not
entirely clear what the issue is, but it looks like it might be an issue
with py2app:

https://bitbucket.org/ronaldoussoren/py2app/issue/111/user_base-cannot-be-imported


glyph:
RFR, or at least it will be if the builds pass.

sam-falvo:
+1 pending Travis-CI's approval.

PR #249, opened 2015-05-06
False-negative create behavior for Nova

cyli:
Where Nova returns a 500, but the server is created anyway.


lekhajee:
+ 1 lgtm

glyph:
I have cancelled the "push" build on Travis, so we should land as soon as "pr" passes.

glyph:
Looks like we have a winner!

PR #250, opened 2015-05-06
moved get_load_balancers to object

kivattik:


lekhajee:
@kivattik : Not sure if this is ready for review. Please add the labels accordingly.

Also how about deleting `verify_and_update_lb_state` from `canned_responses` and importing it into `canned_responses`  from `models` until the others methods are moved over.

glyph:
FYI: merging master will make future commits to this branch trigger one build instead of two.

glyph:
Assuming it's WIP until @kivattik says otherwise.

kivattik:
@lekhajee : Since we're planning to migrate all the functions shortly, it doesn't seem worth it to me to do a second push for the import. I can include that in the next PR to move a function.

glyph:
Every PR should really be a self-consistent change that we are comfortable releasing.  While we do expect that things will move along quickly, we should not *rely* on that; every project I've ever worked on has had a tendency of getting stopped in the middle of important work while other priorities become higher and then lingering in that state for months or years.  Given that the function in question is now duplicated, that means there's a duplicate code path.  My guess is that when coveralls runs it may discover uncovered lines as a result; if it doesn't, that's still a very confusing situation that someone could walk into.  So I do think we need to delete the duplicate definition in `canned_responses` before this can land.

What I was saying before about merging master (now that you'll have to do another push) is that if you do

```
$ git merge master
$ hack hack hack
$ git ci -am "delete the function, add the imports"
$ git push
```

then upon that push it will only run the "pr" builder, which will be about 2x as fast.  So you may want to do that as you make this change.

glyph:
Was the moving of `del_load_balancer` accidental?  It seems unused, and it's another duplicated method.

kivattik:
Yep, it accidentally got included in the copy/paste when I moved functions around after creating a circular import that didn't work well.

glyph:
OK! Looks pretty good. I will merge now.

Issue #204, opened 2015-03-18
"sequence" behavior for server creation (hopefully generalizable for other things)


glyph:
Sometimes you need a server creation to intermittently fail in a deterministic way.

One way you can do this right now is to specify a failure behavior and then delete the behavior.  However, this will not give you an opportunity to delete the behavior between calls if the system under test is making calls close together with no blocking interaction with your test suite in between them.  (For example, a `for` loop that boots 10 servers in a row without even waiting for responses.)

We can implement this by having a `sequence` behavior, whose description takes some JSON that describes an array of *other* behaviors.  Let's take the example of a creation behavior that alternately fails with a 500 and then a 404, over and over again forever.

```json
{
    "name": "sequence",
    "parameters": {
        "behaviors": [
            {
                "name": "fail",
                "parameters": {
                    "code": 500,
                    "message": "lol"
                }
            },
            {
                "name": "fail",
                "parameters": {
                    "code": 404,
                    "message": "wut"
                }
            }
        ],
        "repeat": true
    }
}
```

We should also add a specifically-named `default` behavior so that it can be interleaved.  For example, a server that fails every other time it is created, forever, could look like this:

```json
{
    "name": "sequence",
    "parameters": {
        "behaviors": [
            {
                "name": "fail",
                "parameters": {
                    "code": 500,
                    "message": "lol"
                }
            },
            {
                "name": "default",
                "parameters": {}
            }
        ],
        "repeat": true
    }
}
```

When `"repeat"` is set to `false`, you instead go back to the default behavior after the last behavior described in the sequence's list of behaviors is executed.  So this creation behavior would fail 3 times, then succeed forever after:

```json
{
    "name": "sequence",
    "parameters": {
        "behaviors": [
            {
                "name": "fail",
                "parameters": {
                    "code": 500,
                    "message": "lol"
                }
            },
            {
                "name": "fail",
                "parameters": {
                    "code": 500,
                    "message": "lol"
                }
            },
            {
                "name": "fail",
                "parameters": {
                    "code": 500,
                    "message": "lol"
                }
            }
        ],
        "repeat": false
    }
}
```

glyph:
This may need some refactoring so that created behaviors can be aware of themselves as registered in a particular `BehaviorRegistry` so that the "repeat" behavior can remove itself.

cyli:
Can we close this ticket, since we have the repeat behavior for now?  Was there anything else we wanted to add to this behavior?

glyph:
[...]

PR #251, opened 2015-05-07
Sequence behavior

glyph:
This is a part of the fix for #204, because it doesn't provide the specified `repeat` option; it just always repeats. I wonder if we should do it that way; perhaps we should have different names for the different things you do when you get to the end?

### DIFF
--- a/mimic/model/nova_objects.py
+++ b/mimic/model/nova_objects.py
@@ -4,6 +4,7 @@ Model objects for the Nova mimic.
 
 import re
 
+from itertools import cycle
 from characteristic import attributes, Attribute
 from random import randrange
 from json import loads, dumps
@@ -537,7 +538,7 @@ def sequence(parameters):
     other behaviors.
     """
     behavior_specification = parameters["behaviors"]
-    behavior_objects = [
+    behavior_objects = cycle([
         (
             server_creation.create_behavior(behavior["name"],
                                             behavior["parameters"])
@@ -545,14 +546,11 @@ def sequence(parameters):
             else server_creation.default_behavior
         )
         for behavior in behavior_specification
-    ]
+    ])
 
     def rotating_behavior(collection, http, json, absolutize_url):
-        current = behavior_objects[rotating_behavior.index]
-        rotating_behavior.index += 1
-        rotating_behavior.index %= len(behavior_objects)
+        current = next(behavior_objects)
         return current(collection, http, json, absolutize_url)
-    rotating_behavior.index = 0
     return rotating_behavior
 
 
--- a/mimic/model/nova_objects.py
+++ b/mimic/model/nova_objects.py
@@ -527,6 +527,35 @@ def active_then_error(parameters):
     return fail_later
 
 
+@server_creation.declare_behavior_creator("sequence")
+def sequence(parameters):
+    """
+    Sometimes a sequence of behaviors occur when you try to create a server in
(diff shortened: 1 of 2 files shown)


mimic/model/nova_objects.py
```diff
@@ -527,6 +527,35 @@ def fail_later(server):
     return fail_later
 
 
+@server_creation.declare_behavior_creator("sequence")
+def sequence(parameters):
+    """
+    Sometimes a sequence of behaviors occur when you try to create a server in
+    a predictable pattern.
+
+    Takes one parameter, ``behaviors``, which is a list of specifications of
```
cyli: Would it make sense to document here that like the list is a dictionary containing the name under the key 'name' and the parameters under the name 'parameters'?  This specification is different than the other behavior specifications (since they just take a dictionary as parameters?)

mimic/model/nova_objects.py
```diff
@@ -527,6 +527,35 @@ def fail_later(server):
...
+            server_creation.create_behavior(behavior["name"],
+                                            behavior["parameters"])
+            if behavior["name"] != "default"
+            else server_creation.default_behavior
+        )
+        for behavior in behavior_specification
+    ]
+
+    def rotating_behavior(collection, http, json, absolutize_url):
+        current = behavior_objects[rotating_behavior.index]
```
cyli: Maybe `itertools.repeat` can be used here?

mimic/model/nova_objects.py
glyph: This … isn't quite true? Normally they take a name under "name" and parameters under "parameters" and criteria under "criteria"; here, the criteria are already specified.  Perhaps an example dictionary would be helpful though.

mimic/model/nova_objects.py
glyph: I think you mean `itertools.cycle`?  I explicitly wanted to do repeating in a separate pull request, though, as the description said.

mimic/model/nova_objects.py
cyli: Sorry, I mean they they are documented with name and parameters in a dictionary in the REST api, but not in the object model. The behaviors above describes the dictionary that is in parameters.

Whereas the behavior specification here is expecting a list of dictionaries as would be passed as the body to the REST api call.

glyph: Fundamentally you're right, this documentation is indecipherable, I will include an example and describe it more clearly.

glyph:
Hmm… I was thinking "with minor changes" and "if tests pass" would be mutually exclusive – I really wish github let you put a description on labels so we could explain them in more detail :-\

glyph:
In other words, "with minor changes" would mean "change it and then land", but "if tests pass" would mean "land exactly as-is as long as the tests pass".

glyph:
In both cases you'd need to wait for a passing test run.

cyli:
@glyph - I figured they were composable - e.g. if tests pass after minor changes.

glyph:
Does the docstring look good now though? This is kind of subtle, and I think we don't have the greatest docs, so I'd really like an affirmative "yes" before pulling the trigger on this :).

cyli:
@glyph +1

glyph:
Given the build times on travis, I'm going to optimistically assume that the previous correct build results and the smattering of test passes and coverage results on the current code are good enough for a merge.

PR #254, opened 2015-05-08
move list_load_balancers

kivattik:


kivattik:
This might be a duplicate of Sam's.

glyph:
I only see a PR for `del_load_balancers` from sam so far, so I don't think it's a dup.

glyph:
This is in support of #244.

glyph:
Nice and straightforward, tests pass. Thanks!

PR #255, opened 2015-05-08
move CLB add_node to the clb object

kivattik:


glyph:
The ☂ label is intended to be only for the top-level thing, not for everything that falls under one of those things :).

glyph:
I think what you meant by attaching ☂ was "in support of #244?

kivattik:
Yep, that is what I meant.

glyph:
Also looks pretty straightforward, address merge conflicts and land when ready.

PR #256, opened 2015-05-08
Refine the list loadbalancers function docstring

sam-falvo:


sam-falvo:
I will be AFK until Monday, possibly tuesday.  So, please assign ticket to @kivattik  if I'm not able to address issues in a timely manner.  Thanks!

glyph:
First order of business is addressing the merge conflicts…

glyph:
But, reviewing the changes, assuming the merge is clean and the tests pass, this looks fine.

mimic/model/clb_objects.py
```diff
@@ -6,6 +6,7 @@
 from twisted.python import log
 from characteristic import attributes, Attribute
 from mimic.canned_responses.loadbalancer import (load_balancer_example,
+                                                 _prep_for_list,
```
lekhajee: `_prep_for_list` belongs inside the `model/clb_objects.py` as against the `canned_responses`, as its converting the response object to the format required for LIST call.

sam-falvo: Yes it does.  However, if I put it there, coveralls fails because coverage drops.

lekhajee: Does not look like `_prep_for_list` is used by anything else in `canned_responses/loadbalancers.py`

lekhajee:
@sam-falvo : Looks like merging the master overrode some things and `prep_for_list` is being imported from `canned_responses` as well. Same with `list_load_balancers`.

mimic/model/clb_objects.py
```diff
@@ -10,11 +10,23 @@
 from mimic.canned_responses.loadbalancer import (load_balancer_example,
                                                  _verify_and_update_lb_state,
                                                  _lb_without_tenant,
-                                                 _prep_for_list,
                                                  _format_nodes_on_lb,
                                                  _delete_node)
 
 
+def _prep_for_list(lb_list):
```
cyli: If this is being migrated instead of imported, we can probably delete it in `canned_responses.loadbalancer`, since nothing else uses it there.  Leaving it will probably cause coverage to complain, since that code will no longer be exercised.

cyli:
Awesome, thanks for the docstring fix!  LGTM if travis +1's.  :)

PR #257, opened 2015-05-08
migrate clb get_nodes

kivattik:
In support of #244


glyph:
Another nice straightforward change, assuming tests pass and merge conflicts are addressed.

PR #258, opened 2015-05-08
migrate clb delete_node

kivattik:
Supports #244


glyph:
This is missing the corresponding deletion of `delete_node`, so it will result in duplicate code again.

glyph:
Cool, looking good.

PR #260, opened 2015-05-09
migrate CLB list_nodes

kivattik:
supports #244


glyph:
Also looks pretty straightforward, also requires merge conflict resolution.

PR #263, opened 2015-05-12
Adjust failure on create behaviors

cyli:
1. The JSON blob passed back wasn't really a canonical Nova error, which has the type of failure in it.
2. We sometimes want to inject non-JSON errors.  For example: `403 Forbidden\\n\\nAccess was denied to this resource.\\n\\n Networks (00000000-0000-0000-0000-000000000000,11111111-1111-1111-1111-111111111111) required but missing` is an error we've seen.

### DIFF
--- a/mimic/model/nova_objects.py
+++ b/mimic/model/nova_objects.py
@@ -14,7 +14,6 @@ from six import string_types
 
 from mimic.util.helper import (
     seconds_to_timestamp,
-    invalid_resource,
     random_string,
 )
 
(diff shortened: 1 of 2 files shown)


glyph:
I think I might want to be a *little* more awake before I approve this, but it looks pretty good right now.

mimic/model/nova_objects.py
```diff
@@ -418,27 +417,88 @@ def hooked(collection, http, json, absolutize_url):
     return hooked
 
 
-@server_creation.declare_behavior_creator("fail")
-def create_fail_behavior(parameters):
+def _get_failure_behavior(parameters, create=False):
     """
-    Create a failing behavior for server creation.
+    Helper function to produce a failure to create function.  Either creating
```
glyph: One minor nitpick: I try to avoid ever saying the type of the thing that I'm documenting in the documentation.  Otherwise eventually every class would start with "This is a class." and every module would start with "This is a module."

Also, "Either creating the server or not." is a sentence fragment.

This is a little tricky to describe but try to stick to its purpose: "Build a response body indicating a failure in Nova…"; something like that.

Also, given the plethora of types might be talking about when one describes a "response" or even a "response body", an `:rtype:` and `:return:` annotation would be useful here.

mimic/model/nova_objects.py
```diff
@@ -447,23 +507,29 @@ def create_success_report_failure_behavior(parameters):
...
+    Otherwise, the following JSON body will be synthesized (as per the
+    canonical Nova error format):
 
-    def create_then_fail(collection, http, json, absolutize_url):
-        Server.from_creation_request_json(
-            collection, json, lambda: randrange(255))
-        http.setResponseCode(status_code)
-        return dumps(invalid_resource(failure_message, status_code))
-    return create_then_fail
+    ```
```
glyph: this syntax (```` ``` ````) is markdown; I am pretty sure the right way to format a code block in ReST is `::` then indent.  In any case you're missing the ending ```` ``` ```` :)

glyph: We should probably have a docstring linter.  I guess we already do in some sense, but there are very few things that are outright *errors* in ReST, just lots of things that format horribly. I don't think we have any online generation of docs I could point to to see how this formatted.

mimic/test/test_nova.py
```diff
@@ -1261,15 +1284,17 @@ def test_create_server_failure_based_on_metadata(self):
...
+        self.assertEquals(
+            failing_create_response_body['specialType']['code'], 503)
 
-    def test_create_false_negative_failure_using_behaviors(self):
+    def _try_false_negative_failure(self, failure_type=None):
         """
-        :func:`create_server` fails with given error message and response
-        code, but creates the server anyway, when a behavior is registered
-        that matches its hostname.
+        Helper function to list servers and verify that there are no servers,
```
glyph: Once again, starting things with "helper function" is an anti-pattern (albeit one I do myself sometimes too) and this should just say what it does without qualifying its role as a function or helper or utility or whatever.

glyph:
The minor docstring issues don't strike me as important enough to block on, especially given that so much of this stuff was undocumented to begin with.  I'd appreciate it if you could follow some follow-on low-priority issues to clean these things up, but I'll land this now.

PR #264, opened 2015-05-12
Mimic identity changes for ele

lekhajee:
This a temporary change that implements only a few rbac like roles (required by ele tests). This should  exist only until RBAC is globally applied for users within mimic.

Eventually tests within ele are going to be changed to hit the mimic identity's control plane to specify the role/behavior expected for a tenant.

### DIFF
--- a/mimic/canned_responses/mimic_presets.py
+++ b/mimic/canned_responses/mimic_presets.py
@@ -19,5 +19,19 @@ get_presets = {"loadbalancers": {"lb_building": "On create load balancer, keeps 
                            "invalid_flavor_ref": ["INVALID-FLAVOR-ID", "8888", "-4", "1"],
                            "server_error": "sets server state to error on create",
                            "server_building": "sets the server to be in building state for given time"
-                                              " in seconds"}
-               }
+                                              " in seconds"},
+               "identity": {
+                   "maas_admin_roles": {
+                       "this_is_an_impersonator_token": ["impersonator.joe", "12345"],
+                       "this_is_an_impersonator_token_also": ["impersonator.eve", "12453"],
+                       "impersonate_watson": ["impersonator.watson", "09876"],
+                       "impersonate_creator": ["impersonator.freeman", "09090"],
+                       "this_is_an_impersonator_token_also_2": ["impersonator.foonatical", "91169"],
+                       "impersonate_foo_token": ["impersonate_foo", "91111"]},
+               "racker_token": {"this_is_a_racker_token": ["racker.support", "12345"]},
+               "observer_role": ["09876"],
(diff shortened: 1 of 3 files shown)


mimic/canned_responses/mimic_presets.py
```diff
@@ -19,5 +19,19 @@
...
+                                              " in seconds"},
+               "identity": {
+                   "maas_admin_roles": [
+                       "this_is_an_impersonator_token",
+                       "this_is_an_impersonator_token_also",
+                       "impersonate_watson",
+                       "impersonate_creator",
+                       "this_is_an_impersonator_token_also_2",
+                       "impersonate_foo_token"],
+               "racker_token": ["this_is_a_racker_token"],
```
cyli: Non-blocking - I think the indentation starting from here might be kind of messed up (should it be lined up with `"maas_admin_roles"`?)

mimic/rest/auth_api.py
```diff
@@ -190,6 +191,46 @@ def validate_token(self, request, token_id):
             response["access"]["RAX-AUTH:impersonator"] = impersonator_user_role(
```
cyli: Non-blocking - just wanted to make sure that it's ok for this logic that sets the impersonation token to get run even if it's one of the tenants in the hard-coded maas roles?

cyli:
Just some nitpicks - particularly the indentation.  But other than that, LGTM if this is needed right away.  :)

glyph:
I am guessing this is RFR since @cyli is already reviewing it :).

mimic/test/test_auth.py
```diff
@@ -915,6 +916,114 @@ def test_response_for_validate_impersonated_token_multiple_users(self):
         self.assertEqual(json_body6["access"]["RAX-AUTH:impersonator"]["name"],
                          json_body1["access"]["user"]["name"])
 
+    def test_response_for_validate_token_with_maas_admin_role(self):
+        """
+        Test to verify :func: `validate_token` when the token_id provided
+        is of an maas admin user specified in `mimic_presets`.
+        """
+        core = MimicCore(Clock(), [ExampleAPI()])
+        root = MimicRoot(core).app.resource()
```
glyph: Given that each of these tests starts with *exactly* these two lines, maybe you could factor out just a simple function so we could do `core, root = core_and_example_root()`?

mimic/rest/auth_api.py
lekhajee: Yes, we the token being validated was got from impersonation, we expect the impersonator role to exist.

cyli:
LGTM!

lekhajee:
@cyli: Thanks much for the quick review and merge :)

PR #266, opened 2015-05-14
Move del_load_balancer to the clb object

kivattik:
There was confusion over duplication when we previously attempted to move this function over, and it didn't look like the changes took.

Supports #244

### DIFF
--- a/mimic/canned_responses/loadbalancer.py
+++ b/mimic/canned_responses/loadbalancer.py
@@ -5,9 +5,7 @@ add/get/delete/list nodes
 """
 from random import randrange
 from copy import deepcopy
-from mimic.util.helper import (EMPTY_RESPONSE,
-                               not_found_response, invalid_resource,
-                               set_resource_status, seconds_to_timestamp)
+from mimic.util.helper import (set_resource_status, seconds_to_timestamp)
 from twisted.python import log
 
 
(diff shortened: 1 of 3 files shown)


mimic/canned_responses/loadbalancer.py
```diff
@@ -5,9 +5,7 @@
 """
 from random import randrange
 from copy import deepcopy
-from mimic.util.helper import (EMPTY_RESPONSE,
-                               not_found_response, invalid_resource,
-                               set_resource_status, seconds_to_timestamp)
+from mimic.util.helper import (set_resource_status, seconds_to_timestamp)
```
glyph: FYI, the parentheses here are unnecessary unless the list of imports is long enough to wrap lines.

mimic/model/clb_objects.py
```diff
@@ -260,6 +260,42 @@ def add_node(self, node_list, lb_id, current_timestamp):
 
         return not_found_response("loadbalancer"), 404
 
+    def del_load_balancer(self, lb_id, current_timestamp):
+        """
+        Returns response for a load balancer
```
glyph: The wrapping here looks a little off... I should probably give you my python-docstring-wrapping script.

glyph:
Looks good to merge; hopefully take those comments under advisement for future PRs but none of them should block. So I will merge now.

PR #267, opened 2015-05-14
pin the version of cryptography used by the mac app

glyph:
The release today hard depends on idna.  While this should be installed, it sadly breaks py2app due to something about the syntax.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,5 @@
 klein==15.0.0
+cryptography==0.8.2
 twisted==15.0.0
 jsonschema==2.0
 treq==15.0.0


glyph:
This fixes the currently failing master build, which first showed up on https://travis-ci.org/rackerlabs/mimic/jobs/62500691

glyph:
There is also an upstream bug for the underlying issue, filed here: https://bitbucket.org/ronaldoussoren/py2app/issue/156/virtualenvpy-recipe-calls-methods-renamed#comment-None

cyli:
LGTM

glyph:
Thanks for the quick review.

I think there's a better solution to this problem that would allow us to package idna and upgrade to current cryptography, but it's tricky and involves convincing modulefinder not to look at the idna package.

PR #268, opened 2015-05-14
Modulefinder dont find idna

glyph:
This is a different workaround, slightly trickier, that will allow us to upgrade IDNA.


cyli:
The fix has been released, so we can probably just upgrade:  https://pypi.python.org/pypi/modulegraph

glyph:
I think I'll just close this PR and open a new one, then - glad we don't have to merge these shenanigans ;)

PR #272, opened 2015-05-18
Attempt to level up the travis config

reaperhulk:
This uses the docker infra where it can and overrides the proper variables for OS X so you can have both. In theory.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,45 +1,38 @@
-language: c
-
-os:
-  - osx
-  - linux
-
-compiler:
-  - clang
-  - gcc
-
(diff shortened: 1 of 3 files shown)


glyph:
Thanks, @reaperhulk !

This appears to provide us with the same test matrix, and radically faster builds - just on the elapsed build time (once everything's started) it's still a 2x speedup, and by using the container-based infrastructure builds *start* much faster.  There's still probably more to do here (particularly, using wheels somehow for the pyobjc frameworks on the OS X builders) but this is a huge improvement for now.

PR #273, opened 2015-05-19
Wheel some things

reaperhulk:
These wheels will be cached on the linux side between runs so this might go faster. Anything is possible if you believe hard enough.

Of note, only wheeling cryptography and lxml (used by sphinx) right now.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,11 @@ cache:
     directories:
       - $HOME/.cache/pip
 
+env:
+    global:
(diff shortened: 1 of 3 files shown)


glyph:
My only concern is the `passenv` thing in the tox config... but... looks like it works fine on the latest, and our CI only covers the latest!  I think I will merge.

glyph:
Thanks again @reaperhulk - this definitely looks like another good speed-up.

Issue #158, opened 2015-01-22
Change the `data_for_api` to be able to return correct data for the desired plugin


lekhajee:
Change the `data_for_api` to be able to return the correct data type for the desired plugin instead of the caller passing a factory.

For example: When RCV3 needs to talk to Nova, `data_for_api` should be able to return `GlobalServerCollection` from the nova plugin.

lekhajee:
To accomodate #152

lekhajee:
This is being worked as part of #148

glyph:
It didn't really end up happening in #148 :)

PR #274, opened 2015-05-19
Create clb ctrl api

sam-falvo:
This PR allows a client to alter a CLB's status setting from within an integration test.

### DIFF
--- a/mimic/model/clb_objects.py
+++ b/mimic/model/clb_objects.py
@@ -21,6 +21,11 @@ class BadKeysError(Exception):
     doesn't exist.
     """
     def __init__(self, msg, keys):
+        """
+        Exception constructor.
+
+        :param list keys: The set of keys that are considered bad.
+        """
         super(BadKeysError, self).__init__(msg)
         self.keys = keys
 
--- a/mimic/model/clb_objects.py
+++ b/mimic/model/clb_objects.py
@@ -37,6 +37,7 @@ class RegionalCLBCollection(object):
         """
         self.lbs = {}
         self.meta = {}
+        self.returnOverrides = {}
 
     def add_load_balancer(self, tenant_id, lb_info, lb_id, current_timestamp):
         """
@@ -78,11 +79,22 @@ class RegionalCLBCollection(object):
 
         return {'loadBalancer': new_lb}, 202
 
+    def set_return_override(self, lb_id, statusCode):
+        """
+        Sets the return override for a load balancer (statusCode != 0)
+        or removes it (statusCode == 0).
+        """
+        self.returnOverrides[lb_id] = statusCode
+
     def get_load_balancers(self, lb_id, current_timestamp):
         """
         Returns the load balancers with the given lb id, with response
         code 200. If no load balancers are found returns 404.
         """
+        print("RegionalCLBCollection({})".format(self))
+        if lb_id in self.returnOverrides:
+            return "Response overridden", self.returnOverrides[lb_id]
+
         if lb_id in self.lbs:
             _verify_and_update_lb_state(self, lb_id, False, current_timestamp)
             log.msg(self.lbs[lb_id]["status"])
--- a/mimic/model/clb_objects.py
+++ b/mimic/model/clb_objects.py
@@ -14,19 +14,20 @@ from mimic.canned_responses.loadbalancer import (load_balancer_example,
                                                  _delete_node)
 
 
+@attributes(["keys"])
 class BadKeysError(Exception):
     """
     When trying to alter the settings of a load balancer, this exception will
     be raised if you attempt to alter an attribute which doesn't exist.
     """
-    def __init__(self, msg, keys):
-        """
-        Exception constructor.
 
-        :param list keys: The set of keys that are considered bad.
-        """
-        super(BadKeysError, self).__init__(msg)
-        self.keys = keys
+
+@attributes(["value", "accepted_values"])
+class BadValueError(Exception):
+    """
+    When trying to alter the settings of a load balancer, this exception will
+    be raised if you attempt to set a valid attribute to an invalid setting.
+    """
 
 
 def _prep_for_list(lb_list):
@@ -113,7 +114,7 @@ class RegionalCLBCollection(object):
             if k not in supported_keys:
                 badKeys.append(k)
         if len(badKeys) > 0:
-            raise BadKeysError("Attempt to alter a bad attribute", badKeys)
+            raise BadKeysError("Attempt to alter a bad attribute", keys=badKeys)
 
         if "status" in kvpairs:
             supported_statuses = [
@@ -121,10 +122,11 @@ class RegionalCLBCollection(object):
             ]
             s = kvpairs["status"]
             if s not in supported_statuses:
-                raise ValueError(
-                    "Unsupported status {} not one of {}".format(
+                raise BadValueError(
+                    "Unsupported status {0} not one of {1}".format(
                         s, supported_statuses
-                    )
+                    ),
+                    value=s, accepted_values=supported_statuses
                 )
 
         self.lbs[lb_id].update(kvpairs)
--- a/mimic/model/clb_objects.py
+++ b/mimic/model/clb_objects.py
@@ -91,7 +91,6 @@ class RegionalCLBCollection(object):
         Returns the load balancers with the given lb id, with response
         code 200. If no load balancers are found returns 404.
         """
-        print("RegionalCLBCollection({})".format(self))
         if lb_id in self.returnOverrides:
             return "Response overridden", self.returnOverrides[lb_id]
 
--- a/mimic/model/clb_objects.py
+++ b/mimic/model/clb_objects.py
@@ -52,10 +52,17 @@ class RegionalCLBCollection(object):
         """
         Returns the cloud load balancer identified by clb_id.
         """
+        return self.lbs[clb_id]
+
+    def __contains__(self, clb_id):
+        """
+        Returns true if the CLB ID is registered with our list of load
+        balancers.
+        """
         for i in self.lbs:
             if i == clb_id:
-                return self.lbs[i]
-        raise KeyError(clb_id)
+                return True
+        return False
 
     def add_load_balancer(self, tenant_id, lb_info, lb_id, current_timestamp):
         """
--- a/mimic/model/clb_objects.py
+++ b/mimic/model/clb_objects.py
@@ -17,8 +17,7 @@ from mimic.canned_responses.loadbalancer import (load_balancer_example,
 class BadKeysError(Exception):
     """
     When trying to alter the settings of a load balancer, this exception will
-    be raised if you misspell or otherwise attempt to alter an attribute which
-    doesn't exist.
+    be raised if you attempt to alter an attribute which doesn't exist.
     """
     def __init__(self, msg, keys):
         """
--- a/mimic/model/clb_objects.py
+++ b/mimic/model/clb_objects.py
@@ -14,6 +14,17 @@ from mimic.canned_responses.loadbalancer import (load_balancer_example,
                                                  _delete_node)
 
 
+class BadKeysError(Exception):
+    """
+    When trying to alter the settings of a load balancer, this exception will
+    be raised if you misspell or otherwise attempt to alter an attribute which
+    doesn't exist.
+    """
+    def __init__(self, msg, keys):
+        super(BadKeysError, self).__init__(msg)
+        self.keys = keys
+
+
 def _prep_for_list(lb_list):
     """
     Removes tenant id and changes the nodes list to 'nodeCount' set to the
@@ -92,8 +103,27 @@ class RegionalCLBCollection(object):
         Sets zero or more attributes on the load balancer object.
         Currently supported attributes include: status.
         """
+        supported_keys = ["status"]
+        badKeys = []
+        for k in kvpairs:
+            if k not in supported_keys:
+                badKeys.append(k)
+        if len(badKeys) > 0:
+            raise BadKeysError("Attempt to alter a bad attribute", badKeys)
+
         if "status" in kvpairs:
-            self.lbs[lb_id]["status"] = kvpairs["status"]
+            supported_statuses = [
+                "ACTIVE", "ERROR", "PENDING_DELETE", "PENDING_UPDATE"
+            ]
+            s = kvpairs["status"]
+            if s not in supported_statuses:
+                raise ValueError(
+                    "Unsupported status {} not one of {}".format(
+                        s, supported_statuses
+                    )
+                )
+
+        self.lbs[lb_id].update(kvpairs)
 
     def get_load_balancers(self, lb_id, current_timestamp):
         """
--- a/mimic/model/clb_objects.py
+++ b/mimic/model/clb_objects.py
@@ -19,10 +19,8 @@ def _prep_for_list(lb_list):
     Removes tenant id and changes the nodes list to 'nodeCount' set to the
     number of node on the LB
     """
-    entries_to_keep = (
-        'name', 'protocol', 'id', 'port', 'algorithm', 'status', 'timeout',
-        'created', 'virtualIps', 'updated', 'nodeCount'
-    )
+    entries_to_keep = ('name', 'protocol', 'id', 'port', 'algorithm', 'status', 'timeout',
+                       'created', 'virtualIps', 'updated', 'nodeCount')
     filtered_lb_list = []
     for each in lb_list:
         filtered_lb_list.append(
@@ -42,27 +40,12 @@ class RegionalCLBCollection(object):
         self.lbs = {}
         self.meta = {}
 
-    def __len__(self):
-        """
-        The number of cloud load balancers in this collection.
-        """
-        return len(self.lbs)
-
-    def __getitem__(self, clb_id):
-        """
-        Returns the cloud load balancer identified by clb_id.
-        """
-        return self.lbs[clb_id]
-
-    def __contains__(self, clb_id):
+    def lb_in_region(self, clb_id):
         """
         Returns true if the CLB ID is registered with our list of load
         balancers.
         """
-        for i in self.lbs:
-            if i == clb_id:
-                return True
-        return False
+        return clb_id in self.lbs
 
     def add_load_balancer(self, tenant_id, lb_info, lb_id, current_timestamp):
         """
--- a/mimic/model/clb_objects.py
+++ b/mimic/model/clb_objects.py
@@ -19,11 +19,15 @@ def _prep_for_list(lb_list):
     Removes tenant id and changes the nodes list to 'nodeCount' set to the
     number of node on the LB
     """
-    entries_to_keep = ('name', 'protocol', 'id', 'port', 'algorithm', 'status', 'timeout',
-                       'created', 'virtualIps', 'updated', 'nodeCount')
+    entries_to_keep = (
+        'name', 'protocol', 'id', 'port', 'algorithm', 'status', 'timeout',
+        'created', 'virtualIps', 'updated', 'nodeCount'
+    )
     filtered_lb_list = []
     for each in lb_list:
-        filtered_lb_list.append(dict((entry, each[entry]) for entry in entries_to_keep))
+        filtered_lb_list.append(
+            dict((entry, each[entry]) for entry in entries_to_keep)
+        )
     return filtered_lb_list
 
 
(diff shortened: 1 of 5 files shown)


mimic/rest/loadbalancer_api.py
```diff
@@ -54,10 +57,104 @@ def resource_for_region(self, region, uri_prefix, session_store):
...
+        return clb_region_collection
+
+    def _collection_from_tenant(self, tenant_id):
+        """
+        Retrieve the server collection for this region for the given tenant.
+        """
+        return (self.api_mock.lb_api._get_session(self.session_store, tenant_id)
+                .collection_for_region(self.region))
+
+    @app.route('/v2/<string:tenant_id>/loadbalancer/<string:clb_id>/returnOverride/<int:statusCode>', methods=['POST'])
```
cyli: The load balancer API specifies `lb_id` as an int, rather than a string, so the storage dictionary is keyed by ints, rather than strings.  In order to make that match, maybe this should be changed to `<int:clb_id>`?

mimic/rest/loadbalancer_api.py
```diff
@@ -54,10 +57,104 @@ def resource_for_region(self, region, uri_prefix, session_store):
...
+
+@attributes(["api_mock", "uri_prefix", "session_store", "region"])
+class LoadBalancerControlRegion(object):
+    """
+    Klein routes for load balancer's control API within a particular region.
+    """
+
+    app = MimicApp()
+
+    def session(self, tenant_id):
```
cyli: With `_collections_from_tenant` implemented, this can probably be eliminated, since nothing else in this object is using this API mock's storage.

mimic/rest/loadbalancer_api.py
sam-falvo: Thank you; I made this change locally and now all tests pass.

mimic/rest/loadbalancer_api.py
```diff
@@ -54,10 +55,104 @@ def resource_for_region(self, region, uri_prefix, session_store):
         Get an :obj:`twisted.web.iweb.IResource` for the given URI prefix;
         implement :obj:`IAPIMock`.
         """
+        self.session_store = session_store
```
cyli: Apologies, I messed this up before and got confused - I think with `_collections_from_tenant`, probably storing this on the `LoadBalancerApi` may not be necessary?  (so it doens't need to be stored here or in the `init`, and in `LoadBalancerControlApi.resource_for_region`, the `session_store` passed in the function parameter can be used instead of `self.lb_api.session_store` (since it should be the same session store either way).

Really really sorry I misunderstood this before.

mimic/rest/loadbalancer_api.py
sam-falvo: The test test_different_tenants_same_region_different_lbs fails if I remove this method.

sam-falvo: N/m -- that test is now suddenly _always_ failing.  What did I change *THIS* time?!

mimic/rest/loadbalancer_api.py
```diff
@@ -54,10 +55,104 @@ def resource_for_region(self, region, uri_prefix, session_store):
...
+        and :obj:`SessionStore`.
+
+        For use with ``data_for_api``.
+
+        Temporary hack; see this issue
+        https://github.com/rackerlabs/mimic/issues/158
+        """
+        return (
+            session_store.session_for_tenant_id(tenant_id)
+            .data_for_api(self, lambda: GlobalServerCollections(
```
cyli: This should probably be `GlobalCLBCollections` :)  Else if this function is called for a new tenant, it'd probably break.

mimic/rest/loadbalancer_api.py
sam-falvo: Fixed in 21c18b6281e5c0cb3d1bf178f59e1bde761b443e

mimic/rest/loadbalancer_api.py
cyli: Hm... I think this may be causing a test failure.

cyli:
I think `loadbalancer_plugin.loadbalancer_control` needs to be added to `mimic.test.test_core.CoreBuildingTests.test_from_plugin_includes_all_plugins`, and that should make the rest of the tests pass.

mimic/rest/loadbalancer_api.py
sam-falvo: The aforementioned test appears to be flakey.  I opened a separate mimic issue concerning it.

mimic/model/clb_objects.py
```diff
@@ -19,11 +19,15 @@ def _prep_for_list(lb_list):
     Removes tenant id and changes the nodes list to 'nodeCount' set to the
     number of node on the LB
     """
-    entries_to_keep = ('name', 'protocol', 'id', 'port', 'algorithm', 'status', 'timeout',
-                       'created', 'virtualIps', 'updated', 'nodeCount')
+    entries_to_keep = (
```
glyph: For future PRs, I'd appreciate it if you could open a separate PR for re-factorings / re-formattings like this.  Since I'm looking at this as changes to CLBs, I assumed that this list had changed for some reason, and that the list comprehension below had also changed somehow, so I had to carefully read it repeatedly to make sure that I wasn't missing something.  If I were reviewing this in the context of a simple reformatting PR, I would just have scanned it to make sure that no elements in the list were different, and quickly hit the merge button.  So context is important and apparently trivial cleanups like this can slow down reviews a bit.

Please keep in mind that this suggestion comes from someone who does this *all the time*, so I'm noticing it because it's a bad habit I have :).

mimic/model/clb_objects.py
```diff
@@ -38,6 +42,28 @@ def __init__(self):
         self.lbs = {}
         self.meta = {}
 
+    def __len__(self):
```
glyph: I generally avoid making objects special unless they're going to be "all the way" special, and documented as such.  For example, this change makes `RegionalCLBCollection` somewhat list-like (as you discovered, it is now accidentally iterable) but also somewhat mapping-like.  However, although it is mapping-like it is missing big pieces of the mapping interface: `.__setitem__(...)`, `.clear()`, `.copy()`, `.get(...)`, `.has_key()`, `.keys()`, and so on and so on.  It also gets rid of the symmetry of `RegionalCLBCollection` with other similar types in e.g. the Nova mock which are not partial-mapping-like.

The reason that this not-quite-a-dict-like thing is important is that it becomes a point that maintainers can easily be tripped up on.  When looking at a variable in python code, dynamic typing means you don't have to know which type it necessarily *is*, but in order to be able to make substantive changes you need to know what kind of type it is supposed to be *like*.  If you have tedious method names like `load_balancer_with_names` or `load_balancer_exists`, then it's very easy to go from a caller to a `grep` command to the implementation.

Then there's the issue of the fact that there are two attributes here: `lbs` and `meta`, but we are only forwarding mappish operations to `lbs`.

Finally, a convenient syntax like this should be reserved for a highly polished type with really clear `repr`s and so on so that any clarity lost to using this convenience API is made up for in smooth runtime error reporting and formatting in debug messages.  That makes this a poor *time* to start adding special methods, because the returned values here are going to be simple dictionaries, and if you end up staring at one in a debugger or a stack dump, you won't necessarily know where it came from unless you happen to know what format our idiom of a "load balancer" dictionary happens to look like at the moment.

*Finally* finally there's also the issue that not all of these methods are even used; some of them are responsible for the coverage decrease currently failing the builders.

glyph: To be clear I would not veto a change based on the use of special methods, but I thought it was important to lay out my thoughts on this so we don't start going too far in this direction.

mimic/model/clb_objects.py
```diff
@@ -38,6 +42,28 @@ def __init__(self):
...
+        Returns the cloud load balancer identified by clb_id.
+        """
+        return self.lbs[clb_id]
+
+    def __contains__(self, clb_id):
+        """
+        Returns true if the CLB ID is registered with our list of load
+        balancers.
+        """
+        for i in self.lbs:
```
glyph: This is a particularly inefficient and verbose way of implementing this check.  (If efficiency were the only concern I almost certainly wouldn't care, but it also makes the code less clear).

`self.lbs` is a dictionary, and therefore already has a `__contains__` method which checks for key presence using hashing.  So you could implement this to simply do `return (clb_id in self.lbs)`, or even just `return self.lbs.__contains__(clb_id)`.

mimic/model/clb_objects.py
```diff
@@ -78,6 +104,14 @@ def add_load_balancer(self, tenant_id, lb_info, lb_id, current_timestamp):
 
         return {'loadBalancer': new_lb}, 202
 
+    def set_attributes(self, lb_id, kvpairs):
+        """
+        Sets zero or more attributes on the load balancer object.
+        Currently supported attributes include: status.
+        """
+        if "status" in kvpairs:
```
glyph: I am concerned that if we silently ignore any values other than `"status"` it is going to be fiendishly difficult to debug misspellings on the client side.  I would suggest adding a check to make sure that `kvpairs` is a dictionary containing *only* supported keys (i.e. exclusively `"status"` at this moment, but presumably expanded in the near future?) and then after having performed the check, simply doing `self.lbs[lb_id].update(kvpairs)`.

mimic/rest/loadbalancer_api.py
```diff
@@ -18,6 +18,7 @@
 from random import randrange
 
 from mimic.util.helper import invalid_resource, json_dump
+from characteristic import attributes
```
glyph: Definitely not a blocker for this PR, but we are slowly migrating from [Characteristic](https://warehouse.python.org/project/characteristic/) to [`attrs`](https://warehouse.python.org/project/attrs/), so you should be aware of that for new objects.  (You can think of `attrs` as the next version of Characteristic, with a more compact API.

mimic/rest/loadbalancer_api.py
```diff
@@ -26,20 +27,20 @@
 @implementer(IAPIMock, IPlugin)
 class LoadBalancerApi(object):
     """
-    Rest endpoints for mocked Load balancer api.
+    This class registers the load balancer API in the service catalog.
+    It does NOT implement individual endpoints.
```
glyph: I'm curious why you felt it necessary to call out that this doesn't implement individual endpoints here; it seems much like every other plugin in that regard.  Should we add something to more central documentation about that, so we can refer to it from many plugins?

mimic/rest/loadbalancer_api.py
```diff
@@ -26,20 +27,20 @@
...
     """
     def __init__(self, regions=["ORD"]):
         """
         Create an API with the specified regions.
         """
         self._regions = regions
+        self.session_store = None
```
glyph: `IAPIMock` providers should not store a reference to a session store.  Luckily it looks like if this was needed, the code that needed it has been eliminated, so it can be removed without causing problems.

mimic/rest/loadbalancer_api.py
```diff
@@ -54,10 +55,128 @@ def resource_for_region(self, region, uri_prefix, session_store):
...
         lb_region = LoadBalancerRegion(self, uri_prefix, session_store,
                                        region)
         return lb_region.app.resource()
 
+    def _get_session(self, session_store, tenant_id):
+        """
+        Retrieve or create a new LoadBalancer session from a given tenant identifier
+        and :obj:`SessionStore`.
+
+        For use with ``data_for_api``.
```
glyph: Given that this *calls* `data_for_api`, it is a bit misleading to me to document it this way; it makes it sound like the caller should be doing something with `_get_session` and then calling `data_for_api_ themselves, which is backwards; they should let this function call it.

mimic/rest/loadbalancer_api.py
sam-falvo: Because the original docstring (originating from nova's code) says that it implements endpoints.  I spent 4 days wondering why nothing worked.

mimic/test/test_core.py
```diff
@@ -29,10 +29,13 @@ def test_from_plugin_includes_all_plugins(self):
         :class:`MimicCore`, the nova and loadbalancer plugins are included.
         """
         core = MimicCore.fromPlugins(Clock())
-        plugin_apis = set((nova_plugin.nova, loadbalancer_plugin.loadbalancer,
-                           swift_plugin.swift, queue_plugin.queue,
-                           maas_plugin.maas, rackconnect_v3_plugin.rackconnect,
-                           nova_plugin.nova_control_api, glance_plugin.glance))
+        plugin_apis = set((
```
glyph: We should probably change this list to be one-per-line so that adds and removes are more clearly visible.

mimic/rest/loadbalancer_api.py
sam-falvo: This is not my code.  I copy/pasted it from nova's example.

mimic/test/test_loadbalancer.py
```diff
@@ -109,6 +110,31 @@ def _create_loadbalancer(self, name=None, api_helper=None):
         create_lb_response_body = self.successResultOf(treq.json_content(create_lb_response))
         return create_lb_response_body['loadBalancer']['id']
 
+    def test_lb_status_changes_as_requested(self):
+        """
+        Perhaps a bit mislabeled, this test exercises the endpoint that causes
```
glyph: We try hard in mimic to have descriptive test docstrings, that assert something positive about the system under test.  Don't apologize for the mislabeling, just re-label it :).

Particularly, we try to avoid words like "process", "handle", "exercise", "correct", "test", "valid", "execute" etc, since all of these are ways to avoid describing what is happening directly and instead say *that* something is happening.

But to give you an example what I mean, something like this, for this test:

```python
        """
        Clients can ``POST`` to the ``.../loadbalancer/<lb-id>/setAttr``
        :obj:`LoadBalancerControlApi` endpoint in order to change the
        ``status`` attribute on the load balancer identified by the given
        load-balancer ID for the same tenant in the :obj:`LoadBalancerApi`.

        This attribute controls the status code returned when the load balancer
        is retrieved by a ``GET`` request.
        """
```

mimic/rest/loadbalancer_api.py
```diff
@@ -54,10 +55,128 @@ def resource_for_region(self, region, uri_prefix, session_store):
...
+        return clb_region_collection
+
+    def _collection_from_tenant(self, tenant_id):
+        """
+        Retrieve the server collection for this region for the given tenant.
+        """
+        return (self.api_mock.lb_api._get_session(self.session_store, tenant_id)
+                .collection_for_region(self.region))
+
+    @app.route('/v2/<string:tenant_id>/loadbalancer/<int:clb_id>/setAttr', methods=['POST'])
```
glyph: Using `setAttr` as an endpoint here is unfortunately inconsistent with the noun/verb structure of Mimic APIs (and generally things which claim to be REST, but I'll spare you my rant about REST for another day).

URIs should generally identify a noun, and the verb you're executing should be identified by a verb.  There are all kinds of silly reasons that people use to justify this about "state transfers" and "read operation idempotence" and "cache friendliness" which don't really have much value (oops that REST rant leaked through a little).  However, there is a very real reason that you want to follow that pattern: namespace pollution.  The control plane may have subsidiary components of a load-balancer that we want to expose - like, for example, the attributes themselves - which might have their own identifiers.  So `/loadbalancer/<clb_id>/setAttr` masks a potential future noun endpoint called `setAttr`.

The Nova control endpoint's equivalent API is under `/attributes/` and the equivalent verb is a POST.  It's done that way to facilitate bulk updates, and the noun involved is the *attribute* rather than the server because it makes it easier to simulate data-center conditions like a whole rack failing at once, which is something that might happen with load balancers as well.

I think mirroring that structure as closely as possible would make this easier to use for people learning Mimic.  However, it's possible that we would want to support multiple styles of this for Nova, which is why the present control endpoint is under `/attributes/`, to preserve the `/servers/` namespace for individual-server inspection and update functionality.

So, in keeping with the "noun" being operated on here being "attributes" and the operation being "change some of them", I think this endpoint should live at `/v2/<string:tenant_id>/loadbalancer/<int:clb_id>/attributes/`.  `POST` is a good catch-all "do stuff" HTTP verb, and it's what we used in the Nova stuff so it's probably fine.  (Putting on my protocols wonk hat for a moment though, really the *correct* correct verb here would be [`PATCH`](http://datatracker.ietf.org/doc/rfc5789/?include_text=1) because this is a partial update to an existing object which should always fail if the object being identified doesn't exist already.)

mimic/rest/loadbalancer_api.py
glyph: In a sense, it *does* implement endpoints; as in, this is the plugin which dispatches to the relevant objects that implements endpoints.  I guess "implements" is an unhelpful word here.

mimic/rest/loadbalancer_api.py
glyph: Oh, ugh, I see the code you copied it from.  Probably code I wrote.  This is terrible but it's something I should clean up later, so please feel free to disregard this comment for now.

glyph: To be clear: leave the documentation in place here, because at least that way it's consistent and I'll find it more easily when updating.

(I was about to file a new issue but after looking at it I will address the poor documentation in #158 .)

mimic/model/clb_objects.py
sam-falvo: I'll revert the change.  Thanks.

mimic/model/clb_objects.py
sam-falvo: All good points; my goal was to more clearly express intent in the REST and test modules.  However, you're right in that they can mistakenly mislead a coder into thinking that the regional classes are sufficiently container-like that they can be used interchangably with them.

I'll rename the methods that I use in my code (and eliminate those required to just implement the minimum interface requirements for containership).

mimic/model/clb_objects.py
sam-falvo: I had intended the REST handler to perform input validation for this call, but I see the point.  Changes shortly.

mimic/rest/loadbalancer_api.py
sam-falvo: Removing this logic causes exceptions to be thrown elsewhere.  If this is a significant concern, I recommend opening a technical debt ticket as a separate action item.

mimic/rest/loadbalancer_api.py
sam-falvo: It's more compact only for the most naive use-cases.  If I want to use any validators, holy cow does it get wordy.  Very, very wordy.  :(

mimic/rest/loadbalancer_api.py
sam-falvo: I added methods which were decocorated with `@app.route`, and none of them worked (just kept getting 404s).  I agree this class would route the requests to objects with endpoint implementations, but I don't think it's built to provide implementations of endpoints.  That explains why I chose the wording I did.

mimic/rest/loadbalancer_api.py
sam-falvo: I fixed this in a commit earlier today; not sure why GIthub is still listing this as a concern.  :(

mimic/test/test_loadbalancer.py
sam-falvo: Thanks for the string.  I just used it verbatim.

mimic/rest/loadbalancer_api.py
glyph: Yes, that's fine.

mimic/model/clb_objects.py
```diff
@@ -14,6 +14,17 @@
                                                  _delete_node)
 
 
+class BadKeysError(Exception):
+    """
+    When trying to alter the settings of a load balancer, this exception will
+    be raised if you misspell or otherwise attempt to alter an attribute which
```
glyph: minor nitpick: "misspell or otherwise" are redundant words here; simply saying "attempt to alter an attribute which doesn't exist" is pretty comprehensive documentation :)

mimic/rest/loadbalancer_api.py
glyph: Personally, I would just remove this line, but if you feel strongly we can leave it for now. Regardless we probably need some kind of document laying out plugin architecture, a more formal version of @cyli's very helpful diagram wiki page.

sam-falvo:
YAY!  At long last, I finally got coveralls to pass!

sam-falvo:
Travis failed due to MacApp failing, and doesn't appear to be related to changes to this PR.

glyph:
Looks like Travis is happy; I don't see any mac app failures - I guess it was another "build failed to start" issue?

glyph:
I think this is good enough for now.  Merging!

PR #275, opened 2015-05-20
Support cpu test check and alarm

tokunbo:


glyph:
Tests are passing but there are a bunch of lint failures - https://travis-ci.org/rackerlabs/mimic/jobs/63265755 - `tox` locally can you spot these before travis does :).

glyph:
Assuming that this is needed by CI…

glyph:
OK, time to do a real review then :)

mimic/canned_responses/maas_alarm_examples.py
```diff
@@ -294,3 +294,20 @@ def alarm_examples():
             "type": "whole number (may be zero padded)"
         }]
     }]
+
+
+def alarm_examples_template(metric, crit_threshold, warn_threshold):
+    """
+    There's actually a POST version of this to ask for a specific alarm example
```
glyph: This docstring doesn't really explain anything about what the function does.

We should start getting better about adding `:param:` and `:return:` declarations for what all of our functions take and return, and since I can't understand what this is doing, it seems like this would be a good place to start :).

mimic/canned_responses/maas_alarm_examples.py
```diff
@@ -294,3 +294,20 @@ def alarm_examples():
...
+    alarm_templates['agent.cpu_usage_average'] = {
+        "bound_criteria": "if (metric['usage_average'] > CRIT_THRESHOLD) {\nreturn new AlarmStatus" +
+        "(CRITICAL,'CPU usage is #{usage_average}%, above your critical threshold of CRIT_THRESHOLD%')" +
+        ";\n}\n\nif (metric['usage_average'] > WARN_THRESHOLD) {\nreturn new AlarmStatus(WARNING, 'CPU" +
+        "usage is #{usage_average}%, above warning threshold of WARN_THRESHOLD%');\n}\nreturn new " +
+        "AlarmStatus(OK, 'CPU usage is #{usage_average}%, below your warning threshold of " +
+        "WARN_THRESHOLD%');\n"
+        .replace("CRIT_THRESHOLD", crit_threshold).replace("WARN_THRESHOLD", warn_threshold)
+    }
+    return alarm_templates[metric]
```
glyph: Is the intention for this function to raise a `KeyError` if passed anything other than `agent.cpu_usage_average` for `metric`?

mimic/canned_responses/maas_test_check.py
```diff
@@ -0,0 +1,67 @@
...
+
+
+def test_check(checkType):
+    """
+    Support test-check API
+    """
+    test_check_responses = {}
+    test_check_responses["agent.cpu"] = [
+        {
+            "timestamp": int(time.time() * 100),
```
glyph: Nothing in mimic should use the real `time.time` - this function needs to get passed an `IReactorTime` so that the current synthetic time used by the rest of mimic matches it.

glyph: (In other words, you need to do `session_store.clock.seconds()` and not `time.time()`)

mimic/test/test_maas.py
```diff
@@ -586,6 +586,37 @@ def test_list_monitoring_zones(self):
         mz = data['values'][0]
         self.assertEquals('mzdfw', mz['id'])
 
+    def test_test_check(self):
```
glyph: Please write a docstring for this test; all tests in mimic require docstrings.

mimic/test/test_maas.py
```diff
@@ -586,6 +586,37 @@ def test_list_monitoring_zones(self):
...
+        postdata = {"type": "agent.cpu", "label": "test check", "details": {},
+                    "target_alias": None, "target_resolver": None, "target_hostname": None}
+        req = request(self, self.root, "POST",
+                      self.uri + '/entities/' + self.entity_id + '/test-check', json.dumps(postdata))
+        resp = self.successResultOf(req)
+        data = self.get_responsebody(resp)
+        self.assertEquals("success", data[0]["status"])
+
+    def test_test_alarm(self):
```
glyph: test docstring please :)

mimic/test/test_maas.py
```diff
@@ -597,6 +628,15 @@ def test_list_alarm_examples(self):
         ax = data['values'][0]
         self.assertEquals('remote.http_body_match_1', ax['id'])
 
+    def test_alarm_examples_template(self):
```
glyph: test docstring please :)

mimic/rest/maas_api.py
```diff
@@ -333,6 +334,43 @@ def list_entities(self, request, tenant_id):
         request.setResponseCode(200)
         return json.dumps({'metadata': metadata, 'values': entities})
 
+    @app.route('/v1.0/<string:tenant_id>/entities/<string:entity_id>/test-check', methods=['POST'])
+    def test_check(self, request, tenant_id, entity_id):
+        """
+        Replies the test-check
```
glyph: Can you add a documentation citation for this so that we can easily reference what this API is actually supposed to do?

mimic/rest/maas_api.py
```diff
@@ -333,6 +334,43 @@ def list_entities(self, request, tenant_id):
...
+        """
+        postdata = json.loads(request.content.read())
+        result = test_check(postdata['type'])
+        request.setResponseCode(200)
+        return json.dumps(result)
+
+    @app.route('/v1.0/<string:tenant_id>/entities/<string:entity_id>/test-alarm', methods=['POST'])
+    def test_alarm(self, request, tenant_id, entity_id):
+        """
+        Replies the test-alarm
```
glyph: An upstream documentation citation would be helpful for this as well.

mimic/rest/maas_api.py
```diff
@@ -935,6 +973,17 @@ def list_alarm_examples(self, request, tenant_id):
         request.setResponseCode(200)
         return json.dumps({'values': axs, 'metadata': metadata})
 
+    @app.route('/v1.0/<string:tenant_id>/alarm_examples/<string:metric>', methods=['POST'])
+    def alarm_examples_template(self, request, tenant_id, metric):
+        """
+        specific alarm example template
```
glyph: Specific to what? What kind of "template"?  Is there actually an endpoint called `alarm_examples` in MAAS?

glyph:
Thanks for the PR, @tokunbo !

Overall, I think this might be OK implementation-wise, but I feel like this change lacks enough documentation for future Mimic maintainers to actually be able to make progress on maintaining or evolving this stuff because it's not clearly documented enough. Please clean up stuff, mainly focusing on the docstrings, so that a new maintainer can more quickly come up to speed with what this code is doing and why.

mimic/canned_responses/maas_test_check.py
```diff
@@ -0,0 +1,67 @@
+"""
+Canned response for /test-check
+"""
+
+import collections
+
+
+def test_check(checkType):
+    """
+    Support test-check API
```
glyph: This documentation isn't sufficient.  What does it mean to support the "test-check API"?  What is the parameter type? What is the return type?  How will a future maintainer understand this in context?

mimic/canned_responses/maas_test_check.py
```diff
@@ -0,0 +1,67 @@
+"""
+Canned response for /test-check
+"""
+
+import collections
+
+
+def test_check(checkType):
```
glyph: `checkType` ought to be `check_type` for coding-standard compliance; Mimic uses PEP 8 naming conventions.

mimic/canned_responses/maas_test_check.py
```diff
@@ -0,0 +1,67 @@
...
+    """
+    Support test-check API
+    """
+    test_check_responses = collections.defaultdict(lambda: "")
```
glyph: Synthesizing a `defaultdict` and then throwing it away is a slightly odd way to implement this.

A more straightforward way to do this would be to move `test_check_responses` out to a global variable, declare it at the top level of the module, as `test_check_responses = { "agent.cpu": [ ... ]}` and then have this function simply do `from copy import deepcopy; deepcopy(test_check_responses.get(check_type, ""))`.

mimic/canned_responses/maas_test_check.py
```diff
@@ -0,0 +1,67 @@
...
+    test_check_responses["agent.cpu"] = [
+        {
+            "timestamp": 0,
+            "available": True,
+            "status": "success",
+            "metrics": {
+                "user_percent_average": {
+                    "data": 7.1695253641707,
```
glyph: I don't understand where these magic values are coming from.  Should they be documented? Do they have any significance?  If they're not significant, could they just be made random instead of static?

glyph:
I deleted the old comments you've already addressed, but I'm still a little concerned about the state of the documentation for the canned response.  (Otherwise looking good though.)

tokunbo:
@glyph  Is there an example somewhere I can look at?

mimic/canned_responses/maas_test_check.py
tokunbo: I won't pretend to understand all of the hard-coded values in any of my mimic PRs either. I just know that when I ask the real production endpoints from a real rackspace-cloud server with a real agent, that's what is returned to me and it makes the cloud-intelligence UI happy. I don't want random stuff that may or may not be valid.

glyph: I had a conversation with @tokunbo in person about this and realized that I did not really make my concerns here clear, so let me spell it out in more detail.

My concern is that any time Mimic produces static, fixed values as part of a mock, that is an opportunity for a test, usually accidentally, to start depending on those values.  Once that happens, if Mimic were to start producing more "correct" values, whether synthetic, injected by a control-plane mutation, or computed by a control-plane behavior, test suites would break and it would become difficult for us to change to make things more accurate.

So my major concern about landing this as-is is that we start depending on these magic values, both in Cloud Intelligence's test suite (again, probably not now, probably in a future change that *accidentally* starts depending on one value being a certain number, or within a certain range), and then we have to live with them forever, and it's hard to explain to future contributors to the cloud monitoring mock that `7.1695253641707` is a number they should care about for some reason :).

One way to address this, hopefully quickly, is to return a deterministic pattern of generated numbers for these values, in a certain range.  If we start adding real, system randomness, this could make Mimic's tests hard to read (since gathering real entropy is a non-deterministic I/O operation).

glyph: Right now, `MimicCore` holds on to the one object which generates I/O events - timeouts - within mimic; `MimicCore.clock`.  We could expand this to also hold on to a `random` instance, and the code in question here could do something like `core.rng.uniform(MIN_CPU_VAL, MAX_CPU_VAL)`.

glyph: Also, important to note, `core.rng` should be a `random.Random(...)` instance owned by `MimicCore`, and not just a reference to the `random` module, since the `random` module's seed is global mutable shared state and therefore unpredictable in the presence of any other libraries which might use it to generate a random number.

glyph:
I expanded my comments above about my concerns about the fixed values.  Hopefully that can be addressed without dumping an unreasonable amount of work on you, @tokunbo ?

mimic/canned_responses/maas_test_check.py
ksheedlo: TBH, and as an Intelligence dev I feel like this is a good use case for the backplane. For now while we're still developing the backplane, we should be intentional about the values we return. After all, these are mocked metrics, not some mystical constants handed down by the programming gods. The last thing I would want to do as an Intelligence dev is do something forward incompatible such that Mimic is stuck with `7.1695253641707` forever.

For now, I feel being intentional about our output means not going crazy with decimal places (max 1 after the decimal point is great!) and leaving clear documentation to the effect that Mimic may change these values at any time without warning.

ksheedlo:
Checking in on this PR, since APIs around it are causing us headaches in Intelligence. Is there anything I can do to help us land?

glyph:
@ksheedlo - really just my comments above, about concern surrounding accidentally depending on the canned values.  Any kind of solution to that issue would probably be OK to get this to land.

tokunbo:
Note that C.Intelligence is now on my branch https://github.com/rackerlabs/mimic/commits/toksPatches It's basically only divergent on this PR plus some other minor things that aren't really needed for the tests, except for that new getAlarm. At some point later, probably after all the UAT stuff and when I'm back from Anime Expo, I'm probably going to pair-program with Lekha to get this PR and all the stuff in my branch merged.

ksheedlo:
@tokunbo I'm more concerned that `toksPatches` is 86 commits behind `master`.

glyph:
If we still plan to get this merged, it should probably still be open.  Unless you mean you're going to rebase these commits or something?

tokunbo:
I have no plans on fixing this PR. If this turns out to be truly needed, it'll be restarted using the control-plane approach. Probably by someone else.

Issue #278, opened 2015-05-21
☂ get rid of nth_endpoint_public


glyph:
As pointed out on #276, any test calling `TenantAuthentication.nth_endpoint_public` will start failing as soon as anyone changes anything about that test's service catalog.

Umbrella issue because first we have to kill all the callers and we shouldn't do that in one big bang, I think there are a lot of them.

- [x] #276 
- [x] #279 
- [x] #280 
- [x] #281

glyph:
Nothing left but to merge #282…

PR #280, opened 2015-05-21
use get_service_endpoint instead of nth_endpoint_public in RackConnectTestMixin.get_lb_ids

glyph:
See #278


cyli:
LGTM, thank you!

PR #281, opened 2015-05-21
replace nth_endpoint_public in test_same_tenant_different_regions

glyph:

PR #279, opened 2015-05-21
use get_service_endpoint instead of nth_endpoint_public in test_different_tenants_same_region_different_lbs

glyph:
Fixes #276

PR #282, opened 2015-05-21
get rid of nth_endpoint_public completely

### LINKED ISSUE #278: ☂ get rid of nth_endpoint_public

glyph:
Fixes #278 

This should only be merged after the other pending PRs - #279 #280 #281 - are all merged, since it includes all of them.

### DIFF
--- a/mimic/test/fixtures.py
+++ b/mimic/test/fixtures.py
@@ -34,16 +34,6 @@ class TenantAuthentication(object):
             }
         ))
 
-    def nth_endpoint_public(self, n):
-        """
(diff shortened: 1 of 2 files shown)


cyli:
LGTM!

Issue #262, opened 2015-05-11
Support error behavior in identity


cyli:
For instance if identity returning 500, or 401.

PR #283, opened 2015-05-21
Refactor identity authentication and impersonation to have a default behavior

cyli:
This is in preparation for #262.

### DIFF
--- a/mimic/model/identity.py
+++ b/mimic/model/identity.py
@@ -66,17 +66,15 @@ class PasswordCredentials(object):
         """
         Given an authentication JSON blob, which should look like:
 
-        ```
-        {
-            "auth": {
-                "passwordCredentials": {
-                    "username": "user",
-                    "password": "pass"
-                },
-                "tenantId": "111111"
+            {
+                "auth": {
+                    "passwordCredentials": {
+                        "username": "user",
+                        "password": "pass"
+                    },
+                    "tenantId": "111111"
+                }
             }
-        }
-        ```
 
         ``"tenantId"`` is interchanable with ``"tenantName"``, and is
         optional.
@@ -115,17 +113,15 @@ class APIKeyCredentials(object):
         """
         Given an authentication JSON blob, which should look like:
 
-        ```
-        {
-            "auth": {
-                "RAX-KSKEY:apiKeyCredentials": {
-                    "username": "user",
-                    "apiKey": "key"
-                },
-                "tenantId": "111111"
+            {
+                "auth": {
+                    "RAX-KSKEY:apiKeyCredentials": {
+                        "username": "user",
+                        "apiKey": "key"
+                    },
+                    "tenantId": "111111"
+                }
             }
-        }
-        ```
 
         ``"tenantId"`` is interchanable with ``"tenantName"``, and is
         optional.
@@ -163,16 +159,14 @@ class TokenCredentials(object):
         """
         Given an authentication JSON blob, which should look like:
 
-        ```
-        {
-            "auth": {
-                "token": {
-                    "id": "my_token"
-                },
-                "tenantId": "111111"
+            {
+                "auth": {
+                    "token": {
+                        "id": "my_token"
+                    },
+                    "tenantId": "111111"
+                }
             }
-        }
-        ```
 
         ``"tenantId"`` is interchanable with ``"tenantName"``, and is
         optional.
@@ -216,16 +210,14 @@ class ImpersonationCredentials(object):
         """
         Given an impersonation JSON blob, which should look like:
 
-        ```
-        {
-            "RAX-AUTH:impersonation": {
-                "expire-in-seconds": 1000,
-                "user": {
-                    "username": "my_user"
+            {
+                "RAX-AUTH:impersonation": {
+                    "expire-in-seconds": 1000,
+                    "user": {
+                        "username": "my_user"
+                    }
                 }
             }
-        }
-        ```
 
         Along with a header "X-Auth-Token" with the impersonator's token,
 
--- a/mimic/model/identity.py
+++ b/mimic/model/identity.py
@@ -1,7 +1,11 @@
 """
 Models relating to identity.
 """
-from attr import attributes, attr, validators
+from uuid import uuid4
+
+from attr import Factory, attributes, attr, validators
+
+from six import string_types, text_type
 
 from zope.interface import Interface, implementer
 
@@ -36,12 +40,6 @@ class ICredentials(Interface):
         :param session_store: a :class:`mimic.session.SessionStore`
         """
 
-    def from_json(json_blob):  # pragma:nocover
-        """
-        Given a JSON dictionary containing credentials, creates a ICrednetials
-        object.
-        """
-
 
 @implementer(ICredentials)
 @attributes
@@ -49,8 +47,8 @@ class PasswordCredentials(object):
     """
     An object representing username/password + optional tenant credentials
     """
-    username = attr()
-    password = attr()
+    username = attr(validator=validators.instance_of(string_types))
+    password = attr(validator=validators.instance_of(string_types))
     tenant_id = attr(default=None)
 
     def get_session(self, session_store):
@@ -98,8 +96,8 @@ class APIKeyCredentials(object):
     """
     An object representing username/api-key + optional tenant credentials
     """
-    username = attr()
-    api_key = attr()
+    username = attr(validator=validators.instance_of(string_types))
+    api_key = attr(validator=validators.instance_of(string_types))
     tenant_id = attr(default=None)
 
     def get_session(self, session_store):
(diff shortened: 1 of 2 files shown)


mimic/rest/auth_api.py
```diff
@@ -2,38 +2,107 @@
...
+                     session.tenant_id, prefix_map,
+                     base_uri_from_request(http_request))),
+                prefix_for_endpoint=lookup,
+                response_token=session.token,
+                response_user_id=session.user_id,
+                response_user_name=session.username,
+            )
+            return json.dumps(result)
+    else:
+        session = credentials.get_session(core.sessions)
```
cyli: Reviewer: should impersonation be an authentication behavior?
Arguments for:

1.  It is, in a sense, authenticating - it is doing some session creation.
1.  It could use the same behavior criterion as the other authentication creds (username, token)

Arguments against:

1.  It's technically the "auth token" not a token, which comes from the headers and not the JSON blob.

mimic/model/identity.py
```diff
@@ -20,3 +26,213 @@ def from_store(cls, session_store, tenant_id):
...
+        """
+        return session_store.session_for_api_key(
+            self.username, self.api_key, self.tenant_id)
+
+    @classmethod
+    def from_json(cls, json_blob):
+        """
+        Given an authentication JSON blob, which should look like:
+
+        ```
```
glyph: this is Markdown, not ReST :). We really need that documentation builder…

mimic/model/identity.py
```diff
@@ -20,3 +26,213 @@ def from_store(cls, session_store, tenant_id):
...
+        """
+        return session_store.session_for_username_password(
+            self.username, self.password, self.tenant_id)
+
+    @classmethod
+    def from_json(cls, json_blob):
+        """
+        Given an authentication JSON blob, which should look like:
+
+        ```
```
glyph: MD not ReST

mimic/model/identity.py
```diff
@@ -20,3 +26,213 @@ def from_store(cls, session_store, tenant_id):
...
+        :param session_store: a :class:`mimic.session.SessionStore`
+        """
+        return session_store.session_for_token(self.token, self.tenant_id)
+
+    @classmethod
+    def from_json(cls, json_blob):
+        """
+        Given an authentication JSON blob, which should look like:
+
+        ```
```
glyph: MD not ReST

mimic/model/identity.py
```diff
@@ -20,3 +26,213 @@ def from_store(cls, session_store, tenant_id):
...
+            self.expires_in,
+            self.impersonator_token,
+            self.impersonated_token)
+
+    @classmethod
+    def from_json(cls, json_blob, auth_token):
+        """
+        Given an impersonation JSON blob, which should look like:
+
+        ```
```
glyph: MD not ReST

cyli:
Also, it'd be really nice to maybe define an input schema or something at some point.

mimic/model/identity.py
```diff
@@ -20,3 +26,205 @@ def from_store(cls, session_store, tenant_id):
...
+class ICredentials(Interface):
+    """
+    An :obj:`ICredentials` provides identity authentication credentials.
+    """
+    def get_session(session_store):  # pragma:nocover
+        """
+        Get a session corresponding to the user and tenant from the given
+        session store.
+
+        :param session_store: a :class:`mimic.session.SessionStore`
```
glyph: Given the perpetual confusion in the area of "what even is a Session", a `:return:` or `:rtype:` would definitely not be misplaced.

mimic/model/identity.py
```diff
@@ -20,3 +26,205 @@ def from_store(cls, session_store, tenant_id):
...
+        """
+        return session_store.session_for_username_password(
+            self.username, self.password, self.tenant_id)
+
+    @classmethod
+    def from_json(cls, json_blob):
+        """
+        Given an authentication JSON blob, which should look like:
```
glyph: On the "ReST versus Markdown" front, I think you need a second `:` at the end of the line here to make the next section a code block.

mimic/model/identity.py
```diff
@@ -20,3 +26,205 @@ def from_store(cls, session_store, tenant_id):
...
+            }
+
+        ``"tenantId"`` is interchanable with ``"tenantName"``, and is
+        optional.
+
+        :return: a class:`APIKeyCredentials` object
+        """
+        creds = json_blob['auth']
+        tenant_id = (creds.get('tenantName', None) or
+                     creds.get('tenantId', None))
```
glyph: Minor nitpick: if you have a falsey `tenantName` or `tenantId` (`""` or `0` or `None` in JSON-land) this will give effectively a wrong result.  Given that bad results are Mimic's whole bailiwick, it might actually be important to preserve this and have a test for it and stuff.  Possibly in a separate issue.  But for now, you can shorten the implementation by doing `creds.get('tenantName', creds.get('tenantId'))` to give this the proper priority.

mimic/model/identity.py
```diff
@@ -20,3 +26,205 @@ def from_store(cls, session_store, tenant_id):
...
+        username = creds['RAX-KSKEY:apiKeyCredentials']['username']
+        api_key = creds['RAX-KSKEY:apiKeyCredentials']['apiKey']
+        return cls(username=username, api_key=api_key, tenant_id=tenant_id)
+
+
+@implementer(ICredentials)
+@attributes
+class TokenCredentials(object):
+    """
+    An object representing token + optional tenant credentials
```
glyph: Another minor nitpick; the token is really the "credential" here, whereas the optional part is the `tenant_id` which is the *identifier*.  The gramar here just makes it a little ambiguous what's optional.

mimic/model/identity.py
```diff
@@ -20,3 +26,205 @@ def from_store(cls, session_store, tenant_id):
...
+    """
+    token = attr(validator=validators.instance_of(string_types))
+    tenant_id = attr(validator=validators.instance_of(string_types))
```
glyph: I thought that `attr`'s validators couldn't support nullability.  Are you actually testing the no-tenant-ID case?

mimic/model/identity.py
```diff
@@ -20,3 +26,205 @@ def from_store(cls, session_store, tenant_id):
 
     identity_admin = attr(default=False,
                           validator=validators.instance_of(bool))
+
+
+class ICredentials(Interface):
```
glyph: This is really just me thinking out loud, but - wouldn't this interface be singular? `ICredential`?  Each token or password or whatever is just one credential…

mimic/rest/auth_api.py
```diff
@@ -2,38 +2,107 @@
...
 from mimic.session import NonMatchingTenantError
-from mimic.canned_responses.mimic_presets import get_presets
+from mimic.util.helper import invalid_resource
 
 Request.defaultContentType = 'application/json'
 
 
-class AuthApi(object):
+authentication = EventDescription()
+"""
```
glyph: I believe that if you want this docstring to document `authentication` it should be in the module docstring as a `:var:`.

mimic/model/identity.py
cyli: It does not support nullability.  You have to provide either a `tenantId` or `tenantName`:  http://docs.rackspace.com/auth/api/v2.0/auth-client-devguide/content/POST_authenticate_token_v2.0_tokens_Token_Calls.html

mimic/rest/auth_api.py
cyli: There does seem to be an option to support this format of docstrings:  http://sphinx-doc.org/ext/autodoc.html#directive-autoattribute.  I think the version of Sphinx we are using supports it (1.2.3), but maybe we are not passing the right directive.

glyph:
I really like the `ICredential`(`s`) refactoring, but would it be possible without too much trouble to split this up into a patch that just adds `ICredential` and puts the credential-parsing into objects, and then a second one for the behaviors stuff?

glyph:
Mainly I would like to think about the highly subtle question about impersonation while looking at a patch which *just* adds the behaviors.

cyli:
> Mainly I would like to think about the highly subtle question about impersonation while looking at a patch which just adds the behaviors.

Could you explain this comment?  Should I not do any impersonation stuff until error injection is in place?

glyph:
> Could you explain this comment? Should I not do any impersonation stuff until error injection is in place?

By "subtle question about impersonation", I mean, "should impersonation be a second `EventDescription` object or the same one as authentication?".  So I'm asking if you can put the parse-credentials-into-objects stuff into the first PR (including `ImpersonationCredentials`), and the do-authentication-using-behaviors stuff into the second PR.

glyph:
I will be thinking about whether impersonation is a separate behavior event in the meanwhile :)

glyph:
Most everything I can see about the new impersonation objects is a totally straightforward improvement so I think we can land that reasonably quickly.

mimic/model/identity.py
glyph: Yeah, OK. Thanks for the explanation, this is fine.

mimic/rest/auth_api.py
glyph: Let's not block this PR on this feedback, but we should figure it out.

cyli:
Sounds good.  Closing in favor of newer smaller PRs.

PR #284, opened 2015-05-21
“Changes Since” filter

### LINKED ISSUE #252: Changes-since for nova server details
http://docs.rackspace.com/servers/api/v2/cs-devguide/content/ChangesSince.html

manishtomar:
Closes  #252.

- When deleting server, it is marked as DELETED instead of removing it from cache and update_time is updated with latest clock seconds
- The update_time is also updated when changing status or metadata of server
- `GET ../servers/detail` honors changes-since filter based on update_time. It was trivial to implement it for `GET ../servers` but I didn't do it due lazyness of writing similar tests. Morever, autoscale currently uses only detail call. So, `../servers` can be implemented later when required.

### DIFF
--- a/mimic/model/nova_objects.py
+++ b/mimic/model/nova_objects.py
@@ -762,7 +762,7 @@ class RegionalServerCollection(object):
 
         if changes_since is not None:
             since = timestamp_to_seconds(changes_since)
-            to_be_listed = filter(lambda s: s.update_time > since, to_be_listed)
+            to_be_listed = filter(lambda s: s.update_time >= since, to_be_listed)
 
         # marker can be passed without limit, in which case the whole server
         # list, after the server that matches the marker, is returned
--- a/mimic/model/nova_objects.py
+++ b/mimic/model/nova_objects.py
@@ -751,7 +751,8 @@ class RegionalServerCollection(object):
         """
         Request the list JSON for all servers.
 
-        :param str changes_since: ISO8601 formatted datetime
+        :param str changes_since: ISO8601 formatted datetime. Based on
+            http://docs.rackspace.com/servers/api/v2/cs-devguide/content/ChangesSince.html
 
         Note: only supports filtering by name right now, but will need to
         support more going forward.
--- a/mimic/model/nova_objects.py
+++ b/mimic/model/nova_objects.py
@@ -15,6 +15,7 @@ from six import string_types
 from mimic.util.helper import (
     seconds_to_timestamp,
     random_string,
+    timestamp_to_seconds
 )
 
 from mimic.model.behaviors import (
@@ -218,6 +219,13 @@ class Server(object):
             }
         }
 
+    def set_metadata(self, metadata):
+        """
+        Replace all metadata with given metadata
+        """
+        self.metadata = metadata
+        self.update_time = self.collection.clock.seconds()
+
     def set_metadata_item(self, key, value):
         """
         Set a metadata item on the server.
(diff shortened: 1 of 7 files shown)


cyli:
iso8601 probably needs to be added to setup.py to make the tests pass.  :)

mimic/model/nova_objects.py
```diff
@@ -739,6 +760,10 @@ def request_list(self, http_get_request, include_details, absolutize_url,
         """
         to_be_listed = self.servers
 
+        if changes_since is not None:
+            since = timestamp_to_seconds(changes_since)
+            to_be_listed = filter(lambda s: s.update_time > since, to_be_listed)
```
cyli: Non-blocking - hard to test with Nova, but should this be `>` or `>=` ?

cyli: No I'm crazy, sorry, we can test with Nova.  :D  It seems to be `>=` (I made a change, took the update time, used that as the `changes-since` param, and it returned that one server)

mimic/test/test_nova.py
```diff
@@ -47,6 +48,53 @@ def quick_create_server(helper, region="ORD"):
     )["server"]["id"]
 
 
+def delete_server(helper, server_id):
```
cyli: Awesome!  These would be very helpful elsewhere too.

mimic/model/nova_objects.py
```diff
@@ -728,10 +747,12 @@ def request_ips(self, http_get_ips_request, server_id):
         return dumps({"addresses": server.addresses_json()})
 
     def request_list(self, http_get_request, include_details, absolutize_url,
-                     name=u"", limit=None, marker=None):
+                     name=u"", limit=None, marker=None, changes_since=None):
         """
         Request the list JSON for all servers.
 
+        :param str changes_since: ISO8601 formatted datetime
```
cyli: Non-blocking - wondering if a link should be provided to `http://docs.rackspace.com/servers/api/v2/cs-devguide/content/ChangesSince.html`?

mimic/test/test_util.py
```diff
@@ -50,23 +55,28 @@ def test_seconds_to_timestamp_default_timestamp(self):
...
                    ("%Y-%m-%d", "1970-01-01"),
                    ("%H %M %S (%f)", "00 00 00 (000000)")]
-        for match in matches:
-            self.assertEqual(match[1],
-                             helper.seconds_to_timestamp(0, match[0]))
+        for fmt, timestamp in formats:
+            self.assertEqual(timestamp,
+                             helper.seconds_to_timestamp(0, fmt))
+
+    def test_timestamp_to_seconds(self):
```
cyli: Non-locking - Nova also accepts incomplete timestamps (e.g. `"1970-01-01"`, `"1970-01-01T00"`) - should we add tests to support those too?

manishtomar: Yes. Good idea.

mimic/model/nova_objects.py
manishtomar: This function even though currently used for changes-since should be independent. Will it help if the link is provided in `request_link` function?

cyli: Erm, sorry, not sure I know where the `request_link` function is.  :)

mimic/model/nova_objects.py
manishtomar: [Yes](https://github.com/openstack/nova/blob/master/nova/db/sqlalchemy/api.py#L1978) :)

mimic/model/nova_objects.py
manishtomar: Sorry.. my bad. I meant `request_list` function in model that is actually using this function.

cyli: Oh, yes, sorry - I thought this comment was in the `request_list` function - at least that's what github seems to be showing me.  It may be messing up the display though.  :)  Yes I meant for this comment to go in the `request_list` function.

cyli:
This looks great - thanks @manishtomar!  Merging.

PR #285, opened 2015-05-22
Refactor auth API to use credential objects

cyli:
Split off from #283.
Part of #262.

This creates credential objects for authenticating against Identity, and uses those in the auth api.

Was hoping to just use the objects in the existing code, but determining whether tenant ID is present is handled in the credentials objects, which means if there was an error creating the credentials object, then it's an invalid input error/bad request and should return a 400.

Adding in that logic made the function too complex and it failed to lint. Hence the more heavyweight refactoring of `get_token_and_service_catalog`

### DIFF
--- a/mimic/model/identity.py
+++ b/mimic/model/identity.py
@@ -89,8 +89,10 @@ class PasswordCredentials(object):
 
         :return: a class:`PasswordCredentials` object
         """
-        tenant_id = (json_blob['auth'].get('tenantName', None) or
-                     json_blob['auth'].get('tenantId', None))
+        tenant_id = json_blob['auth'].get('tenantName')
+        if tenant_id is None:
+            tenant_id = json_blob['auth'].get('tenantId')
+
         username = json_blob['auth']['passwordCredentials']['username']
         password = json_blob['auth']['passwordCredentials']['password']
         return cls(username=username, password=password, tenant_id=tenant_id)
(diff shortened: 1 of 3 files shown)


mimic/model/identity.py
```diff
@@ -20,3 +26,228 @@ def from_store(cls, session_store, tenant_id):
 
     identity_admin = attr(default=False,
                           validator=validators.instance_of(bool))
+
+
+class ICredential(Interface):
```
glyph: Since this is an interface, it should really go into `imimic.py`.

mimic/rest/auth_api.py
```diff
@@ -88,35 +90,33 @@ def lookup(entry):
...
                                   exception.session.user_id))
 
-        if content['auth'].get('passwordCredentials'):
-            username = content['auth']['passwordCredentials']['username']
-            password = content['auth']['passwordCredentials']['password']
-            return format_response(
-                lambda: self.core.sessions.session_for_username_password(
-                    username, password, tenant_id),
-                username_generator)
+        cred_mappings = {
```
glyph: Just a thought for future PRs: this mapping here is really populated with keys that are mostly within the domain of the credentials themselves, i.e. `RAX-KSKEY:apiKeyCredentials` is a string that really only `APIKeyCredentials` ought to know about.  So this should really be constructing this dictionary by looping over `PasswordCredentials`/`APIKeyCredentials`/`TokenCredentials`… and once it's abstracted out like that, you could even have a plugin API :).  (Which might be useful given the Kerberos and OAuth stuff I keep hearing rumours about)

glyph:
If you could just move the interface to the interfaces module, then I think this is good to go; please do so and land after a successful build.

mimic/rest/auth_api.py
cyli: Looping over them and doing a try/except to construct them?

cyli: Hm... I wonder what happens if you pass more than 1 type of credential to nova....

glyph: Not even a try/except necessarily.  Preserving this exact structure, just something like:

```python
for credential_type in PasswordCredentials, APIKeyCredentials, TokenCredentials:
    cred_mappings[credential_type.type_key] = (credential_type, credential_type.default_generator)
```

cyli: Apparently multiple authentications are cool.  But `tenantId` + `tenantName` is not.

cyli: https://github.com/rackerlabs/mimic/issues/286

PR #287, opened 2015-05-22
squarewave PING graph

tokunbo:
For potential future graph testing. Want a static, yet unique-looking, graph for possible screen-image compare tests.

### DIFF
--- a/mimic/rest/maas_api.py
+++ b/mimic/rest/maas_api.py
@@ -253,6 +253,11 @@ def create_multiplot_from_metric(metric, reqargs, allchecks):
     """
     Given a metric, this will produce fake datapoints to graph
     This is for the multiplot API call
+
+    Also, if you name a PING check "squarewave", you will get
+    datapoints that form such a wave. This is for testing the
+    graphing-solution you have in place. Produces a static, yet
+    unique-looking, graph for screen comparison solutions.
     """
     fromdate = int(reqargs['from'][0])
     todate = int(reqargs['to'][0])
--- a/mimic/rest/maas_api.py
+++ b/mimic/rest/maas_api.py
@@ -258,6 +258,7 @@ def create_multiplot_from_metric(metric, reqargs, allchecks):
     todate = int(reqargs['to'][0])
     points = int(reqargs['points'][0])
     multiplot = {}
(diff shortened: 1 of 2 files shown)


glyph:
The implementation seems OK, but this really needs some accompanying documentation so that others will know how to use it in the future.

mimic/rest/maas_api.py
```diff
@@ -276,7 +282,15 @@ def create_multiplot_from_metric(metric, reqargs, allchecks):
                     d = {}
                     d['numPoints'] = 4
                     d['timestamp'] = timestamp
-                    d['average'] = random.randint(1, 99)
+                    if not q % (points / 4):
+                        squarewave_downtrend = squarewave_downtrend ^ 1
```
glyph: minor nitpick: this can be more succinctly expressed as `squarewave_downtrend ^= 1`.

mimic/rest/maas_api.py
```diff
@@ -253,11 +253,17 @@ def create_multiplot_from_metric(metric, reqargs, allchecks):
     """
     Given a metric, this will produce fake datapoints to graph
     This is for the multiplot API call
+
+    Also, if you name a PING check "squarewave", you will get
```
glyph: This documentation leaves open a lot of questions, but they're things that would be perhaps more obvious if the function were generally documented better.  For example, what makes a check a "PING check"?  I would have at least said a ``` ``"remote.ping"`` check ``` to make it clear that's what the check's `type` ought to be, and said "label" rather than "name" to give a hint that it's the "label" field and not, say, a "name" field.

That said, the real issue here is that `metric`, `reqargs`, and `allchecks` are not documented.  Can you open a follow-up ticket to add `:param:` and `:return:` documentation to this function?

glyph:
Just checking - if Cloud Intelligence needs this, can you add the requisite label?

glyph:
I think if you open the follow-up issues (and hopefully work on them ;-)) this is OK to land.

tokunbo:
@glyph I'm going to need to see an example of what you're looking for somewhere in this repo.

glyph:
I will go ahead and land this and open the follow-ups later.

PR #288, opened 2015-05-22
Provide a default authentication behavior

cyli:
And use it for authenticating username/password/token/api key creds, and for impersonation.  The second of of #283, which got split up into multiple PRs.

Part of fixing #262

### DIFF
--- a/mimic/rest/auth_api.py
+++ b/mimic/rest/auth_api.py
@@ -4,6 +4,8 @@ Defines get token, impersonation
 """
 import json
 
+import attr
+
 from twisted.web.server import Request
 from twisted.python.urlpath import URLPath
 from mimic.canned_responses.auth import (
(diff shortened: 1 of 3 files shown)


mimic/rest/auth_api.py
```diff
@@ -21,24 +23,86 @@
...
+"""
+Event refers to authenticating against Identity using a username/password,
+username/api-key, token, or getting an impersonation token.
+"""
 
+
+@authentication.declare_default_behavior
+def default_authentication_behavior(core, http_request, credentials):
+    """
+    Default behavior in response to a server creation.
```
glyph: This should really say what the default behavior *is* (provide a token).

mimic/rest/auth_api.py
```diff
@@ -21,24 +23,86 @@
...
+
+    :param core: An instance of :class:`mimic.core.MimicCore`
+    :param http_request: A twisted http request/response object
+    :param credentials: An `mimic.model.identity.ICredentials` provider
```
glyph: Given that this is the main documentation of what the signature for the `authentication` event is, it should really also include a `:return:` field as well.

mimic/rest/auth_api.py
```diff
@@ -48,75 +112,23 @@ def get_token_and_service_catalog(self, request):
...
-                return format_response(cred, error_handler)
+            pass
+        else:
+            for cred_type in (PasswordCredentials, APIKeyCredentials,
+                              TokenCredentials):
+                if cred_type.type_key in content['auth']:
+                    try:
+                        cred = cred_type.from_json(content)
+                    except Exception:
+                        pass
```
glyph: `except:pass` is always a real red flag :).  `from_json` should raise a *specific* exception type, and this `except` should catch only that type.

glyph:
The things I spotted above are pretty minor so please land this after addressing them.

PR #289, opened 2015-05-26
put back python 2.6 build on linux

glyph:
`tox` presently has configuration for doing Python 2.6 builds, so travis ought to as well.  If we're going to eliminate one we should eliminate both at the same time.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,8 @@ env:
 
 matrix:
     include:
+        - python: 2.6
+          env: TOXENV=py26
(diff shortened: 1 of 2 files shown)


sam-falvo:
Pending Travis' approval, looks good to me.  Thanks for helping!

glyph:
Thanks for the review @sam-falvo !

glyph:
Oops, I totally should have put "ready for review" on there.  Thanks nevertheless :).

PR #290, opened 2015-05-26
Refactor all of the auth tests to use core_and_root

cyli:
Does what it says on the box.  :)


glyph:
Seems pretty simple.  Although it looks like a lot of these tests could use further refactoring to reduce duplication…

Issue #271, opened 2015-05-14
Make behavior injection a real REST interface


cyli:
- [x] When a behavior is specified, provide an ID for it.
- [ ] Allow listing of existing behaviors
- [x] Be able to delete behaviors by ID

Otherwise if Mimic isn't restarted, behaviors can persist across test runs.

cyli:
Note - my plan is to turn the register creation behavior/list creation behavior/delete creation behavior into a decorator (???), so that people don't have to write all these endpoints again from scratch for other plugins' control planes.  So it'd look something like:

```
@behaviors({"creation": server_creation_event,
                      "deletion": server_deletion_event})
class NovaBehaviorControl(object):
    "pass"
```

This would be the bottom-level Twisted resource, that adds endpoints like `/<event-name>/<behavior_id>` only.  This way the control plane handles the rest of the URL routing.

glyph:
That looks like a good plan to me.

cyli:
@glyph Thoughts on providing a standard suite of tests?  Or should the tests on the utility function be good enough?

glyph:
@cyli Already typing that - github needs a better way to do multi-part replies :smile:

glyph:
> Suggestions/thoughts? I am not sure how to provide standard tests that work with all kinds of plugins and failures, though...

The way to do this is to parameterize the test suite with the parts that vary per suite.  So for example, in this case, one thing that differs per system is the "create some JSON for a valid criterion" step, and another is the "discover the control plane endpoint" step.  Have a function that creates a `TestCase` class where you have to give it an object that knows how to create some valid criterion, and then has a `test_` method which creates a valid criterion, discovers the control plane endpoint, then, in test code common to all backends, POSTs to the behaviors/creation endpoint.

Does that make sense?

cyli:
@glyph Wondering if this should just be a partially-implemented class that others can subclass?

glyph:
@cyli sigh; I wrote out that whole paragraph without saying "inheritance" or "subclass" once. but yeah, that's one way to do it.

cyli:
@glyph Yes yes I know.  :)  I just want to make sure that all the FQDNs for the test methods are right, and not some anonymous `TestCase`

glyph:
A decorator, maybe?

cyli:
Eh, listing is not done, but we can probably close this ticket since we can create and delete.  We can add an issue for listing if necessary later - not sure it's that useful.

PR #293, opened 2015-05-26
Generate an ID for each behavior that is added.

cyli:
Part of #271.

In the interest of smaller PRs, this PR only returns the UUID when the behavior is created and does not support listing or deleting by ID.

But it is actually not that small, because I also snuck in a conversion of `behaviors.py` from `characteristic` to `attrs` (since otherwise this would be like 5 lines).

### DIFF
--- a/mimic/rest/nova_api.py
+++ b/mimic/rest/nova_api.py
@@ -151,14 +151,19 @@ class NovaControlApiRegion(object):
                     "message": "Stuff is broken, what"
                 }
             }
+
+        The response looks like:
+
+            {
+                "id": "this-is-a-uuid-here"
+            }
         """
         behavior_description = json.loads(request.content.read())
         region_collection = self._collection_from_tenant(tenant_id)
-        region_collection.create_behavior_registry.register_from_json(
-            behavior_description
-        )
+        behavior_id = (region_collection.create_behavior_registry
+                       .register_from_json(behavior_description))
         request.setResponseCode(CREATED)
-        return b''
+        return json.dumps({'id': text_type(behavior_id)})
 
     @app.route("/v2/<string:tenant_id>/attributes/", methods=['POST'])
     def change_attributes(self, request, tenant_id):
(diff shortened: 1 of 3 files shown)


cyli:
Marking as RFR since most of the tox envs have passed, just waiting on mac and pypy.

mimic/rest/nova_api.py
```diff
@@ -151,14 +151,19 @@ def register_creation_behavior(self, request, tenant_id):
                     "message": "Stuff is broken, what"
                 }
             }
+
+        The response looks like:
```
glyph: I think you're missing a colon here.

mimic/test/test_nova.py
```diff
@@ -1374,12 +1374,13 @@ def use_creation_behavior(self, name, parameters, criteria):
...
-                               json.dumps(criterion))
-        set_criteria_response = self.successResultOf(set_criteria)
-        self.assertEqual(set_criteria_response.code, 201)
+        set_criteria = json_request(self, self.root, "POST",
+                                    self.nova_control_endpoint +
+                                    "/behaviors/creation/",
+                                    json.dumps(criterion))
+        response, body = self.successResultOf(set_criteria)
+        self.assertEqual(response.code, 201)
+        self.assertTrue(body.get("id"))
```
glyph: `assertTrue` is just a *tad* too weak of an assertion for the behavior that the docstring of the method alleges to add.  I'd do something like this:

```python
behavior_id = body.get("id")
self.assertIsInstance(behavior_id, text_type)
parsed = uuid.UUID(hex=behavior_id)
self.assertEquals(parsed.version, 4)
```

I also assume clients are going to want to do something with the behavior ID, although it's fine to leave that for a future PR.

glyph:
Aside from the one minor test issue I think this is fine.

mimic/test/test_nova.py
cyli: Yep, adding support for deleting behaviors in a future PR.  Would also be good to list at some point.  Will make the test change, thanks!

Issue #199, opened 2015-03-12
Tox uses a different version of lint, doc, docspell requirements compared to what in dev-requirements


lekhajee:
Possibly create different requirements files for lint, doc, docspell and use the `deps = -r{toxinidir}/lint-requirements.txt` and such

glyph:
A couple of thoughts:

- Put all the requirements into a `requirements` directory, then we can have  `requirements/run.txt`, `requirements/lint.txt`, `requirements/dev.txt`, `requirements/all.txt` and not clutter up the top level directory.
- Any requirements which require other requirements (development requires running, for example) can use `-r other-requirements.txt`
- Each requirements file should have a comment at the top saying what exactly you need it for.

glyph:
When installing Mimic itself, we have two use-cases.

One is to deploy on exact, known versions of everything - which is what our continuous infrastructure should be doing; generally pinning everything.  These exact versions of dependencies should go in `requirements.txt` and should be updated explicitly by pull request when they become outdated.  I just set up requires.io for this project, so we can start to get visibility into that: https://requires.io/github/rackerlabs/mimic/requirements/?branch=master

Another is to deploy on an abstract, constrained set of requirements, which is what our `setup.py` should be doing.  `setup.py` should be abstract, even if we don't test every possible supported configuration, because we may have users who want or need to install mimic into the same environment as other things with higher version requirements, and we shouldn't prohibit that.

glyph:
When installing mimic's development dependencies, we want to have a single source of truth which you can pull from for random development command-line stuff.  Right now there is `dev-requirements.txt` as well as various sections in `tox.ini`.

glyph:
I just turned on automatic pull requests on requires.io… hopefully that means we'lll see an automatic PR come in soon that updates our dependencies.

glyph:
#497 attempts to address this.

glyph:
I think #497 actually addressed these problems reasonably well.

glyph:
Feel free to reopen if someone disagrees, but I guess I will proactively close this :)

PR #294, opened 2015-05-26
Remove dev-requirements

cyli:
Since it wasn't really being maintained, and the testing requirements were installed in `setup.py` anyway.

And just copied all the requirements from `setup.py` to `requirements.txt`.

Question: should we just get rid of `requirements.txt` and everyone can just `pip install -e .`?


glyph:
I will put comments on #199 since I think that's what this is a part of :)

glyph:
Or rather, I'll put comments relevant to the general design there.

glyph:
Can you revert out the parts of this which change `requirements.txt`?  The reason that `requirements.txt` is different from `setup.py` is that `requirements.txt` should always be pinned and `setup.py` should never be unless we know something is broken.  See [the canonical holy text](https://caremad.io/2013/07/setup-vs-requirement/) on this subject.

cyli:
@glyph Ah ok, sure.  This change also adds testtools to `requirements.txt` - should I pin that version?  Or should I leave it out of `requirements.txt` entirely?

glyph:
Pinning that version would be good.

glyph:
It looks like `getting-started.rst` still has a couple of references to `dev-requirements.txt`; those need to be updated to do something else.  And this is why `dev-requirements.txt` still exists, I think - there isn't really a good way to do what those instructions propose if we *don't* ship a `dev-requirements.txt`.  So I think this is going to need to be more holistic in how it addresses things, and actually move the requirements out of the travis config and into a `requirements/` directory as independent `.txt` files.

cyli:
Closing this in favor of a better solution.

PR #295, opened 2015-05-27
Refactor Nova tests

cyli:
This PR:

1.  Factors out the create server code into two helper functions.
1.  Factors out the use creation behavior code from individual TestCases.

### DIFF
--- a/mimic/test/test_nova.py
+++ b/mimic/test/test_nova.py
@@ -76,7 +76,7 @@ def create_server(helper, body, region="ORD", json=False):
         helper.test_case,
         helper.root,
         "POST",
-        '{}/servers'.format(helper.get_service_endpoint(
+        '{0}/servers'.format(helper.get_service_endpoint(
             "cloudServersOpenStack", region)),
         body
     )
@@ -153,7 +153,7 @@ def use_creation_behavior(helper, name, parameters, criteria):
                  "criteria": criteria}
     set_criteria = json_request(
         helper.test_case, helper.root, "POST",
-        "{}/behaviors/creation".format(
+        "{0}/behaviors/creation".format(
             helper.auth.get_service_endpoint("cloudServersBehavior")),
         json.dumps(criterion))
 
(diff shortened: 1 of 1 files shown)


mimic/test/test_nova.py
```diff
@@ -35,20 +34,67 @@ def status_of_server(test_case, server_id):
...
+        helper.root,
+        "POST",
+        '{0}/servers'.format(helper.get_service_endpoint(
+            "cloudServersOpenStack", region)),
+        body
+    )
+    return helper.test_case.successResultOf(create_server)
+
+
+def quick_create_server(helper, region="ORD", name=None):
```
glyph: Now that there's another kind of server creation to contrast it with, "quick"ness doesn't seem to be its defining property. I am thinking that this should be something like `id_for_new_server`?

mimic/test/test_nova.py
```diff
@@ -35,20 +34,67 @@ def status_of_server(test_case, server_id):
     return get_server_response_body['server']['status']
 
 
-def quick_create_server(helper, region="ORD"):
+def server_args(name=None, imageRef=None, flavorRef=None,
```
glyph: I'd like to avoid polluting the namespace with too many extra names of helper functions (we already have a ton that we need, so extras are bad).  It seems like every single invocation of `server_args` is passed as the second argument to `create_server`, and no calls to `create_server` are ever made without also calling `server_args`.  Whenever I see two functions being used together like that, it seems to me that they're really just two parts of the same whole and they ought to be presented as such.

In this case, that would mean `create_server` should take all the keyword args of `server_args` and do the serialization itself.  If a case were to arise where we needed to provide a bad request body (for example to test 400 replies) then we could add an additional `raw_body` kwarg to `create_server` which would take precedence.

What do you think?

glyph:
Looking at the patch overall I think I'm going to land this, since it is an improvement by itself, but I'll file a follow-up issue for the further refactoring.

PR #297, opened 2015-05-27
Add an endpoint for deleting a create server behavior by ID

cyli:
Which would allow a test to clean up injected error behavior.

Part of #271

### DIFF
--- a/mimic/model/behaviors.py
+++ b/mimic/model/behaviors.py
@@ -184,6 +184,6 @@ class BehaviorRegistry(object):
         for i, behaviors in enumerate(self.registered_behaviors):
             b, c, b_id = behaviors
             if b_id == behavior_id:
-                self.registered_behaviors.pop(i)
+                del self.registered_behaviors[i]
                 return
         raise NoSuchBehaviorError(behavior_id=behavior_id)
--- a/mimic/model/behaviors.py
+++ b/mimic/model/behaviors.py
@@ -2,9 +2,18 @@
 General-purpose utilities for customizing response behavior.
 """
 import re
-from uuid import uuid4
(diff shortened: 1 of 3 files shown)


mimic/model/behaviors.py
```diff
@@ -167,3 +176,14 @@ def behavior_for_attributes(self, attributes):
...
         return self.event.default_behavior
+
+    def remove_behavior_by_id(self, behavior_id):
+        """
+        Remove a previously-registered behavior given the behavior's ID.
+        """
+        for i, behaviors in enumerate(self.registered_behaviors):
+            b, c, b_id = behaviors
+            if b_id == behavior_id:
+                self.registered_behaviors.pop(i)
```
radix: stylistically I would just use `del self.registered_behaviors[i]` instead of `pop` since you're not using the result

mimic/rest/nova_api.py
```diff
@@ -165,6 +166,28 @@ def register_creation_behavior(self, request, tenant_id):
...
+        The response is a 204 with no body if successful.
+
+        If the behavior does not exist, the response is a 404 with no body.
+        """
+        region_collection = self._collection_from_tenant(tenant_id)
+        try:
+            region_collection.create_behavior_registry.remove_behavior_by_id(
+                UUID(behavior_id)
+            )
+        except (ValueError, NoSuchBehaviorError):
```
radix: Why do you need to catch ValueError?

radix: Never mind. UUID.

radix:
LGTM. Just a super minor stylistic nitpick :)

PR #300, opened 2015-06-01
Generic behavior API generator

cyli:
Part of #271 

This only implements the ~~decorator~~ API generation function, and shows that existing tests pass.  Will provide generic tests in another PR.


glyph:
After looking at this for quite a while I don't have much in the way of feedback - it looks great :).

PR #301, opened 2015-06-03
Refactor nova tests so that only a single create server function is used.

### LINKED ISSUE #296: further refactor nova tests by removing server_args
see pre-merge feedback on #295

cyli:
Also removed `server_args` as per #296.
Fixes #296 

Also snuck in a change so that only the single top-level `use_creation_behavior` function is used.

### DIFF
--- a/mimic/model/nova_objects.py
+++ b/mimic/model/nova_objects.py
@@ -714,7 +714,7 @@ class RegionalServerCollection(object):
             behavior = self.create_behavior_registry.behavior_for_attributes({
                 "tenant_id": self.tenant_id,
                 "server_name": creation_json["server"]["name"],
-                "metadata": creation_json["server"].get("metadata")
+                "metadata": creation_json["server"].get("metadata", {})
             })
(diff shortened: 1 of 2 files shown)


mimic/model/nova_objects.py
```diff
@@ -714,7 +714,7 @@ def request_creation(self, creation_http_request, creation_json,
             behavior = self.create_behavior_registry.behavior_for_attributes({
                 "tenant_id": self.tenant_id,
                 "server_name": creation_json["server"]["name"],
-                "metadata": creation_json["server"].get("metadata")
+                "metadata": creation_json["server"].get("metadata", {})
```
cyli: Note that this was added in because if metadata was None before, using behaviors seemed to fail (unit tests failed).  This came up as a result of removing the test-specific `create_server` function which always provided a default empty dictionary for metadata.

glyph:
LGTM

glyph:
(*Very* nice that this spotted a bug.)

PR #302, opened 2015-06-03
Provide a way to automatically generate CRUD behavior tests for any behavior event

cyli:
And convert the Nova creation behavior CRUD tests to use it.

Part of #271

### DIFF
--- a/mimic/test/behavior_tests.py
+++ b/mimic/test/behavior_tests.py
@@ -1,0 +1,198 @@
+"""
+Module that contains utilities to automatically generate or help generate
+some tests.
+"""
+import json
+
+from uuid import UUID, uuid4
+
+from six import text_type, string_types
+
+from twisted.trial.unittest import SynchronousTestCase
+
+from zope.interface import Attribute, Interface
+
+from mimic.test.helpers import json_request, request_with_content
+
+
+class IBehaviorAPITestHelper(Interface):
+    """
+    Helper class that provides some setup and assertion methods that tests
+    for behavior CRUD need.  An instance will be created and
+    used for each CRUD test case.
+    """
+    root = Attribute("The root resource for mimic.")
+    behavior_api_endpoint = Attribute(
+        "The API endpoint for behaviors for the event to be tested.")
+    criteria = Attribute("""
+    A list of criteria for the behaviors to be injected.  Example::
+
+        [
+            {"criteria1": "regex_pattern.*"},
+            {"criteria2": "regex_pattern.*"},
+        ]
+    """)
+    name_and_params = Attribute("""
+    An list of 1 or 2 tuples of name and parameters, which together
+    with the criteria, can form a behavior specification.  Any more than
+    2 will ignored.  Example::
+
+        # tuples of (name of the behavior, parameters for the behavior)
+        [
+            ("fail", {"code": 500, "message": "Stuff is broken"}),
+            ("fail", {"code": 404, "message": "Stuff doesn't exist"})
+        }
+    """)
+
+    def trigger_event_and_validate_injected_behavior(name, test_case):
+        """
+        Make a request to the regular API that this behavior event modifies,
+        and validate that the behavior was triggered as opposed to the default
+        behavior or some other behavior.  Note that the event should conform
+        to the criteria given.
+
+        :param name: The name of the behavior to trigger - this will be one
+            of the names provided in :ivar:`name_and_params`.
+        :param test_case: An instance of a test case that has assertion
+            and cleanup functions.
+        """
+
+    def trigger_event_and_validate_default_behavior(test_case):
+        """
+        Make a request to the regular API that this behavior event modifies,
+        and validate that the default behavior was triggered as opposed to
+        some other behavior.  Note that the event should conform
+        to the criteria given, otherwise we won't be able to test that
+        deleting a behavior will return to the default behavior.
+
+        :param test_case: An instance of a test case that has assertion
+            and cleanup functions.
+        """
+
+
+def register_behavior(test_case, root, uri, behavior_name, parameters,
+                      criteria):
+    """
+    Register a particular behavior.
+
+    :param test_case: the test case with which to make assertions
+    :param root: A mimic root API object
+    :param str uri: The uri fo the behavior resource to register.
+    :param str behavior_name: The name of the behavior
+    :param dict parameters: A dictionary of parameters to pass to the behavior
+    :param list criteria: The criteria for which this behavior should be
+        applied.
+
+    :return: The behavior ID of the registered behavior.
+    """
+    behavior_json = {"name": behavior_name,
+                     "parameters": parameters,
+                     "criteria": criteria}
+    response, body = test_case.successResultOf(json_request(
+        test_case, root, "POST", uri, json.dumps(behavior_json)))
+
+    test_case.assertEqual(response.code, 201)
+    behavior_id = body.get("id")
+    test_case.assertIsInstance(behavior_id, string_types)
+    test_case.assertEqual(UUID(behavior_id).version, 4)
+    return behavior_id
+
+
+def make_behavior_tests(behavior_helper_klass):
+    """
+    :param behavior_helper_lkass: a class that implements
+        :class:`IBehaviorAPITestHelper`
+
+    Will generate a test suite containing:
+
+    - 1 test case that deleting a behavior will revert to the underlying
+        behavior.  If 2 behaviors are provided, will also assert that the
+        first behavior registered will be the behavior used, and that deleting
+        it means the second mock takes over.  Deleting that will revert to
+        default behavior.  If only 1 behavior is provided, it will just test
+        that deleting it reverts to default behavior.
+
+    - 1 test case validating that deleting an invalid behavior for will
+      result in a 404.
+
+    - 1 test case asserting that providing invalid JSON will result in a 400
+      when creating the behavior.
+
+    :return: a :class:`twisted.trial.unittest.SynchronousTestCase` object
+        containing the above tests, and named "TestsFor<behavior_helper name>"
+    """
+    class Tester(SynchronousTestCase):
+        """Tests for behavior API crud that uses {}""".format(
+            behavior_helper_klass.__name__)
+
+        def setUp(self):
+            self.bhelper = behavior_helper_klass()
+
+        def delete_behavior(self, behavior_id, status=204, expected_body=b''):
+            """
+            Given a behavior ID, attempts to delete it.
+            """
+            response, body = self.successResultOf(request_with_content(
+                self, self.bhelper.root, "DELETE",
+                "{0}/{1}".format(self.bhelper.behavior_api_endpoint,
+                                 behavior_id)))
+            self.assertEqual(response.code, status)
+            self.assertEqual(body, expected_body)
+
+        def test_deleting_nonexistant_creation_behavior_fails(self):
+            """
+            Deleting a non-existant behavior ID fails with a 404.  Similarly
+            with an invalid behavior ID.
+            """
+            for invalid_id in (text_type(uuid4()), "not-even-a-uuid"):
+                self.delete_behavior(invalid_id, status=404)
+
+        def test_providing_invalid_json_fails_with_400(self):
+            """
+            Providing invalid JSON for the behavior registration request
+            results in a 400.
+            """
+            one_invalid = dict(self.bhelper.behaviors[0])
+            del one_invalid["criteria"]
+
+            for invalid in ('', '{}', one_invalid):
+                response, body = self.successResultOf(request_with_content(
+                    self, self.bhelper.root, "POST",
+                    self.bhelper.behavior_api_endpoint,
+                    invalid))
+                self.assertEqual(response.code, 400)
+                self.assertEqual(b"", body)
+
+        def test_deleting_creation_behavior_removes_top_behavior(self):
+            """
+            If deleting a behavior succeeds, and there were other behaviors
+            the first behavior was masking, then the next behavior is used.
+            When there are no more behaviors, the default behavior is used.
+            """
+            names_and_params = self.bhelper.names_and_params[:2]
+            behavior_ids = []
+            for i, name_and_params in enumerate(names_and_params):
+                name, params = name_and_params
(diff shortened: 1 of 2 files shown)


glyph:
Looks like this broke py26.

mimic/test/behavior_tests.py
```diff
@@ -0,0 +1,213 @@
+"""
+Module that contains utilities to automatically generate or help generate
```
glyph: I think that "Module that contains utilities to" is a phrase that doesn't convey any additional meaning.  I think something like "Automatically generate tests for behaviors." would be better?

mimic/test/behavior_tests.py
```diff
@@ -0,0 +1,213 @@
...
+    test_case.assertEqual(response.code, 201)
+    behavior_id = body.get("id")
+    test_case.assertIsInstance(behavior_id, string_types)
+    test_case.assertEqual(UUID(behavior_id).version, 4)
+    return behavior_id
+
+
+def make_behavior_tests(behavior_helper_klass):
+    """
+    :param behavior_helper_lkass: a class that implements
```
glyph: Field declarations should generally go after the description.

mimic/test/behavior_tests.py
```diff
@@ -0,0 +1,213 @@
...
+    """
+    Helper class that provides some setup and assertion methods that tests
+    for behavior CRUD need.  An instance will be created and
+    used for each CRUD test method.
+    """
+    root = Attribute("The root resource for mimic.")
+    behavior_api_endpoint = Attribute(
+        "The API endpoint for behaviors for the event to be tested.")
+    criteria = Attribute("""
+    A list of criteria for the behaviors to be injected.  Example::
```
glyph: This is a silly nitpick (let's not address it now) but the docstrings for `Attribute` should probably be indented so they're not flush with the indentation level where they're being declared.

mimic/test/behavior_tests.py
```diff
@@ -0,0 +1,213 @@
...
+            ("fail", {"code": 404, "message": "Stuff doesn't exist"})
+        }
+    """)
+
+    def trigger_event():
+        """
+        Make a request to the regular API that this behavior event modifies.
+        Note that the event should conform to the criteria given.
+
+        :return: a tuple of (response, body)
```
glyph: What is the type of `response` and `body` here?

mimic/test/behavior_tests.py
```diff
@@ -0,0 +1,213 @@
...
+        """
+
+    def validate_injected_behavior(name_and_params, response, body):
+        """
+        Validate that the named behavior was triggered as opposed to the
```
glyph: "validate" is a bit vague: is it meant to return a boolean? raise an exception?

mimic/test/behavior_tests.py
```diff
@@ -0,0 +1,213 @@
...
+    def validate_default_behavior(response, body):
+        """
+        Validate that the default behavior was triggered as opposed to
+        some other behavior.
+
+        :param response: The response from triggering the event.
+        :param body: The response body from triggering the event.
+        """
+
+    def from_test_case(test_case):
```
glyph: Technically speaking this is not a valid way to declare an `Interface`.  An interface defines the behavior on a single object, and `from_test_case` is not a method on that same object.  There should be a separate interface, `IBehaviorAPITestHelperFactory`, which describes this one method.

mimic/test/behavior_tests.py
```diff
@@ -0,0 +1,213 @@
...
+from twisted.trial.unittest import SynchronousTestCase
+
+from zope.interface import Attribute, Interface
+
+from mimic.test.helpers import json_request, request_with_content
+
+
+class IBehaviorAPITestHelper(Interface):
+    """
+    Helper class that provides some setup and assertion methods that tests
```
glyph: I think we could get rid of "Helper class that provides some setup and assertion methods that" and instead simply say:

> "Set-up and assertions required for testing behavior CRUD."

It might be good to have a reference to `make_behavior_tests` here, because what this docstring is *really* trying to say is "that thing you need for `make_behavior_tests` and not really for anything else".

mimic/test/behavior_tests.py
```diff
@@ -0,0 +1,213 @@
...
+    test_case.assertEqual(response.code, 201)
+    behavior_id = body.get("id")
+    test_case.assertIsInstance(behavior_id, string_types)
+    test_case.assertEqual(UUID(behavior_id).version, 4)
+    return behavior_id
+
+
+def make_behavior_tests(behavior_helper_klass):
+    """
+    Will generate a test suite containing:
```
glyph: Two minor nitpicks here:

1. beginning with a verb like "will" or "should" or "does" or "handles" (etc) doesn't add information.  "Generate a test suite containing:" is equally helpful.
2. every bullet point starts with "1 test case" and then either implicitly or explicitly "validating".  A cleaner parallel construction would get rid of that and just say "Generate a test suite containing test cases that validate:"

mimic/test/behavior_tests.py
```diff
@@ -0,0 +1,213 @@
...
+    - 1 test case validating that deleting an invalid behavior for will
+      result in a 404.
+
+    - 1 test case asserting that providing invalid JSON will result in a 400
+      when creating the behavior.
+
+    :param behavior_helper_klass: a class that implements
+        :class:`IBehaviorAPITestHelper`
+
+    :return: a :class:`twisted.trial.unittest.SynchronousTestCase` object
```
glyph: The type notation for the thing that is returned here would be a `SynchronousTestCase` *subclass*; saying an "object" vaguely implies an instance.

glyph:
All of the above feedback points are just docstring nitpicking, except for the one about `from_test_case`; if you can address that one consider the rest optional.

glyph:
Since the thing about the interface documentation might be a little confusing, let me expand.

When you are describing an interface, you are describing a thing you might *already have*; this is what we call a "provider" of that interface.  So you can't describe class methods, or constructors, since those are things that you have a reference to when you have a reference to an *implementer*, not a *provider*, of that interface.  In some cases (like in the case of a dynamic proxy that is passing through methods to another object) there *is* no "implementer", and you *just* have a provider.

Sometimes you need module-level objects (like classes, or functions, or maybe even just global instances that get built out of pieces) that you're going to interact with directly, which is the case here.  You want a thing that *provides* the Factory interface so you can get something that *provides* the Helper interface.

Zope Interface does have a separate `provider` decorator which can be used to make this explicit for a class for function, as well as `directlyProvides` to specify it for an instance.

The long way to spell this out in a declaration is

```python
from mimic.tests.behavior_tests import make_behavior_tests
from zope.interface import implementer, provider
@make_behavior_tests
@implementer(IBehaviorAPITestHelper)
@provider(IBehaviorAPITestHelperFactory)
class MyBehaviorStuff(object):
    "..."
```

but that is obviously pretty wordy, so you could call `implementer` and `provider` yourself, either in an explicit step, like

```
def behavior_helper_factory(cls):
    return implementer(IBehaviorAPITestHelper)(provider(IBehaviorAPITestHelperFactory)(cls))
# ...
from mimic.tests.behavior_tests import make_behavior_tests, behavior_helper_factory
@make_behavior_tests
@behavior_helper_factory
class MyBehaviorStuff(object):
    "..."
```

or even have a higher-level API which is just `@behavior_tests_for_class` which calls all three things (`make_behavior_tests`, `implementer`, `provider`).

I'm probably in favor of that last option, given that this is the way the API will be used 99% of the time.

mimic/test/behavior_tests.py
cyli: This has been addressed in 0b2f1ac

mimic/test/behavior_tests.py
```diff
@@ -0,0 +1,242 @@
...
+
+class IBehaviorAPITestHelper(Interface):
+    """
+    Set-up and assertion methods required for testing behavior CRUD.
+    :class:`IBehaviorAPITestHelperFactory.from_test_case` will be called at
+    the beginning of every CRUD method, the intent of which is to generate
+    a new instance should this be a class.  If the provider is not a class,
+    then please keep in mind the intended usage.
+
+    The reason there are two interfaces, both clearly intended to be
```
glyph: The reason that there are two interfaces is that there are two separate responsibilities.  One thing creates a `Helper` thing, the other thing provides the behavior of that `Helper` thing.  Diamond inheritance, while always a nasty edge-case, is not the only reason to avoid inheritance.  For example, if Trial itself had a clean implementer/provider distinction around how test cases are discovered and instantiated, rather than requiring discoverable test suites to be literal class objects inheriting from an `ITestCase` implementor, this code would be a lot simpler to implement and would be using test-suite APIs rather than special attributes like `__name__` and `__module__`.

I gather that you added this comment in response to the fact that factory-pattern interfaces are a bit more verbose to document with zope.interface than other kinds of interface, because you need a class/class docstring/method/method docstring as opposed to just class/class docstring as with a class.  However, documenting creation patterns with a clear distinction between what is expected to be provided by the creator and what is expected to be provided by the createe is a worthwhile habit to get into.

So rather than saying that they are "clearly intended to be implemented by a single class", since there's nothing special about this interface requiring it to be one class any more than any other interface-that-returns-something, would be to have a single high-level API for the common case of using a single class, as I recommended in my previous comment. It's a little more verbosity in the implementation, but in all of the API clients, it turns multiple lines of boilerplate into a one-liner which can be clearly documented.

cyli: There are a couple reasons I think that these should both a single class:

1. Splitting documentation across multiple interfaces when they are coupled by extra requirements makes it harder to get the documentation right.  For example, I don't actually want a  `from_test_case` method, I just want an `__init__` that takes a `test_case`.  Right now I require a docstring to warn people in the interface in both `from_test_case` and in `IBehaviorAPITestHelper` not to keep any extra state around if they chose not to use a single class, because a fresh one needs to be created on every test method, whereas I can just call `__init__` if it were a single class.  I guess as per our conversation, people could do other silly things like store state to a global variable and restore it on every `__init__`, but that seems less likely.

2.  It doesn't seem like there's much use case in the near future for having a factory be separate from the helper.  It seems like splitting off the constructor as a separate object than the thing being constructed, without any concrete examples as to why or what the use case is would make this code harder to understand.

3. Even with the decorator that reduces the boilerplate, documentation is scattered across several interfaces, and it seems like maybe a lot of the documentation for `make_behavior_tests`, et. al. would have to be repeated in `behavior_tests_for_class` if we wanted to make that the friendly function everyone should use.  (e.g. if that is the method we want everyone to use most of the time, that should be the best-documented one of them all, rather than just have links to every other function and interface.)

What you suggest would be a more precise/theoretically correct interface definition though, and moving this comment to the helper decorator and making it the recommended way to generate test cases would be more useful, since this comment doesn't actually make any recommendations.

mimic/test/behavior_tests.py
```diff
@@ -0,0 +1,307 @@
...
+    #. declaring that ``klass`` also implements
+       :class:`IBehaviorAPITestHelperFactory`
+    #. calling :func:`make_behavior_tests` on ``klass``
+
+    This decorator will also validate that ``klass`` correctly implements
+    :class:`IBehaviorAPITestHelper`.  It creates an instance of ``klass`` to
+    do so.
+
+    :param klass: a class that implements :class:`IBehaviorAPITestHelper`,
+        although it does not have to declare that it does (this decorator will
```
cyli: Question: should this require that `klass` already be declared a `IBehaviorAPITestHelper`?

glyph: Short answer: no.

glyph: Long answer: I think there are two layers here.  At the lower layer (`make_behavior_tests`), you care about `implements` and `provides` and formal contracts between minimal interfaces.  At the higher level (`behavior_tests_helper_class`) you care about "write a class that looks like this, don't worry too much about it, see the lower layer if you want to do something fancy".

So if you want to step outside the nice, convenient class-based API provided by `behavior_tests_helper_class`, you should have to dive right down into the lower level and build something that does what you want.  I can't think of a single reason to pass something to this function that is *not* the correct shape to be a class-which-provides-`IBehaviorAPITestHelperFactory`-and-implements-`IBehaviorAPITestHelper `, so requiring that declaration would simply be pointless boilerplate.

So, there is a further consequence beyond "no" - although the bulleted list of what-this-is-syntactic-sugar-for is very comprehensive, and possibly interesting enough to leave in there, it should definitely be preceded by an example code snippet, both of how you'd declare a hypothetical user of this, and also possibly a snippet of what `trial`'s output should look like once you've done that.

cyli: Done

mimic/test/behavior_tests.py
```diff
@@ -0,0 +1,357 @@
...
+                else:
+                    self.test_case.assertEquals(response.code, 400)
+
+            def validate_default_behavior(self, response, body):
+                self.test_case.assertEquals(response.code, 201)
+
+
+    This will produce trial output like the following::
+
+        <class's module FQDN>.TestsForMyPluginBehaviorAPI
```
glyph: The "D" in "FQDN" refers to "domain".  The Twisted-ish jargon for this is "FQPN", where the "P" means "Python".  (Other projects tend to refer to this as "The full name of a Python class or module" or something similarly awkward; I haven't seen any other conventions emerge for this kind of string.)

glyph:
I think the module documentation should also direct people to `behavior_tests_helper_class`.

With this and the `s/FQDN/FQPN/g` comment I think this is ready to merge.

cyli:
Ok travis has been running for more than 12 hours now on the mac builds, and this was only a doc change.  The previous commit was green, and the linux builds on the linux commit were green, so I'm just going to merge.

glyph:
Even stranger, https://travis-ci.org/rackerlabs/mimic/builds/65507204 appears to have passed; only the Github integration broke on that commit.

glyph:
Just for my own personal curiosity, I am re-starting this build to see if the checkmark turns green; despite the fact that all the builds had passed on the travis-ci site proper, the yellow circle seemed stuck here.

glyph:
Aaaand travis can't build closed PRs, so it appears to be a failure over there now.  Oh well.

Issue #108, opened 2014-11-12
Support for non Openstack services


malini-kamalambal:
Currently any non Openstack API will need to go to resource.py (See example https://github.com/rackerlabs/mimic/pull/107/files). In the long run this can cause the resource.py to grow huge. Need a better solution to handle this.

glyph:
Proposal: let's have a new plugin API which uses a fixed URL - let's say `/mimicking/external/` - and then your plugin declares a namespace under there it wants to claim.

Since most "outside the service catalog" APIs are associated with a specific provider, or at least a piece of software, in order to avoid conflicts, let's use full domain names as the recommended string for the plugin.

So, for example, if someone wanted to write a Stripe mock, they would write an `IExternalAPIMock` plugin, which would have an `api_domain` attribute of `"api.stripe.com"`, and then its root resource would show up as `http://localhost:8900/mimicking/external/api.stripe.com/` and it could then have a `@route("/v1")` and so on dispatching down from there.

glyph:
(That example domain taken from https://stripe.com/docs/api )

glyph:
This should also log a message at startup listing all the external URLs that are loaded; see #69 .

glyph:
Since we already have several hard-coded top-level non-openstack services, we should continue to support both paths for a few releases before deleting the existing ones.

glyph:
In the case of a conflict between two namespaces (we might have two separate external implementations of the stripe API, for example) we should just log an error message about the conflict and let one "win".

PR #303, opened 2015-06-03
Adding mock email api plugin for ele

lekhajee:


mimic/core.py
```diff
@@ -35,6 +36,7 @@ def __init__(self, clock, apis):
         """
         self._uuid_to_api = {}
         self.sessions = SessionStore(clock)
+        self.message_store = MessageStore()
```
cyli: Non-blocking - I just wanted to put this here and make a reference to #108, so it's kept in mind when we look at that issue again.  There should probably be standard place for plugins that do not make use of sessions to store their own data too, so not everyone has to add another item to core.  (e.g. maybe core could create all the storage objects by plugin).

mimic/model/mailgun_objects.py
```diff
@@ -0,0 +1,119 @@
+"""
+Mailgun object storage
+"""
+import time
+from characteristic import attributes, Attribute
+
+
+@attributes(["message_id", "to", "msg_from", "subject", "body",
+             Attribute("headers", default_factory=dict)])
+class Messages(object):
```
cyli: Should this be `Message` (singular)?

mimic/model/mailgun_objects.py
```diff
@@ -0,0 +1,119 @@
+"""
+Mailgun object storage
+"""
+import time
+from characteristic import attributes, Attribute
+
+
+@attributes(["message_id", "to", "msg_from", "subject", "body",
+             Attribute("headers", default_factory=dict)])
```
cyli: Apologies - I was wondering what `headers` here represents.

It looks like it does not get serialized as JSON in `generate_events`, and in fact the headers only consist of to/from/subject/message-id.

But it looks like it does get returned when the someone queries `/messages/header?...`.  I apologize, I couldn't find this in the mailgun documentation - does that return only the non to/from/subject/message-id headers (in all of [these](https://tools.ietf.org/html/rfc4021#section-2))?  Or does it return only the special arbitrary headers that start with X (`h:X-My-Header` values from [the sending docs](https://documentation.mailgun.com/api-sending.html#sending)?

mimic/model/mailgun_objects.py
```diff
@@ -0,0 +1,119 @@
...
+        return template
+
+
+@attributes([Attribute("message_store", default_factory=list)])
+class MessageStore(object):
+    """
+    A collection of message objects.
+    """
+
+    def _add_to_message_store(self, **attributes):
```
cyli: Non-blocking question - out of curiosity, why a private function if it's called by the REST api?

mimic/model/mailgun_objects.py
```diff
@@ -0,0 +1,119 @@
...
+        return {
+            "items": events,
+            "paging": {
+                "next": "http://i-am-a-fake-link-to-nothing",
+                "last": "http://i-am-a-fake-link-to-nothing",
+                "first": "http://i-am-a-fake-link-to-nothing=",
+                "previous": "http://i-am-a-fake-link-to-nothing=="
+            }}
+
+    def filter_message_by_to_address(self, to_address):
```
cyli: Non-blocking question - Apologies, again, I couldn't find the API doc - just wanted to make sure that if there are 8 messages to the same address, then this will only return the latest one?

lekhajee: Yes, that was the idea. Also, apologies this `/500s` and `/headers` endpoints are not supported by Mailgun, but I added them under Mailgun instead of the control plane as it seemed like yet another addition to the `resources.py` just for the control plane.

cyli: That's true - thanks for the docstring update!

cyli:
LGTM!

PR #305, opened 2015-06-04
Identity behavior api

### LINKED ISSUE #262: Support error behavior in identity

cyli:
Use the behavior api generics to add a identity behavior control panel. Also add failure injection for identity.

Fixes #262

### DIFF
--- a/mimic/resource.py
+++ b/mimic/resource.py
@@ -91,7 +91,7 @@ class MimicRoot(object):
             "now": seconds_to_timestamp(self.clock.seconds())
         })
 
-    @app.route("/mimic/v1.1/behaviors", branch=True)
+    @app.route("/mimic/v1.1/IdentityControlAPI/behaviors", branch=True)
     def handle_identity_behaviors(self, request):
         """
         Handle creating/deleting behaviors for mimic identity.
--- a/mimic/resource.py
+++ b/mimic/resource.py
@@ -7,8 +7,13 @@ import json
 from twisted.web.resource import NoResource
 
 from mimic.canned_responses.mimic_presets import get_presets
+from mimic.model.behaviors import BehaviorRegistryCollection
 from mimic.rest.mimicapp import MimicApp
-from mimic.rest.auth_api import AuthApi, base_uri_from_request
+from mimic.rest.auth_api import (
+    AuthApi,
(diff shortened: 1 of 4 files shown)


cyli:
Travis is now waiting on the mac builds only, so putting up for review

mimic/resource.py
```diff
@@ -84,6 +91,14 @@ def advance_time(self, request):
             "now": seconds_to_timestamp(self.clock.seconds())
         })
 
+    @app.route("/mimic/v1.1/behaviors", branch=True)
```
lekhajee: For Nova we had the endpoint  `url` as `http://localhost:8900/mimicking/NovaControlApi-52627e/ORD/v2/91111` and the `name` as `cloudServersBehavior`. For consistency can we have this endpoint be `/mimic/v1.1/IdentityControlApi` or such.

Also, we support v2.0 for Identity. Any particular reason this is v1.1.

cyli: Sure, I can make it `IdentityControlAPI/behaviors` instead.

And this is v1.1 of the mimic API, as opposed to v2.0 of the identity API.  TBH I wasn't really sure where to put this,whether we wanted to put it under `/mimic/` or whether we wanted to make it `/identityControl`, like how identity is actually under `/identity`.

cyli: Also, the reason I didn't make it an extra plugin like the nova control plane was that I figured for identity, we wanted a non-random URL just like identity has a non-random URL (since probably we don't necessarily want to make someone auth first to get the service catalog before issuing identity behavior injection requests) - again, I wasn't sure about this though - maybe that's fine?

lekhajee:
`/mimic/v1.1/IdentityControlAPI/behaviors` sounds good to me from reading the reasoning behind the analogy. And I guess it makes more sense to keep it under `/mimic` rather than `identity`.

Also, I agree with not having a separate plugin for the Identity control plane and having someone auth before they could use the control plane.

Thanks so much for working on this. I tested this as well and everything looks great!

cyli:
@lekhajee Thanks for the review!  I wasn't really sure about what URL it should go under, so that was a very timely comment.  :D

PR #316, opened 2015-06-12
Provide a general utility for adding and testing sequence behavior.

cyli:
So that anything that provides behaviors can support sequence behavior.  Also, add it to identity control behaviors.

Question for the Reviewer:  I want to know your thoughts:  I originally wanted to support sequence behaviors by default.

Pros:
- all `EventDescription` instances would have a sequence behavior by default, no need to declare one - it always seems applicable
- we wouldn't have to generate tests for sequence behavior for each plugin, we could just test the sequence behavior tests for `EventDescription`

Cons:
- seems to be more explicit to require a plugin author to declare a sequence behavior for their plugin
- there wouldn't be two magic behavior names: "default" AND "sequence" - just "default"

There's also the question, in both cases, of where to put the docstring for sequence behaviors... or how to make it obvious at least.

### DIFF
--- a/mimic/model/behaviors.py
+++ b/mimic/model/behaviors.py
@@ -3,6 +3,7 @@ General-purpose utilities for customizing response behavior.
 """
 import json
 import re
+from itertools import cycle
 from uuid import UUID, uuid4
 
 import attr
@@ -85,7 +86,8 @@ class EventDescription(object):
     def declare_behavior_creator(self, name):
         """
         Decorator which declares that the decorated function is a factory,
-        taking parameters (a JSON-serialized object), and returning a behavior.
+        taking parameters (a JSON-serialized object), and returning a
+        behavior.
 
         Use like so::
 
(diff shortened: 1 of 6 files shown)


mimic/model/behaviors.py
```diff
@@ -85,7 +86,8 @@ class EventDescription(object):
     def declare_behavior_creator(self, name):
         """
         Decorator which declares that the decorated function is a factory,
-        taking parameters (a JSON-serialized object), and returning a behavior.
```
glyph: This is weird… it was already only 79 characters, did you re-wrap it intentionally?

mimic/model/behaviors.py
```diff
@@ -312,3 +314,73 @@ def delete_behavior(kl_self, request, behavior_id):
...
+def sequence_behavior(event):
+    """
+    {0}
+
+    This function, given an event, produces a generic behavior-creator that
+    provides this sequence behavior, which is named "sequence".
+
+    :param event: an instance of :class:`EventDescription`
+    :return: a callable behavior-creator as described above
+    """.format(_sequence_docstring)
```
glyph: This doesn't actually work to set a docstring of a function - it just turns into a no-op expression :-\.

mimic/model/behaviors.py
cyli: Yes - odd, my editor said it was over the limit.  I'll check again.

mimic/test/behavior_tests.py
```diff
@@ -128,6 +131,9 @@ def make_behavior_tests(behavior_helper_factory):
 
     - providing invalid JSON will result in a 400 when creating the behavior.
 
+    - sequence behavior will rotate through the behaviors provided to it (if
+      :bool:`behavior_helper_factory.test_sequence` is true)
```
glyph: One of the things that I really like about decorators as opposed to inheritance is that you don't have to do magic passing-implicit-parameters-by-setting-attributes-on-the-class, and you can just take an argument like a regular function.  I think this should be added as an argument to `behavior_tests_helper_class`.  I was going to say something about how to preserve compatibility with the old signature to allow for a graceful transition period, and then I remembered there are very few callers so far and Mimic is application code with no compatibility contract so hooray just go do it :).

glyph:
> Question for the Reviewer: I want to know your thoughts: I originally wanted to support sequence behaviors by default.

I think this is a separate PR at best.  I actually think that this *would* be a pretty cool idea, since, as you point out, it does actually generalize pretty well.

> seems to be more explicit to require a plugin author to declare a sequence behavior for their plugin

We do also need to have some kind of testing for sequence behaviors, 

> there wouldn't be two magic behavior names: "default" AND "sequence" - just "default"

Do you think maybe we should put these magic names into a namespace?  `"mimic:default"` and `"mimic:sequence"`?

cyli:
> We do also need to have some kind of testing for sequence behaviors
Hm... can't we sort of have a general test for creating a fake API with an event, and iterating through behaviors?  Or do we need one test for each plugin that uses it (in the case where we make `EventDescription` have sequence behavior by default)?

> Do you think maybe we should put these magic names into a namespace? "mimic:default" and "mimic:sequence"?

Maybe - is anyone already using "default" and "sequence" though (aside from autoscale)?

mimic/model/behaviors.py
cyli: Boo.  So should I just set it after the fact?

glyph: Yeah. Not great, but probably the best we can do with the current state of the language.

mimic/model/behaviors.py
```diff
@@ -312,3 +323,72 @@ def delete_behavior(kl_self, request, behavior_id):
...
+            )
+            for behavior in behavior_specification
+        ])
+
+        def rotating_behavior(*args, **kwargs):
+            current = next(behavior_objects)
+            return current(*args, **kwargs)
+        return rotating_behavior
+
+    sequence.__doc__ = sequence_docstring
```
glyph: Ironically, the reason that this needed to be moved to an external attribute was that it was calling `.format`.  Now that you've changed it to be the same docstring everywhere, you could put it back inline.  I'll leave it up to you which way you prefer it though.

glyph: The rule, by the way, is that Python will assign a docstring if the first statement in the suite (class or function body) is a string literal.  `"foo".format("bar")` is not a literal, it's a complex expression, so it doesn't count.  But this would.

cyli: I think I'll leave it as is, since it's clearer that its' a generated function.

Issue #270, opened 2015-05-14
Record of all errors seen with Cloud Load Balancers


cyli:
See #269 - this is the same but with CLBs.

This is a poor record, but here's one so far:

- `{"message":"Load Balancer '375263' has a status of 'ERROR' and is considered immutable.","code":422}`

cyli:
Using the following script:  

```python
"""
Script to see what responses CLB returns when it's immutable (BUILD,
PENDING_UPDATE, PENDING_DELETE, DELETED)
"""

from twisted.internet.task import deferLater, react

from twisted.internet.defer import gatherResults, inlineCallbacks

import json
import treq


@inlineCallbacks
def what_happens_when(clock, tenant_id, auth_token):
    """
    Hammer CLB while it's immutable and see what responses are.
    """
    lb_config = {
        "loadBalancer": {
            "name": "cyli-test",
            "protocol": "HTTP",
            "virtualIps": [
                {"type": "PUBLIC"}
            ]
        }
    }
    nodes = {
        "nodes": [
            {
                "address": "124.1.2.3",
                "port": 80,
                "condition": "ENABLED",
                "type": "PRIMARY"
            },
            {
                "address": "125.1.2.4",
                "port": 81,
                "condition": "ENABLED",
                "type": "SECONDARY"
            }
        ]
    }

    url = "/".join((
        "https://iad.loadbalancers.api.rackspacecloud.com/v1.0",
        tenant_id, "loadbalancers"))
    headers = {'x-auth-token': auth_token,
               'content-type': 'application/json',
               'accept': 'application/json'}

    @inlineCallbacks
    def see_how_responds(node_id='12345'):
        resps = yield gatherResults([
            treq.get('/'.join((url, lbid)), headers=headers),
            treq.post('/'.join((url, lbid, 'nodes')), headers=headers,
                      data=json.dumps(nodes)),
            treq.put('/'.join((url, lbid, 'nodes', node_id)),
                     headers=headers,
                     data=json.dumps({'condition': "DRAINING"})),
            treq.delete('/'.join((url, lbid, 'nodes')), headers=headers,
                        params={'id': node_id}),
            treq.delete('/'.join((url, lbid)), headers=headers)
        ])

        bodies = yield gatherResults(
            [treq.json_content(resp) for resp in resps])

        print "------"
        print bodies[0]['loadBalancer']['status']
        for i, thing in enumerate(["ADD N:    ", "CHANGE N: ",
[...]

Issue #269, opened 2015-05-14
Record of all seen errors from Nova


cyli:
These should be classified with:
1. where they were seen
2. conditions under which they were seen
3. date of last verification
4. maybe whether this is Rackspace specific or Openstack generalized?

Some of them also take parameters/formatting, so maybe just having a JSON structure isn't enough.  It'd also be nice if they could encapsulate some logic so we can tell if they're applicable, although they should be usable without that logic too (for error injection).

Here are all the ones autoscale has seen so far:
------

- Certain Rackspace managed customers (?) spinning up cloud servers:  
    - `403 Forbidden\\n\\nAccess was denied to this resource.\\n\\n Networks (00000000-0000-0000-0000-000000000000,11111111-1111-1111-1111-111111111111) required but missing`

- Certain Rackspace managed customers (?) spinning up cloud servers:
   -  `403 Forbidden\\n\\nAccess was denied to this resource.\\n\\n Networks (00000000-0000-0000-0000-000000000000) not allowed`

- Rackspace managed customers with RackConnect:
    - `403 Forbidden\\n\\nAccess was denied to this resource.\\n\\n Exactly 1 isolated network(s) must be attached`

- If network ID is an invalid format:
   -  `{"badRequest": {"message": "Bad networks format: network uuid is not in proper format (2b55377-890e-4fc9-9ece-ad5a414a788e)", "code": 400}}`

- Quota exceeded (the on-metal one may be Rackspace specific):
    - `{"forbidden": {"message": "Quota exceeded for ram: Requested 1024, but already used 131072 of 131072 ram", "code": 403}}`
    - `{"forbidden": {"message": "Quota exceeded for instances: Requested 1, but already used 100 of 100 instances", "code": 403}}`
    - `{"forbidden": {"message": "Quota exceeded for onmetal-compute-v1-instances: Requested 1, but already used 10 of 10 onmetal-compute-v1-instances", "code": 403}}`

- Inactive image:
    - `{"badRequest": {"message": "Image 644d0755-e69b-4ca0-b99b-3abc2f20f7c0 is not active.", "code": 400}}`

- Nonexistant image:
   -  `{"badRequest": {"message": "Can not find requested image", "code": 400}}`

- Not entirely sure - something's wrong with the image
    - `{"badRequest": {"message": "Requested image 1dff348d-c06e-4567-a0b2-f4342575979e has automatic disk resize disabled.", "code": 400}}`

- Image too big for flavor chosen:
[...]

PR #317, opened 2015-06-12
Configurable response based on ERROR status

kivattik:


mimic/model/clb_objects.py
```diff
@@ -287,6 +287,14 @@ def add_node(self, node_list, lb_id, current_timestamp):
             _verify_and_update_lb_state(self, lb_id, False, current_timestamp)
 
             if self.lbs[lb_id]["status"] != "ACTIVE":
+                # This is one possible error response per the "Rackspace Cloud
```
cyli: This seems to be an error that only occurs with creating new load balancers (since otherwise, the load balancer doesn't need to assign any IPs). When adding nodes, if the load balancer is in ERROR state, it seems to be the "Load Balancer ... has a status of ERROR and is considered immutable." ( https://github.com/rackerlabs/mimic/issues/270.)

Could you clarify the goal of this PR?  Is it to inject arbitrary failure modes when adding a node to a CLB?

mimic/rest/loadbalancer_api.py
```diff
@@ -167,6 +167,48 @@ def set_attributes(self, request, tenant_id, clb_id):
...
             return b''
 
+    @app.route(
+        '/v2/<string:tenant_id>/loadbalancer/<int:clb_id>/error_reason',
+        methods=['POST']
+    )
+    def set_error_reason(self, request, tenant_id, clb_id):
+        """
+        Sets a configurable response code and message that will be returned
+        when the load balancer is in an ERROR state.
```
glyph: I'm a little concerned about providing super flexible APIs like this.

One of Mimic's purposes is to provide a flexible test backend.  However, Mimic should also serve as a repository of useful error responses; things that we know can happen for documented reasons, and keep those responses updated as backend services change.

For example, I think it was a mistake to make the Nova mock's "failure" behavior so flexible (allowing an arbitrary status code and message); Nova cannot in reality return anything from anywhere, it has a fixed (albeit large) set of failure modes.  One mode I'd really like to see from Mimic in the future is a "fuzz mode" where it returns a series of realistic errors, and we won't be able to do that until we have a good selection of un-parameterized failures.

There are tickets for this - #270 for doing it for CLBs, #269 for doing it for Nova.

I wouldn't block this from going in on this basis (Nova already has one too-flexible failure API, it would be better to have *something* for CLBs).  But, as I understand it, the way that test case development is going to go is that you will identify some failure cases that need testing, and then go into Autoscale's test suite and add each expected failure from Nova or from CLBs into calls to mimic from Autoscale.

It would be good if we could, instead of doing that, each time an Autoscale test case needs a new error from CLBs or from Nova, instead of hitting a generalized "failure" endpoint for each new thing, make a Mimic PR for each new thing, so that they can be classified, named, and documented as a shared resource.  This way we can also have a separate artifact to show to the various teams (CLBs, Nova, etc) to say "this is the failure mode we think you've described to us, did we get it right", and something that others can check against the documentation.  If the parameters are squirreled away in an Otter test case it seems like it would be harder to get multiple teams to look at and update them.

lekhajee:
@kivattik : can we close this PR if this is not going to be worked upon.

glyph:
4 months with no response; @manishtomar can comment if it's still necessary for his testing.

manishtomar:
It will be nice to have this but I don't intend to work on this anytime. Will come back and reopen if/when required.

PR #319, opened 2015-06-15
fix(maas): update checks in place

ksheedlo:
This change fixes a bug where updating a check would overwrite the
existing check, causing partial updates to not work.

### DIFF
--- a/mimic/rest/maas_api.py
+++ b/mimic/rest/maas_api.py
@@ -477,23 +477,21 @@ class MaasMock(object):
                methods=['PUT'])
     def update_check(self, request, tenant_id, entity_id, check_id):
         """
-        Update an existing check
+        Updates a check in place.
         """
-        checks = self._entity_cache_for_tenant(tenant_id).checks_list
-        newcheck = json.loads(request.content.read())
-        newcheck['entity_id'] = entity_id
-        for k in newcheck.keys():
-            if 'encode' in dir(newcheck[k]):  # because there are integers sometimes.
-                newcheck[k] = newcheck[k].encode('utf-8')
-        for q in range(len(checks)):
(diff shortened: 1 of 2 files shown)


mimic/rest/maas_api.py
```diff
@@ -477,23 +477,21 @@ def get_check(self, request, tenant_id, entity_id, check_id):
                methods=['PUT'])
     def update_check(self, request, tenant_id, entity_id, check_id):
         """
-        Update an existing check
+        Updates a check in place.
         """
-        checks = self._entity_cache_for_tenant(tenant_id).checks_list
-        newcheck = json.loads(request.content.read())
-        newcheck['entity_id'] = entity_id
-        for k in newcheck.keys():
```
cyli: The re-encoding everything in utf-8 seems to have been eliminated in this change.  There wasn't a test for it before, but I'm not sure if this would break anything for anyone.  If you happen to know the exact use case and that it wouldn't break anyone using this plugin, that's cool, but otherwise maybe we should put it back?

If so, I will file an issue to add a test case for this.

cyli:
Also non-blocking - probably a check's update time should also be updated by this API, but we can file an issue for that.  :)

mimic/rest/maas_api.py
ksheedlo: Encoding things on the way in is kind of an anti-pattern - see #320. As much as possible, we should keep objects however we got them and encode on the way out to APIs that depend on bytes being encoded a certain way.

cyli: Ah ok, nm then, this LGTM.  :)

cyli:
Filed https://github.com/rackerlabs/mimic/issues/323 for future reference.

PR #320, opened 2015-06-15
refactor(maas): only encode at the boundary

ksheedlo:
This change updates the MaaS API to only encode strings before passing
them on to Twisted. This should make it easier to work with Unicode in
the MaaS API and help prevent double UTF-8 errors.

### DIFF
--- a/mimic/rest/maas_api.py
+++ b/mimic/rest/maas_api.py
@@ -115,6 +115,16 @@ class MCache(dict):
 def create_entity(params):
     """
     Returns a dictionary representing an entity
+
+    The Entity data model is described in the Rackspace Cloud Monitoring
+    Developer Guide, section 5.4:
+
+        http://docs.rackspace.com/cm/api/v1.0/cm-devguide/content/service-entities.html
+
+    This function returns an instance of the Entity model, mapped to a
+    Python dict. Keys of the returned object are strict ASCII. Where
+    values of the returned object are strings, or objects containing
+    strings, they must generally be considered to contain Unicode.
     """
     params = collections.defaultdict(lambda: '', params)
     newentity = {}
@@ -133,6 +143,16 @@ def create_entity(params):
 def create_check(params):
     """
     Returns a dictionary representing a check
+
+    The Check data model is described in the Rackspace Cloud Monitoring
+    Developer Guide, section 5.7:
+
+        http://docs.rackspace.com/cm/api/v1.0/cm-devguide/content/service-checks.html
+
+    This function returns an instance of the Check model, mapped to a
+    Python dict. Keys of the returned object are strict ASCII. Where
+    values of the returned object are strings, or objects containing
+    strings, they must generally be considered to contain Unicode.
     """
     params = collections.defaultdict(lambda: '', params)
     params['id'] = 'ch' + random_hex_generator(4)
@@ -156,6 +176,16 @@ def create_check(params):
 def create_alarm(params):
     """
     Returns a dictionary representing an alarm
+
+    The Alarm data model is described in the Rackspace Cloud Monitoring
+    Developer Guide, section 5.12:
+
+        http://docs.rackspace.com/cm/api/v1.0/cm-devguide/content/service-alarms.html
+
+    This function returns an instance of the Alarm model, mapped to a
+    Python dict. Keys of the returned object are strict ASCII. Where
+    values of the returned object are strings, or objects containing
+    strings, they must generally be considered to contain Unicode.
     """
     params = collections.defaultdict(lambda: '', params)
     params['id'] = 'al' + random_hex_generator(4)
@@ -171,6 +201,17 @@ def create_alarm(params):
 def create_notification_plan(params):
     """
     Creates a notification plan
+
+    The Notification Plan data model is described in the Rackspace Cloud
+    Monitoring Developer Guide, section 5.11:
+
+        http://docs.rackspace.com/cm/api/v1.0/cm-devguide/content/service-notification-plans.html
+
+
+    This function returns an instance of the Notification Plan model, mapped to
+    a Python dict. Keys of the returned object are strict ASCII. Where values
+    of the returned object are strings, or objects containing strings, they
+    must generally be considered to contain Unicode.
     """
     params['id'] = 'np' + random_hex_generator(4)
     params['critical_state'] = None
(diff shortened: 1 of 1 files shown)


ksheedlo:
/cc @glyph for Twisted expertise. I think this is the best way to do it, but would like a second opinion.

Also needs more manual testing to make sure this doesn't introduce any new Unicode bugs.

ksheedlo:
re: manual testing, it looks fine - no double encoding bugs, which is what I was mostly concerned about.

mimic/rest/maas_api.py
```diff
@@ -118,7 +118,7 @@ def create_entity(params):
     """
     params = collections.defaultdict(lambda: '', params)
     newentity = {}
-    newentity['label'] = params[u'label'].encode("utf-8")
+    newentity['label'] = params['label']
```
glyph: This points at a weakness in the `create_entity` docstring: what is it returning a dictionary of? What are the types of the keys and values?  (I think it's clear that the right answer is "unicode" but it should really say so.)

mimic/rest/maas_api.py
```diff
@@ -345,7 +330,7 @@ def list_entities(self, request, tenant_id):
         metadata['next_marker'] = None
         metadata['next_href'] = None
         request.setResponseCode(200)
-        return json.dumps({'metadata': metadata, 'values': entities})
+        return json.dumps({'metadata': metadata, 'values': entities}).encode('utf-8')
```
glyph: I am pretty sure that this is unnecessary, since `json.dumps` is already returning a `str` object on py2 (which is equivalent to a `bytes` object).  In fact, it's technically wrong, because you are encoding already-encoded data and run the risk of encoding errors if there's any non-ASCII.  On py3 it's a `str` but we're not worrying about that just yet :).

glyph:
@ksheedlo - I definitely prefer this approach, but for the time being I think the right solution is *just* to get rid of the encoding calls in MaaS and *not* to add any additional encoding logic in the request layer.  Moving forward to py3 there will need to be a more involved solution but hopefully by then we will have put all the json encode/decode stuff into a decorator and we can fix it in one place.

glyph:
I guess this is supposed to be a pure refactoring, with no change in behavior, which is why it adds no tests?

glyph:
Thanks again for your contribution! I think this will probably be good to land if we were to remove the (I hope) unnecessary encoding-the-response-on-the-way-out stuff.  That said, we could probably afford to be a little better about testing non-ASCII generally.

mimic/rest/maas_api.py
ksheedlo: The keys should be strict ASCII; generally values can be Unicode.

mimic/rest/maas_api.py
ksheedlo: Yeah, I wasn't sure if encoding the response on the way out was necessary or not. Let me try taking all those calls out and see if it still works.

ksheedlo: OTOH, I know encoding the headers on the way out is necessary.

ksheedlo:
@glyph Encoding the response body on the way out did turn out to be unnecessary. You're correct that this is a pure refactoring, which is why there are no new tests.

mimic/rest/maas_api.py
ksheedlo: Updated docstrings with cross references to the API docs and explanation of key/value types.

ksheedlo:
@glyph I've added additional documentation and fixed the change to not do any unnecessary encodings, so marked back as RFR.

mimic/rest/maas_api.py
glyph: "strict ASCII" isn't a Python data type though :).  Sorry, what I meant to say was "`unicode`".

In Python, there are 3 string types; two that you know about and one secret one.

The first is "bytes", which is `bytes` on python 2 ("`str`" is an alias) and `bytes` on python 3.  This represents binary data, which may of course represent text, but it implies a pile of octets that you haven't fully decided upon the meaning of yet.

The second is "text", which is `unicode` on python 2 and `str` on python 3 - sometimes referred to as `six.text_type` by users of the `six` library for writing code that works on 2 and 3 simultaneously.

The final, secret string type is "native string", which is `str` on python 2 and `str` on Python 3.  Certain interpreter internals, like identifiers and docstrings, are represented with `str` on python 2, but are unambiguously text as far as the user is concerned.  This is just terrible, but since they're almost always ASCII you can usually get along without suffering too much.

In this case, the important thing about the return value is that it's "`dict` mapping `unicode` to `unicode`".  The fact that the keys are ASCII-only `unicode` objects doesn't change the fact that they are still text keys.

Also, Mimic uses ReStructuredText for its docstrings, which means rather than "this function returns…" you can say something like this instead:

```rest
:return: an Entity model, which is described in `the Rackspace Cloud Monitoring
    Developer Guide, section 5.4
    <http://docs.rackspace.com/cm/api/v1.0/cm-devguide/content/service-entities.html>`_
:rtype: ``dict`` mapping ``unicode`` to ``unicode``, ``list``, or ``dict``.
```

ksheedlo: I was afraid this PR was going to spiral into me having to figure out ReStructuredText :scream: Do we have a doc generator that uses it, or some other way for me to know when I'm getting the syntax correct?

Re: Python string types, let's discuss this more in person at some point. I've done a good amount of Python in school and in previous jobs, but know comparatively little about the internals. I knew there was a difference between `'string'` and `u'string'`, but had to go to Stackoverflow to know that `u'string'` is a Unicode string and `'string'` is a byte string.

The thing that puzzles me is how indexing a dict with `'label'` or `u'label'` gives the same result. If they are two entirely different types of thing, shouldn't they key differently? The only reason I can think of is that it's a terrible UI for a programming language, but the way it is isn't consistent with my mental model of the internals.

ksheedlo:
Update: Redid the docstrings in reST format.

mimic/rest/maas_api.py
glyph: We are not currently generating docs from it, but we really should be.  If you want to file a ticket feel free to assign it to me :).

glyph: We should definitely talk more about string types in person; they are unfortunately very puzzling :-\.

glyph:
Cool, looking good. Thanks!

PR #314, opened 2015-06-12
fix(maas): update entities in place

ksheedlo:
This change makes MAAS update the entity in place instead of creating a
new one with the updated fields. This allows us to do partial updates.

Fixes #313.

PR #321, opened 2015-06-15
fix(maas): update alarms in place

ksheedlo:
Also adds the API to get an alarm by ID, which was curiously missing. I needed that API to make the test case work for the fix.

### DIFF
--- a/mimic/rest/maas_api.py
+++ b/mimic/rest/maas_api.py
@@ -558,25 +558,28 @@ class MaasMock(object):
                methods=['PUT'])
     def update_alarm(self, request, tenant_id, entity_id, alarm_id):
         """
-        update alarm
+        Updates an alarm in place.
+
+        Documentation for this API can be found in the Rackspace Cloud
+        Monitoring Developer Guide, section 5.12.5, "Update alarm by ID".
+        The full link is quite long, but you can reach it by browsing
+        to the following goo.gl URL:
+
+            http://goo.gl/NhxgTZ
         """
-        alarms = self._entity_cache_for_tenant(tenant_id).alarms_list
-        newalarm = json.loads(request.content.read())
-        newalarm['entity_id'] = entity_id
-        newalarm['updated_at'] = time.time()
-        for k in newalarm.keys():
-            if 'encode' in dir(newalarm[k]):  # because there are integers sometimes.
-                newalarm[k] = newalarm[k].encode('utf-8')
-        for q in range(len(alarms)):
-            if alarms[q]['entity_id'] == entity_id and alarms[q]['id'] == alarm_id:
-                newalarm['check_id'] = alarms[q]['check_id']
-                del alarms[q]
-                alarms.append(newalarm)
+        update = json.loads(request.content.read())
+        for alarm in self._entity_cache_for_tenant(tenant_id).alarms_list:
+            if alarm['entity_id'] == entity_id and alarm['id'] == alarm_id:
+                for k in ['check_id', 'notification_plan_id', 'criteria', 'disabled', 'label',
+                          'metadata']:
+                    if k in update:
+                        alarm[k] = update[k]
                 break
         myhostname_and_port = 'http://' + request.getRequestHostname() + ':' + self.endpoint_port
         request.setResponseCode(204)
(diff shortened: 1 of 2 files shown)


tokunbo:
As far as I know, partial updates already work. I just tested it. The line of code that is suppose to handle the partial updates for all 3 of those objects is the first line in the create functions <code>params = collections.defaultdict(lambda: '', params)</code> The new entity/check/alarm that's returned should have all the fields of the old one.

tokunbo:
Oh... nvm... I guess CI is sending something that it wasn't doing before.

ksheedlo:
@tokunbo Intelligence as it is currently sends the entire object. My new PR to Intelligence does partial updates, but only for Entities, and that change has already landed on master (see #314)

mimic/test/test_maas.py
```diff
@@ -270,6 +290,24 @@ def test_update_alarm(self):
         alarm2 = self.get_responsebody(resp)['values'][0]['alarms'][0]
         self.assertEquals(alarm['label'], alarm2['label'])
 
+    def test_partial_update_alarm(self):
+        """
+        Update an alarm, fields not specified in the body don't change.
```
glyph: Minor nitpick here - it's generally better to phrase test assertions positively.  So rather than "fields not … don't …", instead, "When a request is received that updates an alarm, fields not specified in that request's body remain the same."

mimic/rest/maas_api.py
```diff
@@ -533,28 +533,46 @@ def create_alarm(self, request, tenant_id, entity_id):
         return ''
 
     @app.route('/v1.0/<string:tenant_id>/entities/<string:entity_id>/alarms/<string:alarm_id>',
+               methods=['GET'])
+    def get_alarm(self, request, tenant_id, entity_id, alarm_id):
+        """
+        Gets an alarm by ID.
```
glyph: Minor nitpick: docstrings should be phrased in the imperative rather than the declarative mood; partially because "implicit you" (make me a sandwich) in a command is more idiomatic than "implicit it" (makes me a sandwich).  However, probably don't change this right now because declarative seems to be the convention within this file and local conventions always win over general style – until they change, anyway :).

glyph: Ooh, and I missed a great opportunity to make an argument from authority - this is actually part of [PEP 257](https://www.python.org/dev/peps/pep-0257/), apparently:

> The docstring is a phrase ending in a period. It prescribes the function or method's effect as a command ("Do this", "Return that"), not as a description; e.g. don't write "Returns the pathname ...".

mimic/test/test_maas.py
```diff
@@ -228,6 +228,26 @@ def test_partial_update_entity(self):
         self.assertEquals('ag13378901234', data['agent_id'])
         self.assertEquals('ItsAnEntity', data['label'])
 
+    def test_get_alarm(self):
+        """
+        test getting an alarm by ID
```
glyph: Test assertions should be described as a description of the behavior involved, and not as an explanation of what the test is doing.  For example, if you have to use the words "test that", "verify that", or "should", it's a hint you're writing it incorrectly.  https://github.com/testing-cabal/testdoc sort of explains this although I am pretty sure @jml actually wrote something longer about it at some point and I would love it if he magically showed up to link to it...

In any case, here, the appropriate translation would be something like "The URL `/entities/<entity_id>/alarms/<alarm_id>` returns a JSON description of the alarm".

glyph:
@ksheedlo - modulo the very minor grammar and spelling nitpicks, there's only one thing that I think really *needs* to change here, and that is a link in the docstring of the update route that points at the documentation for MaaS's API detailing how partial updates are supposed to work.  Generally, we really need to get in the habit, in Mimic, of always linking to reference docs where they exist, since we have such a small team maintaining such a broad diversity of different APIs.

Thanks for the contribution!

mimic/test/test_maas.py
ksheedlo: +1, will fix.

mimic/rest/maas_api.py
ksheedlo: I'll remember that for next time :)

jml:
[How to write test docstrings](https://plus.google.com/+JonathanLange/posts/YA3ThKWhSAj)

PR #322, opened 2015-06-16
Fix a CLB error message when adding nodes

cyli:
The status should be single-quoted.  As part of #270, attempting to put these errors in a new module.

### DIFF
--- a/mimic/model/clb_objects.py
+++ b/mimic/model/clb_objects.py
@@ -2,16 +2,19 @@
 Model objects for the CLB mimic.
 """
 
-from mimic.util.helper import (not_found_response, seconds_to_timestamp,
-                               EMPTY_RESPONSE,
-                               invalid_resource)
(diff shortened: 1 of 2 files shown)


mimic/model/clb_objects.py
```diff
@@ -2,16 +2,19 @@
 Model objects for the CLB mimic.
 """
 
-from mimic.util.helper import (not_found_response, seconds_to_timestamp,
-                               EMPTY_RESPONSE,
-                               invalid_resource)
-from twisted.python import log
 from characteristic import attributes, Attribute
+
+from twisted.python import log
```
sam-falvo: Are my eyes deceiving me, or is this import not used in this patch?

sam-falvo:
LGTM

cyli:
Thanks @sam-falvo!!

Issue #265, opened 2015-05-12
Support changing the attributes of a load balancer node.


cyli:
http://docs.rackspace.com/loadbalancers/api/v1.0/clb-devguide/content/PUT_updateNode_v1.0__account__loadbalancers__loadBalancerId__nodes__nodeId__Nodes-d1e2173.html

PR #325, opened 2015-06-16
Refactor CLB nodes into an object, not a dict.

cyli:
As a precursor to #265.


glyph:
Thanks for refactoring this - death to dictionaries :)

Issue #202, opened 2015-03-18
stop setting Request.defaultContentType


glyph:
I think there was some confusion around this being specific to an individual file, rather than a temporary (and gross) hack to make a specific change.  We should be careful to set the content type in each response.  If copy/pasting the lines of code to do `json.dumps` and `setHeader` are tedious, we should make a decorator that combines it with `route`, perhaps, to avoid duplication, but we shouldn't manipulate global state.

cyli:
Related to #40.  Please close out both if the decorator is created.

PR #331, opened 2015-06-18
Cloud feeds control plane and basic model behind listing product endpoints.

sam-falvo:
I probably should have split this PR into two pieces.

The first aspect of the PR is that it creates the boilerplate needed to create the API for data- and control-plane access.

The second aspect of the PR is that it creates the basic model behind the plugin as a whole, and as a first step, modeling the product endpoint listing.

It should be relatively simple to just do json.dump(...) on the basic model to render the JSON representation for a client, but I don't do that here yet.


mimic/model/cloudfeeds.py
```diff
@@ -0,0 +1,86 @@
...
+import attr
+
+
+@attr.s
+class CloudFeedsProduct(object):
+    """
+    Models a single CloudFeed product endpoint and its respective
+    functionality.
+    """
+    title = attr.ib(attr.validators.instance_of(str))
```
cyli: This may not be as compatible with Python3 - we've been using `six.string_types` for our validation so it accepts both bytes and unicode:

```
from six import string_types
...
title = attr.ib(attr.validators.instance_of(string_types))
```

mimic/rest/cloudfeeds.py
```diff
@@ -0,0 +1,176 @@
...
+        """
+        Returns a list of cloud feeds entries.  Note that these are not
+        cloud feeds product endpoints; this is one step removed from those.
+        You'll need to GET from one of these URLs to see the catalog of
+        product endpoints.
+        """
+        return [
+            Entry(tenant_id, "rax:feeds", "cloudFeeds",
+                  [
+                      Endpoint(tenant_id, region, text_type(uuid4()), prefix="noversion")
```
cyli: Non-blocking - just as a note, leaving out the prefix entirely means that it won't be present at all in the URL returned by identity.

mimic/model/cloudfeeds.py
```diff
@@ -0,0 +1,86 @@
+"""
+Emulate the behavior of Rackspace Cloud Feeds service.
```
cyli: Non-blocking - could we include a link to the API docs (I assume http://docs.rackspace.com/cloud-feeds/api/v1.0/feeds-devguide/content/overview.html?)

mimic/model/cloudfeeds.py
```diff
@@ -0,0 +1,86 @@
...
+@attr.s
+class CloudFeeds(object):
+    """
+    Models CloudFeeds support at the plugin-level.
+    """
+    _endpoints = attr.ib(default=attr.Factory(dict))
+
+    def get_product_endpoints(self):
+        """
+        Return a list of product endpoints registered with this class.
```
cyli: Docstring says list, but it actually returns a copy of the existing endpoints dictionary - it looks like the tests call `render_product_endpoints_dict`  with the result of this value, so is it supposed to be a list?

mimic/model/cloudfeeds.py
```diff
@@ -0,0 +1,86 @@
...
+        """
+        return dict(self._endpoints)
+
+    def register_product(self, title, href):
+        """
+        Register a product by creating a URL that one can GET and/or POST from/to,
+        and providing a descriptive title for it in the product endpoint
+        listing.
```
cyli: Non-blocking doc nitpick: since there is an explicit test to ensure that this just does nothing if you attempt to re-register with either the same or a different title, maybe we should document that?

cyli:
Love the tests!  This looks really clean so far - coveralls is going to be unhappy if there aren't tests for the endpoints themselves, though, so as you say it might be worthwhile to either split this up or just add in some basic tests for regular endpoint and the control endpoint.  :)

mimic/rest/cloudfeeds.py
```diff
@@ -0,0 +1,132 @@
...
+    def __init__(self, api_mock, uri_prefix, session_store, region_name):
+        """
+        Fetches the cloud feeds id for a failure, invalid scenarios, etc.
+        """
+        self.uri_prefix = uri_prefix
+        self.region_name = region_name
+        self._api_mock = api_mock
+        self._session_store = session_store
+
+    def session(self, tenant_id):
```
cyli: This function is still uncovered, because nothing does anything with the session (there are no tests for getting or storing saved data yet).  Can this be removed for this PR?

mimic/model/cloudfeeds.py
sam-falvo: Fixed in an earlier commit.

mimic/rest/cloudfeeds.py
```diff
@@ -0,0 +1,116 @@
...
+from twisted.plugin import IPlugin
+from mimic.catalog import Endpoint, Entry
+from mimic.imimic import IAPIMock
+from mimic.rest.mimicapp import MimicApp
+
+
+from characteristic import attributes
+
+
+Request.defaultContentType = 'application/json'
```
lekhajee: This is setting the global state and should not be set. Refer #202

mimic/rest/cloudfeeds.py
```diff
@@ -0,0 +1,116 @@
...
+    """
+    def catalog_entries(self, tenant_id):
+        """
+        Cloud feeds controller endpoints.
+        """
+        return [
+            Entry(
+                tenant_id, "rax:feeds", "cloudFeedsControl",
+                [
+                    Endpoint(tenant_id, region, text_type(uuid4()), prefix="noversion")
```
lekhajee: Any reason why this is has the `prefix` set to `noversion` while the `CloudFeedsApi` has no version set. Also, looks like the Cloud Feeds API has no version, so how about not setting `prefix` in both `CloudFeedsApi` and `CloudFeedsControlApi`

sam-falvo: Oversight.  Fixed in db0fa7c.

On Tue, Jun 23, 2015 at 11:58 AM, Lekha Jeevan <notifications@github.com>
wrote:

> In mimic/rest/cloudfeeds.py
> <https://github.com/rackerlabs/mimic/pull/331#discussion_r33078222>:
>
> > +@implementer(IAPIMock, IPlugin)
> > +@attributes(["cf_api"])
> > +class CloudFeedsControlApi(object):
> > +    """
> > +    This class registers the load balancer controller API in the service
> > +    catalog.
> > +    """
> > +    def catalog_entries(self, tenant_id):
> > +        """
> > +        Cloud feeds controller endpoints.
> > +        """
> > +        return [
> > +            Entry(
> > +                tenant_id, "rax:feeds", "cloudFeedsControl",
> > +                [
> > +                    Endpoint(tenant_id, region, text_type(uuid4()), prefix="noversion")
>
> Any reason why this is has the prefix set to noversion while the
> CloudFeedsApi has no version set. Also, looks like the Cloud Feeds API
> has no version, so how about not setting prefix in both CloudFeedsApi and
> CloudFeedsControlApi
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/rackerlabs/mimic/pull/331/files#r33078222>.
>



-- 
Samuel A. Falvo II

mimic/rest/cloudfeeds.py
sam-falvo: Fixed in db0fa7c

On Tue, Jun 23, 2015 at 11:50 AM, Lekha Jeevan <notifications@github.com>
wrote:

> In mimic/rest/cloudfeeds.py
> <https://github.com/rackerlabs/mimic/pull/331#discussion_r33077260>:
>
> > +Defines the control plane API endpoints for the Cloudfeeds Plugin.
> > +"""
> > +from uuid import uuid4
> > +from six import text_type
> > +from zope.interface import implementer
> > +from twisted.web.server import Request
> > +from twisted.plugin import IPlugin
> > +from mimic.catalog import Endpoint, Entry
> > +from mimic.imimic import IAPIMock
> > +from mimic.rest.mimicapp import MimicApp
> > +
> > +
> > +from characteristic import attributes
> > +
> > +
> > +Request.defaultContentType = 'application/json'
>
> This is setting the global state and should not be set. Refer #202
> <https://github.com/rackerlabs/mimic/issues/202>
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/rackerlabs/mimic/pull/331/files#r33077260>.
>



-- 
Samuel A. Falvo II

lekhajee:
+1 lgtm

PR #334, opened 2015-06-19
feat(maas): paginate /views/overview

ksheedlo:
This change implements pagination for the /views/overview call in maas,
which is important for testing Intelligence against a simulated large
account. General information about pagination in the maas API can be
found here:

http://docs.rackspace.com/cm/api/v1.0/cm-devguide/content/api-paginated-collections.html

### DIFF
--- a/mimic/rest/maas_api.py
+++ b/mimic/rest/maas_api.py
@@ -30,6 +30,16 @@ from mimic.util.helper import random_hex_generator
 Request.defaultContentType = 'application/json'
 
 
+def _find_index_of_id(xs, val):
+    """
+    Finds the index of the first element in the list with the specified id.
+    """
+    for i, x in enumerate(xs):
+        if x['id'] == val:
+            return i
+    raise ValueError('No element with id {0} in list'.format(val))
+
+
 @implementer(IAPIMock, IPlugin)
 class MaasApi(object):
 
@@ -661,15 +671,30 @@ class MaasMock(object):
         """
         serves the overview api call,returns all entities,checks and alarms
         """
-        entities = self._entity_cache_for_tenant(tenant_id).entities_list
+        all_entities = self._entity_cache_for_tenant(tenant_id).entities_list
         checks = self._entity_cache_for_tenant(tenant_id).checks_list
         alarms = self._entity_cache_for_tenant(tenant_id).alarms_list
-        metadata = {}
-        metadata['count'] = len(entities)
-        metadata['marker'] = None
-        metadata['next_marker'] = None
-        metadata['limit'] = 1000
-        metadata['next_href'] = None
+        page_limit = min(int(request.args.get('limit', [100])[0]), 1000)
+        offset = 0
+        current_marker = request.args.get('marker', [None])[0]
+        if current_marker is not None:
+            try:
+                offset = _find_index_of_id(all_entities, current_marker)
+            except ValueError:
+                offset = 0
+
+        entities = all_entities[offset:offset + page_limit]
+        next_marker = None
+        if offset + page_limit < len(all_entities):
+            next_marker = all_entities[offset + page_limit]['id']
+
+        metadata = {
+            'count': len(entities),
+            'marker': current_marker,
+            'next_marker': next_marker,
+            'limit': page_limit,
+            'next_href': None
+        }
         values = []
         for e in entities:
             v = {}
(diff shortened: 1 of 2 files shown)


mimic/rest/maas_api.py
```diff
@@ -30,6 +30,16 @@
 Request.defaultContentType = 'application/json'
 
 
+def _find_index_of_id(xs, val):
```
glyph: There is a built-in method on `list`, "index", which I think you want to use instead.  It raises an exception (`IndexError`) rather than returning a sentinel (`-1`), but since you're explicitly testing against `-1` anyway (without documenting it as a constant, I might add ;-)) catching the exception might be better.

mimic/rest/maas_api.py
```diff
@@ -642,14 +652,26 @@ def overview(self, request, tenant_id):
...
+        next_marker = None
+        if offset + page_limit < len(all_entities):
+            next_marker = all_entities[offset + page_limit]['id']
+
         metadata = {}
         metadata['count'] = len(entities)
-        metadata['marker'] = None
-        metadata['next_marker'] = None
-        metadata['limit'] = 1000
+        metadata['marker'] = current_marker
```
glyph: Just as a stylistic nitpick; this isn't a violation, but I prefer to do this:
```python
metadata = {
    'marker': current_marker,
    'next_marker': next_marker,
    # ...
}
```
since that puts the whole variable into an expression rather than splitting it into a bunch of statements and leaving the variable in a bunch of half-created states as it's being built up.

glyph:
Thanks for the PR @ksheedlo - can you just remove the `index` method and resubmit?

mimic/rest/maas_api.py
ksheedlo: `list#index` compares elements by value, where this function only looks at the value of the `id` field. We could raise `IndexError` here instead of returning `-1`, which I like because `-1` is usually a valid index into a python list.

ksheedlo:
@glyph see diff comment on why `_find_index_of_id` is necessary.

mimic/rest/maas_api.py
ksheedlo: Tested, and actually it's `ValueError` - I'm going to fix my code to return that.

glyph: (TMYK: Python nomenclature would have it as `list.index`.)

ksheedlo: Noted. I'm used to JSDoc where `list#index` is a method and `list.index` is a static class method, in case you're wondering where that's from.

glyph: OK, the difference is clear to me now.  Sorry I didn't catch that before.  I think I think this is probably fine, but let me propose an alternative.

`index` is doing an `==` comparison, which means that it's calling a method, `__eq__`, on each list item, with what it was passed.  Given Python's rules for equality comparison, this will actually call `__eq__` on the *left*-hand-side argument if the right-hand side doesn't recognize its type.

So, you can implement this a different way, supplying your own comparison data structure but using the built-in implementation of `index`.  Consider this example:

```python
class MatchesID(object):
    def __init__(self, id):
        self.id = id

    def __eq__(self, other):
        return other['id'] == self.id


values = [
    {"id": 3, "other": 4},
    {"id": 4, "other": 81},
    {"id": 5, "other": 24},
    {"id": 6, "other": 4},
    {"id": 3, "other": 9},
    {"id": 3, "other": 12},
    {"id": 6, "other": 14},
]

idx = (values.index(MatchesID(5)))
print(values[idx])
```

this will print out

```python
{'other': 24, 'id': 5}
```

ksheedlo: Interesting. That's probably a more pythonic way to express what I'm trying to do :snake:

glyph: Gotcha. I think Ruby uses a similar method separator? I've definitely seen `#` used that way before, I just thought you'd like to know :).

glyph: Anyway this is *totally* a quibble about an internal implementation detail now so I think I'll merge, you can refactor this later if you like my suggestion.

ksheedlo: go for it, I'll send a follow up PR.

glyph:
Every reason to believe the tests will pass but we probably want to dot all the appropriate I's and cross the relevant T's before merging.

glyph:
Thanks again ken!

glyph:
(Just to be clear, with that tag, go ahead and merge yourself once tests have passed if I don't get to the button first.)

ksheedlo:
+1 sounds good

PR #335, opened 2015-06-20
Fix some CLB  bugs

cyli:
These were introduced by the node model change:

1.  When GET-ing a load balancer, if there are no nodes, the "nodes" attribute does not appear.
1.  When deleting a node, previously it deleted the "nodes" attribute from the whole LB.  We want to leave it and just filter it out when returning responses.

### DIFF
--- a/mimic/test/test_loadbalancer.py
+++ b/mimic/test/test_loadbalancer.py
@@ -327,7 +327,7 @@ class LoadbalancerAPITests(SynchronousTestCase):
                               "port": 80, "condition": "ENABLED"}
                              for i in range(1, 4)])
         list_resp, list_body = self.successResultOf(json_request(
-            self, self.root,  "GET", self.uri + '/loadbalancers'))
+            self, self.root, "GET", self.uri + '/loadbalancers'))
         self.assertEqual(list_resp.code, 200)
         self.assertEqual(len(list_body['loadBalancers']), 2)
 
--- a/mimic/test/test_loadbalancer.py
+++ b/mimic/test/test_loadbalancer.py
@@ -113,7 +113,18 @@ class LoadbalancerAPITests(SynchronousTestCase):
 
     def _create_loadbalancer(self, name=None, api_helper=None, nodes=None):
         """
-        Helper method to create a load balancer and return the lb_id
+        Helper method to create a load balancer and return the lb_id.
+
+        :param str name: The name fo the load balancer, defaults to 'test_lb'
+        :param api_helper: An instance of :class:`APIMockHelper` - defaults to
+            the one created by setup, but if different regions need to be
+            created, for instance, your test may make a different helper, and
+            so that helper can be passed here.
+        :param list nodes: A list of nodes to create the load balancer with -
+            defaults to creating a load balancer with no nodes.
+
+        :return: Load balancer ID
+        :rtype: int
         """
         api_helper = api_helper or self.helper
         lb_body = {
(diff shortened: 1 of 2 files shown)


cyli:
Looks like just the tests are left after merging #336

mimic/test/test_loadbalancer.py
```diff
@@ -336,10 +360,28 @@ def test_delete_loadbalancer(self):
...
+        resp, body = self.successResultOf(json_request(
+            self, self.root, "GET", self.uri + '/loadbalancers/' + str(lb_id)))
+        self.assertEqual(resp.code, 200)
+        self.assertEqual(body['loadBalancer']['id'], lb_id)
+        self.assertNotIn('nodeCount', body['loadBalancer'])
+        self.assertEqual(len(body['loadBalancer']['nodes']), 1)
+
+    def test_get_loadbalancer_with_nodes(self):
+        """
+        Test to verify :func:`get_load_balancers` with on
```
glyph: OK, so, this was the way the docstring started already, but you repeated it.

First off (hi @jml!) we should phrase test docstrings like this.  https://plus.google.com/+JonathanLange/posts/YA3ThKWhSAj 

Second, what is even a grammar? :)  "Test to verify get load balancers with on GET (path)"

mimic/test/test_loadbalancer.py
```diff
@@ -313,6 +315,28 @@ def test_list_loadbalancers(self):
         self.assertTrue(list_lb_response_body['loadBalancers'][0]['id'] !=
                         list_lb_response_body['loadBalancers'][1]['id'])
 
+    def test_list_loadbalancers_have_no_nodes(self):
+        """
+        Test to verify that nodes does not appear on the load balancers when
```
glyph: https://plus.google.com/+JonathanLange/posts/YA3ThKWhSAj

mimic/test/test_loadbalancer.py
```diff
@@ -111,25 +111,26 @@ def setUp(self):
         self.root = self.helper.root
         self.uri = self.helper.uri
 
-    def _create_loadbalancer(self, name=None, api_helper=None):
+    def _create_loadbalancer(self, name=None, api_helper=None, nodes=None):
```
glyph: We have too many methods in mimic without documentation for their parameters.  Since you're changing the signature on this one, can I make you add `:param:` declarations for the parameters to say what they are, what type they should be, and importantly why they're all optional and what the behavior is if they're omitted?  (Thanks.)

mimic/test/test_loadbalancer.py
```diff
@@ -233,23 +234,25 @@ def test_multiple_regions_multiple_endpoints(self):
 
     def test_add_load_balancer(self):
         """
-        Test to verify :func:`add_load_balancer` on ``POST /v1.0/<tenant_id>/loadbalancers``
+        Test to verify :func:`add_load_balancer` on
```
glyph: How about something more like:

```rest
When :func:`add_load_balancer` is run (by ``POST``ing to
``/v1.0/<tenant_id>/loadbalancers`` without any nodes, no node information
appears in the response).
```

mimic/test/test_loadbalancer.py
```diff
@@ -313,6 +315,28 @@ def test_list_loadbalancers(self):
         self.assertTrue(list_lb_response_body['loadBalancers'][0]['id'] !=
                         list_lb_response_body['loadBalancers'][1]['id'])
 
+    def test_list_loadbalancers_have_no_nodes(self):
+        """
+        When listing load balancers, nodes do not appear even if the load
```
glyph: +1 this test docstring looks like a great example, thanks :)

mimic/test/test_loadbalancer.py
cyli: Added in https://github.com/rackerlabs/mimic/commit/f07ca2f6883802b19655ca360ba4b106972f869f

glyph:
Aside from docstrings, I think this looks pretty good.

Issue #330, opened 2015-06-17
Convert the CLB storage from a dictionary to an object


cyli:
Particularly so it can handle checking it's own active state before doing anything to itself, rather than have every method on the regional collection that manipulates nodes first check the state of the load balancer and generate an immutable error.

glyph:
So by "CLB storage" you mean one level up from what was done in #325?

cyli:
Sorry, yes - #325 converted CLB nodes to objects.  I meant convert a CLB itself (which has many nodes) from a dictionary to an object.

PR #336, opened 2015-06-20
Move a little in the direction of having a real CLB object

cyli:
(So moving towards #330)

Part of #335 got confusing because some attributes may be the dict or may not, and some views should show them and some should not...

So this is a little bit of a refactor that presents both the short and full view of a CLB.

These do pass the tests from #335, but there is a bit of a merge conflict between the two.

### DIFF
--- a/mimic/model/clb_objects.py
+++ b/mimic/model/clb_objects.py
@@ -4,6 +4,7 @@ Balancer API docs
 <http://docs.rackspace.com/loadbalancers/api/v1.0/clb-devguide/content/API_Operations.html>`
  for more information.
 """
+from copy import deepcopy
 from random import randrange
 
 import attr
@@ -16,7 +17,6 @@ from twisted.python import log
 
 from mimic.canned_responses.loadbalancer import (load_balancer_example,
                                                  _verify_and_update_lb_state,
-                                                 _lb_without_tenant,
                                                  _delete_node)
 from mimic.model.clb_errors import (
     considered_immutable_error,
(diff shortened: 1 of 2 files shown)


mimic/model/clb_objects.py
```diff
@@ -402,10 +432,9 @@ def add_node(self, node_list, lb_id, current_timestamp):
                         return (resource, 413)
 
             # If there were no duplicates
-            new_nodeCount = self.lbs[lb_id]["nodeCount"] + len(nodes)
+            new_nodeCount = len(self.lbs[lb_id].nodes) + len(nodes)
             if new_nodeCount <= self.node_limit:
-                self.lbs[lb_id]["nodes"] = self.lbs[lb_id]["nodes"] + nodes
-                self.lbs[lb_id]["nodeCount"] = new_nodeCount
+                self.lbs[lb_id].nodes = self.lbs[lb_id].nodes + nodes
```
glyph: Just noticing this could use `+=` to be a little clearer, since `self.lbs[lb_id].nodes` is a complex enough expression I had to read it twice to make sure it was the same :).

mimic/model/clb_objects.py
```diff
@@ -183,20 +225,15 @@ def add_load_balancer(self, tenant_id, lb_info, lb_id, current_timestamp):
         if "lb_building" in self.meta[lb_id]:
             status = "BUILD"
 
-        # Add tenant_id and nodeCount to self.lbs
+        # Add tenant_id to self.lbs
         current_timestring = seconds_to_timestamp(current_timestamp)
-        self.lbs[lb_id] = load_balancer_example(lb_info, lb_id, status,
-                                                current_timestring)
+        self.lbs[lb_id] = CLB(load_balancer_example(lb_info, lb_id, status,
```
glyph: Clean-up idea, possibly for the future: this method (as several in this module) does `self.lbs[lb_id]` a whole bunch of times; it might make sense to extract that into a local variable.

glyph:
Seems pretty straightforward, so let's merge so you can get working on resolving the #335 conflict.

PR #337, opened 2015-06-22
Customer API plugin 

lekhajee:


mimic/model/customer_objects.py
```diff
@@ -0,0 +1,84 @@
...
+    A collection of contact objects for a tenant.
+    """
+
+    def add_to_contacts_store(self, tenant_id, contact_list):
+        """
+        Create a new Contact object for each contact in `contact_list`
+        and append it to :obj: `ContactsStore` of the tenant
+        """
+        self.contacts_store[tenant_id] = [
+            Contact(each_contact) for each_contact in contact_list]
```
sam-falvo: This looks like a problem here.  You're passing a single parameter, but the use of characteristic in the above class indicates you need at least three to construct the object.

mimic/model/customer_objects.py
```diff
@@ -0,0 +1,84 @@
...
+        },
+        "link": [
+            {
+                "rel": "via",
+                "href": "http://link-to-nothing-for-contact-in-customer-api"
+            }
+        ]
+    }
+
+    def generate_contacts(self, tenant_id):
```
sam-falvo: You use this as a class method in the code below, but it's not annotated to be a class method here.  This will cause run-time errors if not addressed.

lekhajee:
@sam-falvo: Sorry, I dint mean for any of this to be reviewed. All that code is incomplete, I just opened a branch so I could reference the PR in the corresponding Ele PR so that the others could track progress.

mimic/model/customer_objects.py
```diff
@@ -0,0 +1,113 @@
+"""
+Customet Contacts storage object
```
ksheedlo: nit: `Customer`

mimic/model/customer_objects.py
```diff
@@ -0,0 +1,113 @@
...
+    """
+    A :obj:`Contact` is a representation for each contact for a tenant.
+    """
+
+    static_defaults = {
+        "firstName": "Pat",
+        "lastName": "Example",
+        "customerAccountType": "CLOUD",
+        "contactNumber": "RPN-994-758-811",
+        "rcn": "RCN-348-367-072",
```
ksheedlo: Where did these numbers come from? Should they be more obviously fake?

mimic/model/customer_objects.py
```diff
@@ -0,0 +1,113 @@
...
+                        }
+            ]
+        },
+        "roles": {
+            "role": []
+        },
+        "link": [
+            {
+                "rel": "via",
+                "href": "http://link-to-nothing-for-contact-in-customer-api"
```
ksheedlo: maybe `customer.example.com` would be better.

mimic/model/customer_objects.py
```diff
@@ -0,0 +1,113 @@
...
+                                    ('example2@example.com', 'TECHNICAL')]
+            self.add_to_contacts_store(tenant_id, default_contact_list)
+
+        contacts = [each.generate_contacts()
+                    for each in self.contacts_store[tenant_id]]
+        return {
+            "link": [
+                {
+                    "rel": "next",
+                    "href": "http://link-to-nothing-in-the-customer-api"
```
ksheedlo: likewise, `customer.example.com`

mimic/rest/customer_api.py
```diff
@@ -0,0 +1,48 @@
...
+        response = self.core.contacts_store.list_contacts_for_tenant(tenant_id)
+        return json.dumps(response)
+
+    @app.route('/<string:tenant_id>/contacts', methods=['POST'])
+    def add_customer_contacts_for_tenant(self, request, tenant_id):
+        """
+        Adds new contacts to a tenant and responds with a 200.
+
+        Note: If there is a GET on the tenant before this `POST` call, the default
+        contacts would have been listed. This POST will overwrite the existing conatacts
```
ksheedlo: nit: spelling *overwrite the existing contacts*

mimic/rest/customer_api.py
```diff
@@ -0,0 +1,48 @@
...
+        Until the contacts are set by the `POST` call, this returns the default
+        contacts for a tenant.
+        """
+        response = self.core.contacts_store.list_contacts_for_tenant(tenant_id)
+        return json.dumps(response)
+
+    @app.route('/<string:tenant_id>/contacts', methods=['POST'])
+    def add_customer_contacts_for_tenant(self, request, tenant_id):
+        """
+        Adds new contacts to a tenant and responds with a 200.
```
ksheedlo: According to internal docs, this API returns a 201 upon success.

mimic/test/test_customer.py
```diff
@@ -0,0 +1,116 @@
...
+        The ``GET /contacts`` call returns a list of default contacts
+        for a tenant with the response code 200.
+        """
+        response = self.get_contacts(self.root, "111111")
+        self.validate_contact_list(response, self.default_contacts)
+
+    def test_get_default_contacts_for_multiple_tenants_successfully(self):
+        """
+        The ``GET /contacts`` call returns a list of default contacts
+        for multiple tenants with the response code 200.
```
ksheedlo: Why is this different than getting default contacts for a single tenant?

mimic/test/test_customer.py
```diff
@@ -0,0 +1,116 @@
...
+        self.validate_contact_list(response, self.default_contacts)
+        expected_contacts = [{"email": "test1@email.com", "role": "PRIMARY"},
+                             {"email": "new1@email.com", "role": "TECHNICAL"}]
+        response = self.successResultOf(request(
+            self, self.root, "POST",
+            "/v1/customer_accounts/CLOUD/555555/contacts",
+            json.dumps(expected_contacts)))
+        self.assertEqual(200, response.code)
+        response = self.get_contacts(self.root, "555555")
+        self.validate_contact_list(response, expected_contacts)
```
ksheedlo: otherwise tests looks good :sparkles:

mimic/rest/customer_api.py
lekhajee: Yea, I think the api itself was returning 200. And this was working against ele tests.

mimic/test/test_customer.py
lekhajee: This has been a while, and I don't remember I thought to do this. But I am guessing I wanted to make sure the defaults was set on all tenants and not just the first get contact call.

ksheedlo:
LGTM :+1:

Merge it if we need it. Are @sam-falvo and/or @ynachiket going to pick up the work that's necessary in MaaS to use this? I'd hate to see all this code not get used, not to mention work going to waste but this is a whole new plugin we'll have to support.

lekhajee:
@ksheedlo: yes, that is the hope; some of the maas team folks have mentioned picking up that work.

PR #338, opened 2015-06-23
More CLB object model refactors

cyli:
1. Since the collection is per tenant, remove references to tenant ids from the method calls
2. Since the object can store a clock, no longer require the timestamp to be passed when calling methods on the CLB objects.


glyph:
Nice to see these cleanups.  Thanks!

PR #339, opened 2015-06-24
style(maas): match ID based on custom __eq__

ksheedlo:
Instead of having a private function to match the first element of a
list by the `id` field, we can use `list.index` with a class
implementing custom equality.


glyph:
Thanks for following up on this one.

PR #340, opened 2015-06-27
Fix impersonation bug

lekhajee:
Allows for impersonation without the optional `expire-in-seconds` field


ksheedlo:
LGTM

lekhajee:
thanks @ksheedlo

PR #341, opened 2015-06-27
Add GET /users endpoint

lekhajee:


mimic/rest/auth_api.py
```diff
@@ -24,7 +25,10 @@
     TokenCredentials)
 from mimic.rest.mimicapp import MimicApp
 from mimic.session import NonMatchingTenantError
-from mimic.util.helper import invalid_resource
+from mimic.util.helper import (
+    invalid_resource,
+    random_hex_generator,
```
ksheedlo: `random_hex_generator` doesn't appear to be used in this file.

mimic/rest/auth_api.py
```diff
@@ -224,6 +228,27 @@ def get_username(self, request, tenant_id):
...
+        return json.dumps(dict(user={
+            "RAX-AUTH:domainId": session.tenant_id,
+            "id": session.user_id,
+            "enabled": True,
+            "username": session.username,
+            "email": "thisisrandom@email.com",
+            "RAX-AUTH:defaultRegion": "ORD",
+            "created": seconds_to_timestamp(time.time()),
+            "updated": seconds_to_timestamp(time.time())
+        }))
```
ksheedlo: This endpoint returns `403 Unauthorized` if the username does not match the auth token. Should we do the same in Mimic?

ksheedlo: Or, what happens if there is no session? Is one created?

lekhajee: The current implementation creates a new session like most auth endpoints in Mimic. The test `test_response_for_list_users` verifies that a new session is created.

ksheedlo: I'm concerned this isn't semantic (GET isn't supposed to have side effects), but would like to know more about your use case.

lekhajee: I understand, but as Mimic has no preset set of users as such we decided that until a use case requires otherwise, a GET call will fetch an existing session or will create session in the absence of an existing session.

My usecase in an ele test in the branch https://github.com/racker/ele/pull/2913 in `test-customer-client.js` test in which the GET on the `/users` occurs after have gotten the user from the customer api.

ksheedlo: OK, sounds good then.

ksheedlo:
:+1:

PR #347, opened 2015-07-29
Nova resize action

tawalton:


mimic/model/nova_objects.py
```diff
@@ -798,6 +798,21 @@ def request_delete(self, http_delete_request, server_id):
...
+        server = self.server_by_id(server_id)
+        if server is None:
+            http_action_request.setResponseCode(500)
+            return b''
+        if 'resize' in action_json:
+            flavor = action_json['resize']['flavorRef']
+            server.flavor_ref = flavor
+            http_action_request.setResponseCode(202)
+            return b''
+
```
tawalton: Only General Purpose and Standard flavor servers can be resized and they can only be resized to a larger flavor.  The API returns a 400 when a request is made to resize a server that is not General Purpose or Standard.  The API returns a 400 when a request is made to resize a server to another flavor class (attempting to resize from a General Purpose flavor class to a Standard flavor class)

mimic/model/nova_objects.py
```diff
@@ -798,6 +798,21 @@ def request_delete(self, http_delete_request, server_id):
...
         return b''
 
+    def action(self, http_action_request, server_id):
+        """
+        Perform the requested action on the provided server
+        """
+        action_json = loads(http_action_request.content.read())
+        server = self.server_by_id(server_id)
+        if server is None:
+            http_action_request.setResponseCode(500)
```
lekhajee: Should this be a 404 server not found like message.

mimic/model/nova_objects.py
lekhajee: Can we create a tickets for this, and link it as an extension to this PR.

mimic/model/nova_objects.py
```diff
@@ -798,6 +798,21 @@ def request_delete(self, http_delete_request, server_id):
...
+            return b''
+        if 'resize' in action_json:
+            flavor = action_json['resize']['flavorRef']
+            server.flavor_ref = flavor
+            http_action_request.setResponseCode(202)
+            return b''
```
lekhajee: Add an `else` that returns a response indication currently supported?

mimic/model/nova_objects.py
```diff
@@ -798,6 +798,21 @@ def request_delete(self, http_delete_request, server_id):
...
+        """
+        Perform the requested action on the provided server
+        """
+        action_json = loads(http_action_request.content.read())
+        server = self.server_by_id(server_id)
+        if server is None:
+            http_action_request.setResponseCode(500)
+            return b''
+        if 'resize' in action_json:
+            flavor = action_json['resize']['flavorRef']
```
lekhajee: If `flavorRef` is not provided, this will fail with `keyerror`. Can we handle that scenario.

mimic/model/nova_objects.py
```diff
@@ -798,6 +798,23 @@ def request_delete(self, http_delete_request, server_id):
...
+        """
+        Perform the requested action on the provided server
+        """
+        server = self.server_by_id(server_id)
+        if server is None:
+            return dumps(not_found("Instance " + server_id + " could not be found",
+                                   http_action_request))
+        action_json = loads(http_action_request.content.read())
+        if 'resize' in action_json:
+            flavor = action_json['resize']['flavorRef']
```
lekhajee: If flavorRef is not provided, this will fail with a keyerror. (I had made the same comment earlier, guess it got lost due to other changes)

Can we handle that scenario as follows:
```
flavor = action_json['resize'].get('flavorRef')
if not flavor:
    return dumps(bad_request("Resize requests require 'flavorRef' attribute.", http_action_request))
```

lekhajee:
@tawalton: I accidentally merged this branch. Could you please address just the one comment in a separate PR and include a test for when `flavorRef` is not provided.

mimic/model/nova_objects.py
glyph: Doesn't that have the problem that if `resize` is not provided you get a `KeyError` as well?

lekhajee: @glyph: The `if` statement is making sure 'resize' exists, failing which the `badRequest` is invoked. Are you suggesting using `get` for it?

glyph: My eyes just slid right over that `if` statement; it's fine as-is :).

PR #350, opened 2015-08-03
move glance images to a json directory inside of canned responses

tawalton:
This is the first step in isolating data that needs to be kept accurate for Reach.  I will be added some additional routes, that will Rackspace server flavors and server images.  The file structure would end up looking something like this:

screen shot 2015-08-03 at 3 24 31 pm


lekhajee:
+1

PR #357, opened 2015-08-04
catch key error exception and test

tawalton:


mimic/model/nova_objects.py
```diff
@@ -808,7 +808,11 @@ def request_action(self, http_action_request, server_id):
                                    http_action_request))
         action_json = loads(http_action_request.content.read())
         if 'resize' in action_json:
-            flavor = action_json['resize']['flavorRef']
+            try:
```
lekhajee: why use `try except` instead of `get`

lekhajee:
+1, good to merge when travis resolves itself.

PR #359, opened 2015-08-05
Support for reverting and confirming a server resize (part of the actions route)

tawalton:


mimic/model/nova_objects.py
```diff
@@ -812,9 +824,26 @@ def request_action(self, http_action_request, server_id):
...
+                http_action_request.setResponseCode(204)
+                return b''
+            elif server.status == 'VERIFY_RESIZE' and 'revertResize' in action_json:
+                server.status = 'ACTIVE'
+                server.flavor_ref = server.oldFlavor
+                http_action_request.setResponseCode(202)
+                return b''
+            else:
+                return dumps(conflicting("Cannot 'revertResize' instance " + server_id +
+                                         " while it is in vm_state active", http_action_request))
```
lekhajee: Should this message be 
```
return dumps(conflicting("Cannot" + action_json.keys()[0]  + "instance " + server_id +
                                         " while it is in vm_state active", http_action_request))
```

mimic/test/test_nova.py
```diff
@@ -618,6 +618,68 @@ def test_server_resize(self):
             }
         })
 
+    def test_confirm_and_revert_server_resize(self):
+        """
+        After a server finishes resizing, the size must be confirmed or reverted
+        A confirmation action should make the server ACTIVE and return a 204
+        A revert action should change the flavor and return a 202
+        Attempting to revert or confirm that is not in VERIFY_RESIZE state returns a 409
```
glyph: Could you add a reference to the nova documentation?  This would also help me, as a lazy reviewer, make sure that this is doing the right thing :).

lekhajee:
+1, modulo comments.

mimic/model/nova_objects.py
tawalton: you are correct.

Issue #360, opened 2015-08-05
Enhance list flavors api call


lekhajee:
The length of flavor list and the flavor types should be set using the servers control plane.

PR #361, opened 2015-08-05
Add flavor list api call

lekhajee:
Currently creates a list of random flavors.


mimic/rest/nova_api.py
```diff
@@ -315,6 +316,16 @@ def get_flavor(self, request, tenant_id, flavor_id):
...
+    @app.route('/v2/<string:tenant_id>/flavors', methods=['GET'])
+    def get_flavor_list(self, request, tenant_id):
+        """
+        Returns a list of flavor with the response code 200.
+        TO DO: The length of flavor list can be set using the control plane.
+               Also be able to set different flavor types in the future.
+        """
+        response_data = get_flavor_list(10)
+        return json.dumps(response_data)
+
```
tawalton: The response code defaults to 200 if not set?

tawalton:
Just one question about  the response code. :+1:

mimic/rest/nova_api.py
lekhajee: yes, it defaults to 200.

mimic/canned_responses/flavors.py
```diff
@@ -0,0 +1,25 @@
...
+    """
+    Populates the flavor template and returns it.
+    """
+    return {"id": "onmetal-" + str(flavor_id),
+            "links": [
+                {"href": "http://this-link-is-not-functional/flavors/" + str(flavor_id),
+                 "rel": "self"},
+                {"href": "http://this-link-is-not-functional/flavors/" + str(flavor_id),
+                 "rel": "bookmark"}],
+            "name": "{0}GB Instance".format(flavor_id)}
```
ClifHouck: I think there's additional data that's returned by nova's flavor.list API call that's not covered here, like the amount of ram is stored in flavor["ram"]. See https://github.com/rackerlabs/arsenal/blob/master/arsenal/director/onmetal_scout.py#L149

lekhajee:
Addressed @ClifHouck's comment by including support for `/flavors/details`.

lekhajee:
@josdotso: As you had enquired about adding support to list flavors in the ##mimic irc channel last week.

mimic/model/nova_objects.py
```diff
@@ -815,6 +816,38 @@ def request_action(self, http_action_request, server_id):
...
 
+    def _create_random_list_of_flavors(self):
+        """
+        Creates a random list of 10 flavors.
+        """
+        for each in range(10):
+            flavor_id = choice([str(each), "onmetal-" + str(each),
+                                "mimic" + str(each)])
+            flavor_name = "{0}GB Instance".format(flavor_id)
+            ram = choice([32768, 131072, 524288, 65536,
```
ClifHouck: I think this is fine, but be aware that OnMetalScout in Arsenal uses amount of ram as a unique key at the moment. So if there are two flavors with the same amount of ram, they will look identical to Arsenal. I plan to update this soon to key off of the flavor metadata added to nodes by @jimrollenhagen.

lekhajee: @ClifHouck : ah ok, as the v2 flavors seemed to have the same ram size right. 

I will add this to be something that can be manipulated using the control plane as indicated in #360.

mimic/rest/nova_api.py
```diff
@@ -315,6 +315,30 @@ def get_flavor(self, request, tenant_id, flavor_id):
...
+        Returns a list of flavor with the response code 200.
+        TO DO: The length of flavor list can be set using the control plane.
+               Also be able to set different flavor types in the future.
+        """
+        return (
+            self._region_collection_for_tenant(tenant_id)
+            .list_flavors(include_details=False, absolutize_url=self.url)
+        )
+
+    @app.route('/v2/<string:tenant_id>/flavors/details', methods=['GET'])
```
tawalton: flavors/detail   http://docs.rackspace.com/servers/api/v2/cs-devguide/content/List_Flavors-d1e4188.html

lekhajee: @tawalton : that is `/flavors` not the `/flavors/details`, for which I could not find docs.

tawalton: the route on line 318 above is the /flavors route and the route on line 330 is /flavors/details.   The details route should be detail

lekhajee: ah, thanks :)

mimic/model/flavor_objects.py
```diff
@@ -0,0 +1,76 @@
...
+        ]
+
+    def extra_specs_json(self):
+        """
+        Create a JSON-serializable data structure describing
+        ``OS-FLV-WITH-EXT-SPECS:extra_specs`` for a flavor.
+        """
+        return {
+            "quota_resources": "instances=mimic-instances,ram=mimic-ram",
+            "class": "mimic",
```
jimrollenhagen: Flavor classes are purely a Rackspace construct, I'm not sure if that's desired in mimic or not, but wanted to point it out. Might be the same for all of these extra specs.

That said, extra specs are a completely arbitrary blob, so maybe it's fine since a Nova operator can put anything here. :)

mimic/model/nova_objects.py
```diff
@@ -847,6 +857,58 @@ def request_action(self, http_action_request, server_id):
...
+        then will return 404.
+        """
+        if flavor_id in get_presets['servers']['invalid_flavor_ref']:
+            return dumps(not_found("The resource could not be found.",
+                                   http_get_request))
+        flavor = self.flavor_by_id(flavor_id)
+        if flavor is None:
+            flavor = Flavor(flavor_id=flavor_id,
+                            name=flavor_id + "Mimic Test Instance",
+                            ram=01, tenant_id=self.tenant_id)
```
jimrollenhagen: s/01/1/ maybe?

jimrollenhagen:
This LGTM, a couple nits inline but it's totally mergable. :+1:

lekhajee:
@manishtomar: Could you please run the autoscale tests against this branch and +1 it, as I have changed the implementation of list flavors.

mimic/model/flavor_objects.py
lekhajee: @jimrollenhagen: Also, Reach is using this as well and I was afraid that they might be needing all the Rackspace specific info.

jimrollenhagen: Cool, I think it's fine :)

cyli:
Autoscale tests pass.  :)  LGTM.

lekhajee:
@cyli: thanks much :)

PR #363, opened 2015-08-06
reboot actions

tawalton:
There are some hard coded times in nova_objects  that allow for the setting of intermediary states that occur when a server is rebooted.  I am going to file an issue to address the durations in the future. We will need to devise a way to allow for the duration to be set to any number including 0.  A 0 duration should skip the intermediary server state and just invoke an immediate "ACTIVE" state.

### DIFF
--- a/mimic/test/test_nova.py
+++ b/mimic/test/test_nova.py
@@ -169,6 +169,7 @@ class NovaAPITests(SynchronousTestCase):
             self, [nova_api, NovaControlApi(nova_api=nova_api)]
         )
         self.root = self.helper.root
+        self.clock = self.helper.clock
         self.uri = self.helper.uri
         self.server_name = 'test_server'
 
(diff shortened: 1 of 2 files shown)


glyph:
@tawalton I know this is still WIP, but I just wanted to make it clear why codecov is failing.  codecov has better coverage reporting than coveralls; specifically, it can report on branch coverage rather than just line coverage.  The report linked from the "details" link shows one line in yellow, the "elif", which means that no run ever fails to execute the implicit, empty "else" at the end of that "if" tree, since that's a separate path execution could take through this code.

mimic/test/test_nova.py
```diff
@@ -688,6 +689,92 @@ def test_confirm_and_revert_server_resize(self):
...
+
+        # Soft reboot tests
+        soft_reboot_request = json.dumps({"reboot": {"type": "SOFT"}})
+        soft_reboot = request(
+            self, self.root, "POST",
+            self.uri + '/servers/' + self.server_id + '/action', soft_reboot_request)
+        soft_reboot_response = self.successResultOf(soft_reboot)
+        self.assertEqual(soft_reboot_response.code, 202)
+
+        soft_reboot_server = request(
```
lekhajee: This is not required here, but in the future you could use `json_request` instead and as it returns a tuple of the response and content. Using that you can skip the 2 `successResultOf` and just have one.

lekhajee:
+1 for the nit

PR #364, opened 2015-08-06
Add a verbose logging option to mimic

### LINKED ISSUE #40: content-type should be application/json and not twisted's unfortunate default of text/html
There really ought to be a decorator which does both this and JSON serialization.

### LINKED ISSUE #202: stop setting Request.defaultContentType

### LINKED ISSUE #362: Verbose mode which logs requests and responses
Log the headers and body of every incoming request, as well as the headers, status code, and body of every response.

cyli:
So it logs all incoming requests and outgoing responses.

Fixes #362.
Fixes #202.
Addresses part of #40.

### DIFF
--- a/mimic/resource.py
+++ b/mimic/resource.py
@@ -1,10 +1,27 @@
 """
-Resources for Mimic's core.
+Resources for Mimic's core application.
 """
-
 import json
 
+from io import StringIO
+
+from six import text_type
+
 from twisted.web.resource import NoResource
+from twisted.web.server import Request, Site
+
+try:
+    from twisted.logger import Logger
+except ImportError:
+    from twisted.python.log import msg
+
+    def log(message, **kwargs):
+        """
+        PEP3101-format the message string.
+        """
+        msg(message.format(**kwargs))
+else:
+    log = Logger("mimic").info
 
 from mimic.canned_responses.mimic_presets import get_presets
 from mimic.model.behaviors import BehaviorRegistryCollection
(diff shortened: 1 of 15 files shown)


cyli:
Hmm... the coverage diff it calculated is wrong.   I think codecov was having trouble with the coverage diff because of the merge from master?  It only seemed to display my last commit.  Rebasing to see if that fixes it.

mimic/resource.py
```diff
@@ -130,3 +147,70 @@ def get_service_resource(self, request, service_id, region_name):
             # workaround for https://github.com/twisted/klein/issues/56
             return NoResource()
         return service_object
+
+
+class MimicRequest(Request):
```
lekhajee: Per @glyph's observation, `MimicRequest` should be inheriting `object` as against `MimicLoggingRequest`

mimic/resource.py
```diff
@@ -1,10 +1,27 @@
...
 import json
 
+from io import StringIO
+
+from six import text_type
+
 from twisted.web.resource import NoResource
+from twisted.web.server import Request, Site
+
+try:
```
lekhajee: Could we please add a comment to explain this, so we know to get rid of it when we make decisions on versions supported and such.

lekhajee:
+1 lgtm modulo the nits

lekhajee:
@cyli: merging this, could you please do a follow up PR for the nits. Thanks !

cyli:
@lekhajee Will do, thanks!

PR #365, opened 2015-08-06
Use codecov instead of coverage

cyli:
Codecov provides unified coverage reports across all the tox environments, and does branch coverage not just line coverage like coveralls.


cyli:
I don't think this is displaying the codecov status because without a reference master sha it can't seem to compute the patch.  :|  It is picking up the commit though.  https://codecov.io/github/rackerlabs/mimic/commits

Issue #369, opened 2015-08-09
Run branch coverage as codecov does using tox


lekhajee:
Currently the test coverage running on tox is not the same as the one on codecov. It would be nice if we could get the same kind of feedback as codecov locally so we don't have to wait for travis builds to finish to realize we are missing some test coverage.

glyph:
Is isn't? The `coverage report` in tox has been showing sub-100% coverage since the beginning, because it *is* the same as the one on codecov; or so I thought.

lekhajee:
yes, but it does not indicate the lines missing branch coverage, which is a bummer.

cyli:
We can probably add a directive to do `coverage html` (and generate an HTML coverage report for every environment that tox is running on) as well as `coverage report -m`

glyph:
I don't know if `tox` is the right place to do the `coverage html`, because we would ideally want to unify the coverage from multiple environments.  Perhaps we could have a script?

PR #375, opened 2015-08-13
Generate HTML coverage reports in tox

cyli:
Fixes https://github.com/rackerlabs/mimic/issues/369

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,5 +8,6 @@ twistd.pid
 docs/_build
 _tmp
 _trial_temp
+_htmlcov
 .tox
 *.egg-info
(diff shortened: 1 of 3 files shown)


glyph:
I changed some behavior in #474 that would make this even harder.  Basically, tox does not have the tools we need for this, which is to wait until all the potentially parallelizable jobs are done (the test runs) and do something with the artifacts produced.

If you want a coverage html report,"`coverage erase && detox && coverage combine && coverage html`" is the right way to get one, and maybe we should have a script for that, but we won't be able to put it in the tox config.  So closing.

Issue #318, opened 2015-06-13
drop python 2.6 support


glyph:
The final release of Python 2.6 was Python 2.6.9 https://www.python.org/download/releases/2.6.9/ on October 9, 2013.  Since then - at this writing, for almost 2 years - Python 2.6 has been officially retired and is now unsupported.

[Twisted is moving to drop Python 2.6 support](https://twistedmatrix.com/pipermail/twisted-python/2015-June/029549.html) because it is a maintenance burden, and it is easier to span the 2.x / 3.x gap when the only 2.x release you have to worry about is 2.7, and the Twisted development team does not wish to encourage the use of unsupported and therefore potentially insecure software.

If Mimic has users who are using python 2.6 they should speak up so that we can advocate for Twisted to continue supporting it; otherwise we should prepare to turn off the Travis builders and announce that we're 2.7-only.

sam-falvo:
When should any Python 2.6 users speak up by?  E.g., what is the deadline for deciding?  (Personally, I +1 this proposal.)

glyph:
Twisted has at least one more release left with 2.6 support, which will probably go out in somewhere between one and three months.  HI @hawkowl.  Any interested parties should pipe up before that goes out, because 2.7-only code might start landing as soon as it's done.  That said, *probably* it will still kinda work for a while once the buildbots are gone, so as long as you get in before the next release *after* that with a really compelling reason, Twisted could restore 2.6 support for some amount of time.

ksheedlo:
+1 for dropping support for old things!

hawkowl:
The Twisted 15.3 release is pegged for July.

ksheedlo:
Twisted 15.3 has landed and no one so far has voiced disagreement. Can we move forward with this?

glyph:
@ksheedlo - I guess we should.  Like Twisted, we should probably do a final release that supports 2.6, and then drop support.  I'll try to get to that in the coming week.

hawkowl:
PURGE THE 2.6

ksheedlo:
I'm going to cite this for each PR that fails because of 2.6 until it gets resolved.

lekhajee:
The Ele team was running mimic on py26 last I checked. We have to make sure they have upgraded before we do this.

lekhajee:
@jayofdoom suggested a great idea of printing a deprecation warning so we can warn users before we actually deprecate py26

ksheedlo:
+1 for a deprecation warning.

[...]

PR #378, opened 2015-08-26
feat(maas_api): add audit logging

### LINKED ISSUE #377: Support MaaS Audit Log
http://docs.rackspace.com/cm/api/v1.0/cm-devguide/content/GET_listAudits_audits_service-account.html

MaaS keeps a log of requests the user performed against their service. Intelligence uses this log to power a widget in the dashboard, which we want to test in Mimic. Log entries map roughly 1:1 to API requests and carry the following information:

- `id` - a UUID, presumably for the API transaction.
- `timestamp` - self explanatory
- `headers` - a hash of all the headers that were sent with the API call.
- `url` - the HTTP url that was used, for instance `/v1.0/123456/entities/enaBcDeFg/test-check`
- `app` - varies depending on the resource. The following apps exist, but may not be the complete set:
    - `agent_installers`
    - `alarms`
    - `checks`
    - `entities`
    - `notification_plans`
    - `notifications`
    - `rollups`
    - `suppressions`
- `query` - a hash of the query string parameters
- `txnId` - a long string including parts of the internal URL for MaaS, a timestamp, and other things. I would fake this with a number or generally some random string.
- `payload` - The HTTP request body, if any.
- `method` - The HTTP method
- `account_id` - The tenant ID
- `who` - I see this always come back as an empty string, don't know what it's for.
- `why` - Ditto.
- `statusCode` - The HTTP status code that was returned.

To implement this feature, we need to get the data, or a reasonable-looking fake, and stash it somewhere in case the user asks for it. I want to look at doing it with a decorator, as it would otherwise be a lot of boilerplate that potentially every new MaaS op needs to be aware of. @glyph what are your thoughts about this? Can we fit what's needed into our existing klein routes?

ksheedlo:
This change implements audit logging as featured in the real MaaS API.
Each destructive operation creates an entry in the audit log, which can
be found at `/audits`.

Fixes #377.

### DIFF
--- a/mimic/rest/maas_api.py
+++ b/mimic/rest/maas_api.py
@@ -5,6 +5,7 @@ MAAS Mock API
 import json
 import collections
 import time
+import urlparse
 import random
 import re
 from uuid import uuid4
@@ -123,6 +124,7 @@ class MCache(dict):
                                                                   with leading + and country \
                                                                   code (E.164 format)'}]}]
         self.suppressions_list = []
+        self.audits_list = []
 
 
 def create_entity(params):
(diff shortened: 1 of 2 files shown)


ksheedlo:
Update: had to fix dict comprehension, since it's not supported in 2.6. Have we considered dropping 2.6?

ksheedlo:
^ to answer myself, we have: #318

lekhajee:
+1

PR #380, opened 2015-08-26
rebuild action and tests

tawalton:


mimic/model/nova_objects.py
```diff
@@ -915,6 +915,26 @@ def request_action(self, http_action_request, server_id):
...
+                    server.update_status,
+                    u"ACTIVE")
+                server_details = server.detail_json(absolutize_url)
+                server_details['adminPass'] = 'password'
+                return dumps({"server": server_details})
+            else:
+                return dumps(conflicting("Cannot 'rebuild' instance " + server_id +
+                                         " while it is in task state other than active",
+                                         http_action_request))
+
```
ksheedlo: This is a long method that performs a variety of different disparate tasks. It might be worth splitting it into pieces and having the API handler select which method to call.

Nova isn't really my area of expertise, though, so don't let this hold you up if someone else +1's.

mimic/test/test_nova.py
```diff
@@ -851,6 +851,53 @@ def test_change_password(self):
...
+            self.uri + '/servers/' + server_id + '/action', rebuild_request))
+        self.assertEqual(response.code, 409)
+        self.assertEqual(body, {
+            "conflictingRequest": {
+                "message": "Cannot 'rebuild' instance " + server_id +
+                           " while it is in task state other than active",
+                "code": 409
+            }
+        })
+
```
ksheedlo: Tests looks good :100:

mimic/model/nova_objects.py
```diff
@@ -822,7 +822,7 @@ def request_delete(self, http_delete_request, server_id):
         server.update_status(u"DELETED")
         return b''
 
-    def request_action(self, http_action_request, server_id):
+    def request_action(self, http_action_request, server_id, absolutize_url):
```
gabecase: Can we change `absolutize_url` variable name to be non-verb such as `absolute_url` since absolutize is a verb and sounds like a method name.

glyph: @gabecase - if you look at what `absolutize_url` actually *is*, you will notice that it is a function which converts a relative URL to an absolute URL, *not* an absolute URL, so its name is accurately describing what it's doing; as it is a function and not data, a verb, not a noun, is appropriate.

mimic/model/nova_objects.py
tawalton: @ksheedlo  I agree with you. I had talked to glyph about this a while ago when we only had support  for 3 actions. Now that more actions are supported, it is time to refactor.  I would rather not do it in this PR, but in follow up PR that will add final action, create image, to the route.  What API handler are you referring to?

glyph: @tawalton - can you create an issue for the refactoring? I tend to agree that that should be a separate change.

glyph: Actually, looking at just *how* long this is, it might even be a good idea to have multiple PRs that split out one thing at a time.

tawalton: @glyph I will write that issue up Monday morning. Can I get +1 on this PR so I can merge it in?

tawalton: Issue created, https://github.com/rackerlabs/mimic/issues/386

ksheedlo: I think you've got it from here, the API handler I was referring to is the handler for `'/v2/<string:tenant_id>/servers/<string:server_id>/action'` that is the only caller of the method.

ksheedlo:
LGTM :+1:

PR #381, opened 2015-08-26
Add rackspace flavors

tawalton:


mimic/model/flavor_objects.py
```diff
@@ -71,6 +66,214 @@ def detailed_json(self, absolutize_url):
...
+               "4 GB Standard Instance": {"id": "5", "ram": 4096, "vcpus": 2, "rxtx_factor": 400,
+                                          "disk": 160},
+               "8 GB Standard Instance": {"id": "6", "ram": 8192, "vcpus": 4, "rxtx_factor": 600,
+                                          "disk": 320},
+               "15 GB Standard Instance": {"id": "7", "ram": 15360, "vcpus": 6, "rxtx_factor": 800,
+                                           "disk": 620},
+               "30 GB Standard Instance": {"id": "8", "ram": 30720, "vcpus": 8, "rxtx_factor": 1200,
+                                           "disk": 1200}}
+
+    def extra_specs_json(self):
```
lekhajee: Is there a reason this cant be an attribute on the `Flavor` object. Similarly for the others.

tawalton: @lekhajee the data in extra_specs_json is different for each flavor class. I would like to leave it here since it is specific info for each each flavor

lekhajee:
+1

PR #382, opened 2015-08-27
Ironic: increasing the default nodes listed 

lekhajee:


mimic/model/ironic_objects.py
```diff
@@ -219,15 +219,15 @@ def list_nodes(self, include_details):
         http://docs.openstack.org/developer/ironic/webapi/v1.html#get--v1-nodes-detail
         """
         if not self.ironic_node_store:
-            for each in range(3):
+            for each in range(30):
```
ksheedlo: style nit: if `each` is not used in the body of the loop, `_` can be used as a throwaway iterator variable. Helps express that the loop is run purely for the side effect.

mimic/test/test_ironic.py
```diff
@@ -84,9 +84,9 @@ def test_list_nodes_with_details(self):
         instance_nodes = [node['extra']['flavor'] if node['extra'].get('flavor')
                           else None
                           for node in content['nodes']]
-        self.assertEqual(instance_nodes.count('onmetal-io1'), 5)
-        self.assertEqual(instance_nodes.count('onmetal-compute1'), 3)
-        self.assertEqual(instance_nodes.count('onmetal-memory1'), 3)
+        self.assertEqual(instance_nodes.count('onmetal-io1'), 32)
```
ksheedlo: Noticing that this is 2 more than the default - where do those come from?

ksheedlo: Nevermind - these are the 2 without the configured flavor at the end.

ksheedlo:
LGTM :+1:

lekhajee:
@ksheedlo : thanks for the review

PR #384, opened 2015-08-28
feat(maas_plugin): add MaasControlApi

ksheedlo:
This change adds an empty control API for MaaS. Future plans for this
control API include

1. Refactoring some of the in-band control mechanisms already used by
   the `MaasApi` into the control API; and
2. Implementing some of the changes from the `toksPatches` branch that
   needed the control plane to be done the right way.

**Update:** This change also implements a `test-alarm` API for MaaS as a first example of what we can do with the MaaS control API.

### DIFF
--- a/mimic/model/maas_objects.py
+++ b/mimic/model/maas_objects.py
@@ -3,46 +3,52 @@ MaaS API data model
 """
 
 import random
-import time
 
 
-def random_letters(len_min, len_max=None):
+def random_letters(len_min, len_max):
     """
     Makes a string with random letters (a-z).
     """
-    if len_max is None:
-        len_max = len_min
     return ''.join([chr(ord('a') + random.randint(0, 25))
                     for _ in xrange(random.randint(len_min, len_max))])
 
 
-class TestCheckMetric(object):
+class Metric(object):
     """
-    Models a metric from the test-check API
+    Models a MaaS metric type.
     """
-    def __init__(self, name, type, unit='other'):
+    def __init__(self, name, type, clock, unit='other'):
         """
-        Constructs a test check metric data source
+        Initializes the metric.
         """
         self.name = name
         self.type = type
         self._unit = unit
-        self._override = None
+        self._overrides = {}
+        self._clock = clock
 
-    def set_override(self, value):
+    def _override_key(self, **kwargs):
         """
-        Sets the override metric value.
+        Computes a key used for hashing overrides, as a tuple of strings.
+        """
+        return (kwargs['entity_id'],
+                kwargs['check_id'],
+                kwargs.get('monitoring_zone', '__AGENT__'))
 
-        We wrap the override value in a dict so that we can distinguish
-        between None (i.e., null) and no override.
+    def set_override(self, **kwargs):
         """
-        self._override = {'value': value}
+        Sets the override metric for a given tenant, entity and check.
 
-    def clear_override(self):
+        Override metrics are defined as a function which takes a timestamp
+        and returns the metric value.
         """
-        Clears the override metric value.
+        self._overrides[self._override_key(**kwargs)] = kwargs['override_fn']
+
+    def clear_overrides(self):
         """
-        self._override = None
+        Clears the override metric values.
+        """
+        self._overrides = {}
 
     def _get_default_data(self):
         """
@@ -60,33 +66,51 @@ class TestCheckMetric(object):
             return random_letters(12, 30)
         raise ValueError('No default data getter for type {0}!'.format(self.type))
 
-    def get_value(self):
+    def get_value_for_test_check(self, **kwargs):
         """
         Gets the metric data object as returned from the test-check API.
         """
-        data = (self._get_default_data() if self._override is None
-                else self._override['value'])
+        override_key = self._override_key(**kwargs)
+        timestamp = kwargs['timestamp']
+        data = self._get_default_data()
+
+        if override_key in self._overrides:
+            data_fn = self._overrides[override_key]
+            data = data_fn(timestamp)
+
         return {'type': self.type,
                 'unit': self._unit,
                 'data': data}
 
 
-class TestCheckData(object):
+class CheckType(object):
     """
-    Models test-check response data
+    Data model for a MaaS check type (e.g., remote.ping).
     """
-    def __init__(self, metrics):
+    def __init__(self, clock, metrics):
         """
-        Initializes the test-check responder
+        Constructs a check type with metrics.
         """
         self.metrics = metrics
-        self.available = True
-        self.status = "code=200,rt=0.4s,bytes=99"
-        self.response_code = 200
+        self.test_check_available = {}
+        self.test_check_status = {}
+        self.test_check_response_code = {}
+        self._clock = clock
+
+    def clear_overrides(self):
+        """
+        Clears the overrides for test-checks and metrics.
+        """
+        self.test_check_available = {}
+        self.test_check_status = {}
+        self.test_check_response_code = {}
+
+        for metric in self.metrics:
+            metric.clear_overrides()
 
     def get_metric_by_name(self, metric_name):
         """
-        Gets the metric on this test-check.
+        Gets the metric on this check type.
 
         This method is useful for setting and clearing overrides on the
         test metrics.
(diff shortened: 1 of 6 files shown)


lekhajee:
@ksheedlo : could you please add the labels going forward :)

ksheedlo:
Update: Added a new `test-alarm` API supported by the control API.

ksheedlo:
I'm noticing that the `test-check` API is going to want some controls to be scoped to the entity, so I'm going to do that to the `test-alarm` API as well. Stand by for an update.

ksheedlo:
Update: Added a `test-check` API with hooks in the control backplane. It might be a good idea to split this PR, since it got huge.

ksheedlo:
There's a design issue I want to work out here. Should `test-alarm` pull responses out of a queue from the control API as it does now, or should it return random responses until the control API pins a response? The latter feels more in line with convention (see for instance, the multiplot metrics call).

ksheedlo:
I like this better. RFR again.

mimic/model/maas_objects.py
```diff
@@ -0,0 +1,190 @@
...
+
+test_check_responses['remote.http'] = lambda: TestCheckData([
+    TestCheckMetric('rx_bytes', 'i', 'bytes'),
+    TestCheckMetric('rx_dropped', 'i', 'bytes'),
+    TestCheckMetric('rx_errors', 'i', 'bytes'),
+    TestCheckMetric('rx_packets', 'i', 'bytes'),
+    TestCheckMetric('tx_bytes', 'i', 'bytes'),
+    TestCheckMetric('tx_dropped', 'i', 'bytes'),
+    TestCheckMetric('tx_errors', 'i', 'bytes'),
+    TestCheckMetric('tx_packets', 'i', 'bytes')])
```
ksheedlo: Needs `remote.ping`.

ksheedlo:
Build failed because test cases don't have an `assertLess` method in 2.6 - #318

mimic/rest/maas_api.py
```diff
@@ -616,6 +620,21 @@ def delete_check(self, request, tenant_id, entity_id, check_id):
...
+        """
+        test_config = json.loads(request.content.read())
+        check_type = test_config['type']
+        check_responses = self._entity_cache_for_tenant(tenant_id).test_check_responses
+        if (entity_id, check_type) not in check_responses:
+            check_responses[(entity_id, check_type)] = get_test_check_response_by_type(check_type)
+
+        response_code, response_body = check_responses[(entity_id, check_type)].get_response()
+        request.setResponseCode(response_code)
+        return json.dumps(response_body)
```
ksheedlo: needs audit

mimic/rest/maas_api.py
```diff
@@ -707,6 +726,32 @@ def delete_alarm(self, request, tenant_id, entity_id, alarm_id):
...
+                                             'status', 'Matched default return statement'),
+                                         'timestamp': int(time.time() * 1000)})
+        else:
+            for i in xrange(n_tests):
+                response_payload.append({'state': random.choice(['OK', 'WARNING', 'CRITICAL']),
+                                         'status': random.choice(['cats', 'dogs', 'chickens']),
+                                         'timestamp': int(time.time() * 1000)})
+
+        request.setResponseCode(200)
+        return json.dumps(response_payload)
```
ksheedlo: needs audit

mimic/rest/maas_api.py
```diff
@@ -1251,3 +1301,122 @@ def multiplot(self, request, tenant_id):
...
+    Klein routes for MaaS control API.
+    """
+
+    def _entity_cache_for_tenant(self, tenant_id):
+        """
+        Retrieve the M_cache object containing all objects created so far
+        """
+        return (self.session_store.session_for_tenant_id(tenant_id)
+                .data_for_api(self.api_mock.maas_api,
+                              lambda: collections.defaultdict(MCache))[self.region])
```
ksheedlo: Codecov gives this 1/2 because tests never execute inside the lambda. Any ideas how to work around this?

cc @lekhajee

ksheedlo:
Currently testing against Intelligence and I'm not happy with the way `test-check` handles monitoring zones.

ksheedlo:
Update: Completely rewrote the in-memory model for testing checks. This version is much better at working with monitoring zones.

lekhajee:
@ksheedlo : please fix conflicts

ksheedlo:
@lekhajee Updated and fixed conflicts.

ksheedlo:
Back to WIP because we want to fake out parse errors in the `test-alarm` API.

ksheedlo:
Update: Added another endpoint for pushing errors to the `test-alarm` API.

mimic/model/maas_objects.py
```diff
@@ -0,0 +1,229 @@
+"""
+MaaS API data model
+"""
+
+import random
+
+
+def random_letters(len_min, len_max):
```
lekhajee: Any reason we cant use https://github.com/rackerlabs/mimic/blob/master/mimic/util/helper.py#l37

mimic/model/maas_objects.py
```diff
@@ -0,0 +1,229 @@
...
+    """
+    Makes a string with random letters (a-z).
+    """
+    return ''.join([chr(ord('a') + random.randint(0, 25))
+                    for _ in xrange(random.randint(len_min, len_max))])
+
+
+class Metric(object):
```
lekhajee: Not essential, but how about using characteristic

mimic/model/maas_objects.py
ksheedlo: Nope, good eye, will use that.

mimic/model/maas_objects.py
```diff
@@ -0,0 +1,229 @@
...
+    """
+    Models a MaaS metric type.
+    """
+    def __init__(self, name, type, clock, unit='other'):
+        """
+        Initializes the metric.
+        """
+        self.name = name
```
glyph: Not a blocking comment, but, for simple objects like this, you should probably be using `@characteristic.attributes` to declare it; objects declared with `characteristic` get a bunch of things for free:

- a nice `repr`
- comparison operators so that instances can be sorted
- hashing operators (which, in this case, would make this object *not* hashable because it contains a dictionary: which is what you want)
- an API for enumerating attributes
- optional metadata for types of attributes

For more detail on the differences, see http://characteristic.readthedocs.org/en/stable/why.html#hand-written-classes

There's no *specific* reason that a `Metric` absolutely needs any of these features, but it's helpful if we consistently do this throughout the codebase, so that more objects provide these features.

mimic/model/maas_objects.py
```diff
@@ -0,0 +1,229 @@
...
+        """
+        self._overrides[self._override_key(**kwargs)] = kwargs['override_fn']
+
+    def clear_overrides(self):
+        """
+        Clears the override metric values.
+        """
+        self._overrides = {}
+
+    def _get_default_data(self):
```
glyph: I don't think I quite get this; it seems like a `Metric` should have a single fixed value, or at least a fixed value for a particular point in time.  It looks here like you'll request the same metric and get a randomly different value?

The use of the `random` module all over the place worries me a little, too.  There are two use-cases for randomness, generally.  In one case, you want deterministic random values, in which case you generally want an instance of `random.Random` that has a given seed; in another case, you want truly unpredictable cryptographic randomness, in which case you want an instance of `random.SystemRandom`.  There isn't really a case where you want to use the global functions in the `random` module (and in fact there is a discussion going on right now in the python community about possibly deprecating those methods entirely).

I think that we may want `MimicCore` to have a source of randomness, a `random` attribute, like `clock`, which gets passed around to do these various randomness-generation things, so that users can select which source of randomness is appropriate for their test setup.  By default it should probably be `SystemRandom` so it's truly unpredictable, since most places we are using randomness are things that tests should be ignoring and able to cope with being non-deterministic…

mimic/model/maas_objects.py
```diff
@@ -0,0 +1,229 @@
...
+        """
+        Gets the default data point. This data point may be overridden by
+        setting the override value.
+        """
+        if self.type == 'i':
```
glyph: It would be helpful to those who are not as familiar with the MaaS API to hoist these magic codes (`'i'`, `'n'`, `'s'`) into named / documented constants.

mimic/model/maas_objects.py
```diff
@@ -0,0 +1,229 @@
...
+
+class CheckType(object):
+    """
+    Data model for a MaaS check type (e.g., remote.ping).
+    """
+    def __init__(self, clock, metrics):
+        """
+        Constructs a check type with metrics.
+        """
+        self.metrics = metrics
```
glyph: (similar note about `characteristic` above)

mimic/model/maas_objects.py
```diff
@@ -0,0 +1,229 @@
...
+class MaasStore(object):
+    """
+    A collection of MaaS configuration objects.
+    """
+    def __init__(self, clock):
+        """
+        Initializes the Maas configuration using the provided clock.
+        """
+        self.check_types = {
+            'agent.cpu': CheckType(clock, [
```
glyph: Could you provide a documentation citation in a docstring to indicate where this blob of values comes from?

mimic/rest/maas_api.py
```diff
@@ -354,6 +359,31 @@ def parse_and_flatten_qs(url):
     return flat_qs
 
 
+def _random_hipsum():
+    """
+    Generates a random sentence using Hipsum ( http://hipsum.co/ ).
```
glyph: This should document its return value and type: is this bytes or text? If it's text the values below should be `unicode` objects; otherwise they should have a `b` prefix on them.

mimic/rest/maas_api.py
```diff
@@ -354,6 +359,31 @@ def parse_and_flatten_qs(url):
...
+        "bespoke migas art party crucifix next level. Street art chia bitters, ",
+        "gastropub mixtape flexitarian Godard occupy lumbersexual."]).split(' ')
+    offset = random.randint(1, len(hipsum))
+    rotated = hipsum[offset:] + hipsum[:offset]
+    return ' '.join(rotated[:12])
+
+
+def _mcache_factory(clock):
+    """
+    Makes a defaultdict that makes MCache objects for each tenant.
```
glyph: Not quite correct; returns a function *that* makes a defaultdict that…

mimic/rest/maas_api.py
```diff
@@ -354,6 +359,31 @@ def parse_and_flatten_qs(url):
     return flat_qs
 
 
+def _random_hipsum():
```
glyph: This should probably be in `mimic/util/helper.py` since text-generation isn't really specific to MaaS.

mimic/rest/maas_api.py
```diff
@@ -629,6 +658,23 @@ def delete_check(self, request, tenant_id, entity_id, check_id):
         self._audit('checks', request, tenant_id, status)
         return ''
 
+    @app.route('/v1.0/<string:tenant_id>/entities/<string:entity_id>/test-check', methods=['POST'])
+    def test_check(self, request, tenant_id, entity_id):
+        """
+        Tests a check
```
glyph: This docstring leaves something to be desired :).

mimic/model/maas_objects.py
ksheedlo: Here's the theory I worked out as I hashed this out a few times. In MaaS, a Metric is a piece of data that represents some specific value, and the value probably changes over time. So, when writing the control API, what you really want to do is to be able to specify a function for a given `(entity, check, metric)` triple that varies over time (or not). It's easy to get from here, to the multiplot mock which returns either random noise or a square wave.

It's also easy to invent metrics to come back with while testing checks without relying on magic numbers; recall the discussion on #275. They're just random, and the user can override the metric results with a custom function (maybe a constant value) that they choose.

mimic/rest/maas_api.py
```diff
@@ -720,6 +766,43 @@ def delete_alarm(self, request, tenant_id, entity_id, alarm_id):
         self._audit('alarms', request, tenant_id, status)
         return ''
 
+    @app.route('/v1.0/<string:tenant_id>/entities/<string:entity_id>/test-alarm', methods=['POST'])
+    def test_alarm(self, request, tenant_id, entity_id):
+        """
+        Test an alarm, pulling responses from the simulated test-alarm queue.
```
glyph: If I were reading this docstring trying to understand where the "simulated test-alarm queue" was set up, I'd be confused.  Can you include a reference for folks to understand how said queue is populated?  "``` :obj:`a.python.name` ```" is a good way to do this if you are just referring to a Python API.

mimic/model/maas_objects.py
ksheedlo: As far as the source of randomness - I completely agree, but the infrastructure isn't in place yet. We should tackle that issue separately.

mimic/model/maas_objects.py
ksheedlo: +1, working on it.

mimic/rest/maas_api.py
ksheedlo: true :100:

mimic/rest/maas_api.py
ksheedlo: Heh

mimic/rest/maas_api.py
ksheedlo: That's actually outdated from an earlier iteration. I'll fix it.

mimic/rest/maas_api.py
```diff
@@ -1264,3 +1347,144 @@ def multiplot(self, request, tenant_id):
...
+        test_responses = self._entity_cache_for_tenant(tenant_id).test_alarm_responses
+        dummy_response = json.loads(request.content.read())
+        test_responses[entity_id] = []
+        for response_block in dummy_response:
+            ith_response = {'state': response_block['state']}
+            if 'status' in response_block:
+                ith_response['status'] = response_block['status']
+            test_responses[entity_id].append(ith_response)
+        request.setResponseCode(204)
+        return ''
```
glyph: No-content responses should generally be bytes; to make the 3.x transition easier, it's best to be doing `b''` everywhere you're returning empty strings.

mimic/test/test_maas.py
```diff
@@ -400,6 +431,272 @@ def test_delete_alarm(self):
...
+        resp = self.successResultOf(req)
+        self.assertEquals(resp.code, 200)
+        data = self.get_responsebody(resp)
+        self.assertEquals(456, data[0]['metrics']['duration']['data'])
+
+    def test_test_check_clears_metrics(self):
+        """
+        The test-check control API can clear metrics.
+
+        NB: Randomly generated string metrics are between 12 and 30
```
glyph: The ReST syntax for this would be [`..note:`](http://docutils.sourceforge.net/docs/ref/rst/directives.html#admonitions).

mimic/test/test_maas.py
```diff
@@ -400,6 +431,272 @@ def test_delete_alarm(self):
...
+                      json.dumps({'criteria': 'return new AlarmStatus(OK);',
+                                  'check_data': [{}]}))
+        resp = self.successResultOf(req)
+        self.assertEquals(resp.code, 200)
+        data = self.get_responsebody(resp)
+        self.assertEquals(1, len(data))
+        self.assertEquals('OK', data[0]['state'])
+        self.assertIn('timestamp', data[0])
+
+    def test_test_alarm_setting_status(self):
```
glyph: needs a test docstring

mimic/test/test_maas.py
```diff
@@ -400,6 +431,272 @@ def test_delete_alarm(self):
...
+                      self.uri + '/entities/' + self.entity_id + '/test-alarm',
+                      json.dumps({'criteria': 'return new AlarmStatus(OK);',
+                                  'check_data': [{}]}))
+        resp = self.successResultOf(req)
+        self.assertEquals(resp.code, 200)
+        data = self.get_responsebody(resp)
+        self.assertEquals(1, len(data))
+        self.assertEquals('test status message', data[0]['status'])
+
+    def test_test_alarm_clearing_response(self):
```
glyph: needs a test docstring

mimic/test/test_maas.py
```diff
@@ -400,6 +431,272 @@ def test_delete_alarm(self):
...
+                      self.uri + '/entities/' + self.entity_id + '/test-alarm',
+                      json.dumps({'criteria': 'return new AlarmStatus(OK);',
+                                  'check_data': [{}]}))
+        resp = self.successResultOf(req)
+        self.assertEquals(resp.code, 200)
+        data = self.get_responsebody(resp)
+        self.assertEquals(1, len(data))
+        self.assertNotEquals('test-alarm working OK', data[0]['status'])
+
+    def test_test_alarm_setting_errors(self):
```
glyph: needs a test docstring

glyph:
This is pretty close, and I don't forsee problems, but there's enough feedback here that it probably deserves one more round.

mimic/model/maas_objects.py
ksheedlo: done, stand by for another commit

mimic/rest/maas_api.py
ksheedlo: what is the unicode idiom for the part at the bottom?

```python
u' '.join(rotated[:12])
```

?

mimic/rest/maas_api.py
ksheedlo: fixed, stand by for commits

mimic/rest/maas_api.py
ksheedlo: done, stand by for commits

mimic/rest/maas_api.py
ksheedlo: I'd like to address this separately as the idiom is peppered throughout the MaaS API.

mimic/rest/maas_api.py
ksheedlo: fixed, stand by for commits.

mimic/rest/maas_api.py
ksheedlo: fixed

ksheedlo:
@glyph d0035b8b34ce6f0439585c676ed275c02d04e375 addresses review feedback.

mimic/test/test_maas.py
```diff
@@ -400,6 +432,282 @@ def test_delete_alarm(self):
         self.assertEquals(resp.code, 200)
         self.assertEquals(0, len(self.get_responsebody(resp)['values'][0]['alarms']))
 
+    def test_test_check(self):
+        """
+        The test-check API should return fake test-check results.
+        """
+        req = request(self, self.root, "POST",
```
lekhajee: All these tests could use the `json_request` or `request`(in case on empty response body) from https://github.com/rackerlabs/mimic/blob/master/mimic/test/helpers.py and skip the three lines
```
req = request(self, self.root, "POST", self.uri + '/entities/' + self.entity_id + '/test-check', json.dumps({'type': 'agent.disk'}))
resp = self.successResultOf(req)
data = self.get_responsebody(resp)
```

lekhajee: Eg: https://github.com/rackerlabs/mimic/blob/master/mimic/test/test_ironic.py#l80, https://github.com/rackerlabs/mimic/blob/master/mimic/test/test_ironic.py#l98

mimic/rest/maas_api.py
lekhajee: could we please create an issue for this, if this is not going to addressed in the near future

lekhajee:
+1 modulo the nits

mimic/rest/maas_api.py
glyph: sure; sorry, should have said "non-blocking" :).

ksheedlo:
@lekhajee addressed your comment, will merge when the build is green. Thanks for reviewing! :sparkles:

ksheedlo:
thanks @glyph as well! :sun_with_face:

PR #387, opened 2015-09-02
Add/Delete Ironic nodes

lekhajee:


mimic/model/ironic_objects.py
```diff
@@ -149,13 +153,14 @@ def detail_json(self):
...
             },
-            "properties": {
+            "properties": self.properties or {
                 "memory_mb": self.memory_mb,
                 "cpu_arch": "amd64",
                 "local_gb": 32,
                 "cpus": 40
             },
-            "driver_info": {
+            "driver": self.driver,
```
ksheedlo:
```python
self.driver or "agent_ipmitool"
```

?

mimic/model/ironic_objects.py
```diff
@@ -170,6 +175,8 @@ def detail_json(self):
         if self.instance_uuid:
             template["instance_info"] = self.static_instance_info
             template["provision_state"] = "active"
+        if not self.driver:
+            template["driver"] = "agent_ipmitool"
```
ksheedlo: Not 100% sure the semantics here, but might could be moved up (see comment above)

mimic/model/ironic_objects.py
```diff
@@ -204,8 +216,49 @@ def add_to_ironic_node_store(self, **attributes):
         Create a new Node object and add it to the
         :obj: `ironic_node_store`
         """
+        if not attributes.get('chassis_uuid'):
+            attributes['chassis_uuid'] = str(uuid4())
+        if (attributes.get("flavor_id", None) is None and
+                attributes.get("memory_mb")):
```
ksheedlo: style nit: this could probably be a one-liner

mimic/model/ironic_objects.py
```diff
@@ -204,8 +216,49 @@ def add_to_ironic_node_store(self, **attributes):
...
+            attributes['flavor_id'] = self.memory_to_flavor_map.get(
+                attributes['memory_mb'], "onmetal-mimic")
         node = Node(**attributes)
         self.ironic_node_store.append(node)
+        return node
+
+    def create_node(self, http_create_request):
+        """
+        Create a node
+        http://bit.ly/1N0O9KM
```
ksheedlo: Thanks for the helpful link! :sparkles:

mimic/model/ironic_objects.py
```diff
@@ -204,8 +216,49 @@ def add_to_ironic_node_store(self, **attributes):
...
+            return dumps(node.detail_json())
+        except:
+            http_create_request.setResponseCode(400)
+            return b''
+
+    def delete_node(self, http_delete_request, node_id):
+        """
+        Delete the `node_id` from the :obj:`ironic_node_store` if
+        the node exists and set response code to be 204.
+        If node does not exist, return response code 404.
```
ksheedlo: :sparkles:

mimic/model/ironic_objects.py
lekhajee: So, that was first instinct as well. But for some reason, that I cant figure why, if I did that, all the nodes seem to have self.driver set to the value sent for the first node's driver.

mimic/rest/ironic_api.py
```diff
@@ -21,6 +21,20 @@ def __init__(self, core):
...
+        Responds with response code 201 and returns the newly created node.
+        """
+        return self.core.ironic_node_store.create_node(request)
+
+    @app.route('/nodes/<string:node_id>', methods=['DELETE'])
+    def delete_node(self, request, node_id):
+        """
+        Responds with response code 204 and delete the node.
+        """
+        return self.core.ironic_node_store.delete_node(request, node_id)
```
ksheedlo: I'm a little concerned about this organization. I feel the `IronicNodeStore` should not deal in HTTP requests and response codes - this smells strongly of inappropriate layer mixing to me.

mimic/rest/ironic_api.py
```diff
@@ -40,7 +54,7 @@ def get_node_details(self, request, node_id):
         """
         Responds with response code 200 with details of the nodes.
         """
-        return self.core.ironic_node_store.get_node_details(node_id)
+        return self.core.ironic_node_store.get_node_details(request, node_id)
```
ksheedlo: Likewise here - passing an HTTP request object to a data store is smelly.

mimic/test/test_ironic.py
```diff
@@ -45,19 +45,114 @@ def setUp(self):
...
+        # get node
+        (response, get_content) = self.successResultOf(json_request(
+            self, self.root, "GET", self.url + '/' + node_id))
+        self.assertEqual(200, response.code)
+        self.assertEqual(content, get_content)
+
+    def test_create_then_get_node_2(self):
+        """
+        Test create node then get the node and verify attributes
+        """
```
ksheedlo: This test case is almost exactly the same as the one above it. Is it a duplicate?

mimic/test/test_ironic.py
```diff
@@ -45,19 +45,114 @@ def setUp(self):
...
+        response = self.successResultOf(request(
+            self, self.root, "DELETE", self.url + "/1234"))
+        self.assertEqual(response.code, 404)
+
+    def test_delete_node(self):
+        """
+        Delete node returns 204 deletes the given node_id.
+        """
+        # create node
```
ksheedlo: "create node" comments are unnecessary because the method name is obvious. "delete node" and "get node" might be ok because HTTP noise is cluttering up those operations. Still, I feel the marginal value of these comments is low.

mimic/model/ironic_objects.py
ksheedlo: weird

ksheedlo:
Otherwise tests looks good.

mimic/rest/ironic_api.py
lekhajee: hmm, this is the pattern followed in the nova_objects.py as well, and was moved from the api layer to model layer during the implementation of the object model. If I remember right, this was proposed by @glyph. 

@glyph: any thoughts?

ksheedlo: @lekhajee I'd appreciate context as well if you have any.

mimic/test/test_ironic.py
lekhajee: This test is different from the first one in the sense that it is creating a node with `self.create_request` json, which has the driver_info, properties and driver properties.

lekhajee:
@ksheedlo : thanks much for the quick reviews :)

mimic/test/test_ironic.py
ksheedlo: maybe `create_node_with_default_attributes`? My main issue with this is I'd prefer a meaningful spec name over `test_some_other_spec_2`

mimic/model/ironic_objects.py
jayofdoom: I think this is fine without a default, as Ironic doesn't have one. If you want to use one, I'd use the "fake" one for testing.

mimic/test/test_ironic.py
```diff
@@ -45,19 +45,114 @@ def setUp(self):
             "power_state", "instance_info", "ports", "name", "driver_internal_info",
             "inspection_finished_at", "inspection_started_at", "clean_step"
         ]
+        self.url = "/ironic/v1/nodes"
+        self.create_request = {
+            "chassis_uuid": str(uuid4()),
+            "driver": "fake",
+            "driver_info": {"cache_image_id": None,
```
jayofdoom: I would not add the caching values here. In fact, for driver_info, typically all you want when creating is creds to use the driver (in the case of fake driver, I think an empty dict is fine).

mimic/rest/ironic_api.py
jimrollenhagen: I agree and thought about commenting on this in the last Ironic review... however that seems to be how mimic is organized already, so this seems fine to me for now. (But we should totally refactor that one day :)

jimrollenhagen:
This seems good to me :+1:

mimic/rest/ironic_api.py
glyph: I agree that it *seems* like there's a layering violation here - and honestly, maybe there is one and we could figure out a better structure - but Mimic is a bit of a special case in terms of the correspondence between its "model" later and its "view" (or public-facing HTTP API layer).  Mimic's models are modeling not just the abstract state of "servers" or "load balancers" or whatever, but *API behaviors* as well, which means that in the "model" layer we occasionally need representations of things like "the backend is down, return a `503` for some reason" or "randomly truncate the HTTP response in the middle, we saw that happen when talking to the load balancer API one time when trying to manipulate this one messed up CLB".  As such, passing along the full HTTP request and all its capabilities seems like the best option available, although I'd love to hear other thoughts, because this definitely feels not quite right based on every *other* project I've ever worked on :).

ksheedlo: Perhaps we could list out the known use cases for modeling API behaviors
and see if we can improve on any of them individually before jumping to a
general solution.

For instance, allowing the backend to be down might not have to be specific
to the contents of each backend. Maybe instead Mimic keeps an up/down state
for each service and serves a generic 503 if it's down. Truncating the http
response in the middle could be done using middleware.

I don't expect to solve any new problems in this thread, but it seems to me
that 503s or truncated requests are concerns at the layer of HTTP or lower.
As a plugin author I would rather work and think in the domain of my
application, i.e., HTTP layer + 1. The simplest way I can think of to not
have this layer violation is to do HTTP-related stuff in the "view" and
only business stuff in the data stores, since the view already does HTTP
stuff. There are other possibilities for which we should start an issue or
discuss offline.

PR #391, opened 2015-09-03
fix(maas_api): prefer mimic clock to time.time()

ksheedlo:
Addresses https://github.com/rackerlabs/mimic/pull/79#discussion_r22687955.

### DIFF
--- a/mimic/rest/maas_api.py
+++ b/mimic/rest/maas_api.py
@@ -4,7 +4,6 @@ MAAS Mock API
 
 import json
 import collections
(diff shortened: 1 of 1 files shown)


ksheedlo:
I want to test Intelligence against this first to make sure it doesn't
break us.

ksheedlo:
This doesn't cause any Intelligence tests to fail that won't pass on my laptop before, so RFR.

lekhajee:
+1

PR #393, opened 2015-09-03
Add node has no required attributes

lekhajee:


mimic/model/ironic_objects.py
```diff
@@ -229,9 +229,12 @@ def create_node(self, http_create_request):
         """
         content = loads(http_create_request.content.read())
         try:
+            memory_mb = 131072
```
jimrollenhagen: Should there be a default for this or should it be allowed to be empty?

mimic/model/ironic_objects.py
```diff
@@ -229,9 +229,12 @@ def create_node(self, http_create_request):
         """
         content = loads(http_create_request.content.read())
         try:
+            memory_mb = None
+            if content.get('properties'):
+                memory_mb = content.get('memory_mb')
```
ksheedlo: should probably be

```python
if 'properties' in content:
    memory_mb = content['properties'].get('memory_mb')
```

since `memory_mb` is a property of the `properties`, not the content itself.

lekhajee: oops, thanks!

ClifHouck:
LGTM

PR #395, opened 2015-09-04
ensure uniqueness of private IPs within a tenant

### LINKED ISSUE #383: Prevent IP collision within a tenant
Since IPs are randomly generated, they might collide with each other.

Autoscale is encountering an issue where, because the the tests are generating so many servers, one autoscale group is removing an IP from a CLB that belongs to one of its servers.

However, that IP also belongs to a different server created by a different test, and hence removing it from the CLB breaks the test.

We should probably have a function `ensure_unique` or something that takes callable which generates a random value, and compares it to a set of already-generated random values to ensure that it produces a unique value.

Autoscale only cares about collision within a single tenant, but probably this should be across all tenants to behave more like the cloud.

radix:
quick fix to make sure private IPs are unique within a tenant, with test.

Fixes #383.

### DIFF
--- a/mimic/test/test_nova.py
+++ b/mimic/test/test_nova.py
@@ -2294,6 +2294,10 @@ class NovaAPIMetadataTests(SynchronousTestCase):
 
 class NovaServerTests(SynchronousTestCase):
     def test_unique_ips(self):
+        """
+        The private IP address of generated servers will be unique even if
+        the given ``ipsegment`` factory generates non-unique pairs.
+        """
         nova_api = NovaApi(["ORD", "MIMIC"])
         self.helper = self.helper = APIMockHelper(
             self, [nova_api, NovaControlApi(nova_api=nova_api)]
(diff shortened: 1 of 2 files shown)


mimic/test/test_nova.py
```diff
@@ -2288,3 +2290,33 @@ def test_set_metadata_item_too_many_metadata_items_takes_precedence(self):
                         for i in xrange(40))
         self.assert_maximum_metadata(
             *self.set_metadata_item(metadata, 'key', {"meta": {"key": []}}))
+
+
+class NovaServerTests(SynchronousTestCase):
+    def test_unique_ips(self):
```
glyph: Can you add a test docstring here?

glyph:
Looks like my adventure down the rabbit hole was too elaborate :-).  Modulo the test docstring and the lint failure this looks good to me.

radix:
Thanks for the review @glyph! I've also addressed a minor lint issue. I'll merge this after the build passes.

PR #398, opened 2015-09-04
fix(maas_api): correctly determine hostname

### LINKED ISSUE #396: Hard coded port in maas_api
https://github.com/rackerlabs/mimic/blob/9a9ea300fc7b934ad8494fcabe22daeb0f044056/mimic/rest/maas_api.py#L362

Has the port number hard coded to 8900.

We recently started running mimic in a container on our build slaves. It is possible to force it to bind to local port 8900 and make this work. But ideally we would want Jenkins to manage the ports.

ksheedlo:
Fixes #396.

### DIFF
--- a/mimic/rest/maas_api.py
+++ b/mimic/rest/maas_api.py
@@ -17,6 +17,7 @@ from twisted.plugin import IPlugin
 
 from mimic.catalog import Entry
 from mimic.catalog import Endpoint
+from mimic.rest.auth_api import base_uri_from_request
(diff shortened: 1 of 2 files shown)


ksheedlo:
Needs tests.

ksheedlo:
Update: added tests.

ksheedlo:
Update: added moar tests.

glyph:
LGTM; pretty straightforward, thanks for the softball :).

PR #408, opened 2015-09-11
fix(maas_api): use bytestrings for empty response

### LINKED ISSUE #407: Empty HTTP responses should always return byte strings
From https://github.com/rackerlabs/mimic/pull/384#discussion_r39227584. Also related to #392 as this will help ease the transition to py3.0.

ksheedlo:
Fixes #407.

Note: code search was done across Mimic, but only found the `return ''`
idiom inside the MaaS API, so fix is restricted to there. Of course,
this does not guarantee that a string of the wrong type won't be stuffed
into a variable and returned as an HTTP response.


lekhajee:
+1

PR #410, opened 2015-09-15
feat(MaasApi): filter overview results by entity ID

ksheedlo:
This change fixes an integration bug that has been worked around
upstream in Intelligence. In this case, Intelligence sometimes requests
a single entity's overview as a performance optimization, and the
backend is assumed to return overview data for exactly one entity. Mimic
did not do anything with the `entityId` query string parameter, and
incorrect results were returned. This change adds support for the
parameter in Mimic.

Usage of the overview call, including the `entityId` query string
parameter, is described in [section 5.18.1](http://goo.gl/c6dsYu)
of the Rackspace Cloud Monitoring Developer Guide.

### DIFF
--- a/mimic/rest/maas_api.py
+++ b/mimic/rest/maas_api.py
@@ -766,6 +766,17 @@ class MaasMock(object):
         serves the overview api call,returns all entities,checks and alarms
         """
         all_entities = self._entity_cache_for_tenant(tenant_id).entities_list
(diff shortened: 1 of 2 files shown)


lekhajee:
+1

lekhajee:
@ksheedlo : thank you for the many (small) PRs!

PR #412, opened 2015-09-15
refactor(MaasApi): clean and beautify code

ksheedlo:


ksheedlo:
Let's hold this for a bit. I have a more interesting MaaS refactor coming.

ksheedlo:
I do want to have some version of this change make it in, but please give priority to #418 in reviewing. I'd rather land that one first.

ksheedlo:
Ready for review.

glyph:
I definitely prefer the more functional style of value construction here, and it sets a better example for future code.  Thanks for taking the time to do this.

PR #415, opened 2015-09-15
refactor flavors to use a collection instead of a store on the region…

tawalton:
…al server collection  This PR addresses issue: https://github.com/rackerlabs/mimic/issues/416

### DIFF
--- a/mimic/model/flavor_objects.py
+++ b/mimic/model/flavor_objects.py
@@ -1,279 +1,149 @@
 """
-Model objects for mimic flavors.
+Model objects for the flavors.
 """
 
-from characteristic import attributes
+from characteristic import attributes, Attribute
+from json import dumps
+from mimic.model.flavors import (
+    RackspaceStandardFlavor, RackspaceComputeFlavor, RackspaceMemoryFlavor,
+    RackspaceOnMetalFlavor, RackspaceIOFlavor, RackspaceGeneralFlavor,
+    RackspacePerformance1Flavor, RackspacePerformance2Flavor)
+
+from twisted.web.http import NOT_FOUND
 
 
-@attributes(['flavor_id', 'tenant_id', 'name', 'ram', 'vcpus', 'rxtx', 'disk'])
-class Flavor(object):
+@attributes(['nova_message'])
+class BadRequestError(Exception):
     """
-    A Flavor object
+    Error to be raised when bad input has been received to Nova.
     """
 
-    static_defaults = {
-        "swap": "",
-        "OS-FLV-EXT-DATA:ephemeral": 0,
+
+def _nova_error_message(msg_type, message, status_code, request):
+    """
+    Set the response code on the request, and return a JSON blob representing
+    a Nova error body, in the format Nova returns error messages.
+
+    :param str msg_type: What type of error this is - something like
+        "badRequest" or "itemNotFound" for Nova.
+    :param str message: The message to include in the body.
+    :param int status_code: The status code to set
+    :param request: the request to set the status code on
+
+    :return: dictionary representing the error body
+    """
+    request.setResponseCode(status_code)
+    return {
+        msg_type: {
+            "message": message,
+            "code": status_code
+        }
     }
 
-    def links_json(self, absolutize_url):
-        """
-        Create a JSON-serializable data structure describing the links to this
-        flavor.
-        """
-        return [
-            {
-                "href": absolutize_url("v2/{0}/flavors/{1}"
-                                       .format(self.tenant_id, self.flavor_id)),
-                "rel": "self"
-            },
-            {
-                "href": absolutize_url("{0}/flavors/{1}"
-                                       .format(self.tenant_id, self.flavor_id)),
-                "rel": "bookmark"
-            }
-        ]
 
-    def extra_specs_json(self):
-        """
-        Create a JSON-serializable data structure describing
-        ``OS-FLV-WITH-EXT-SPECS:extra_specs`` for a flavor.
-        """
-        return {
-            "quota_resources": "instances=mimic-instances,ram=mimic-ram",
-        }
-
-    def brief_json(self, absolutize_url):
-        """
-        Brief JSON-serializable version of this flavor, for the non-details
-        list flavors request.
-        """
-        return {
-            "id": self.flavor_id,
-            "links": self.links_json(absolutize_url),
-            "name": self.name
-        }
-
-    def detailed_json(self, absolutize_url):
-        """
-        Long-form JSON-serializable object representation of this flavor, as
-        returned by either a GET on this individual flavor or a member in the
-        list returned by the list-details request.
-        """
-        template = self.static_defaults.copy()
-        template.update({
-            "id": self.flavor_id,
-            "links": self.links_json(absolutize_url),
-            "name": self.name,
-            "ram": self.ram,
-            "vcpus": self.vcpus,
-            "rxtx_factor": self.rxtx,
-            "disk": self.disk,
-            "OS-FLV-WITH-EXT-SPECS:extra_specs": self.extra_specs_json()
-        })
-        return template
-
-
-class RackspaceStandardFlavor(Flavor):
+def not_found(message, request):
     """
-    A Rackspace standard flavor object representation
+    Return a 404 error body associated with a Nova not found error.
+    Also sets the response code on the request.
+
+    :param str message: The message to include in the bad request body.
+    :param request: The request on which to set the response code.
+
+    :return: dictionary representing the error body.
     """
-    flavors = {"512MB Standard Instance": {"id": "2", "ram": 512, "vcpus": 1, "rxtx_factor": 80,
-                                           "disk": 20},
-               "1GB Standard Instance": {"id": "3", "ram": 1024, "vcpus": 1, "rxtx_factor": 120,
-                                         "disk": 40},
-               "2GB Standard Instance": {"id": "4", "ram": 2048, "vcpus": 2, "rxtx_factor": 240,
-                                         "disk": 80},
-               "4GB Standard Instance": {"id": "5", "ram": 4096, "vcpus": 2, "rxtx_factor": 400,
-                                         "disk": 160},
-               "8GB Standard Instance": {"id": "6", "ram": 8192, "vcpus": 4, "rxtx_factor": 600,
-                                         "disk": 320},
-               "15GB Standard Instance": {"id": "7", "ram": 15360, "vcpus": 6, "rxtx_factor": 800,
-                                          "disk": 620},
-               "30GB Standard Instance": {"id": "8", "ram": 30720, "vcpus": 8, "rxtx_factor": 1200,
-                                          "disk": 1200}}
-
-    def extra_specs_json(self):
-        """
-        Create a JSON-serializable data structure describing
-        ``OS-FLV-WITH-EXT-SPECS:extra_specs`` for a standard flavor.
-        """
-        return {
-            "class": "standard1",
-            "policy_class": "standard_flavor"
-        }
+    return _nova_error_message("itemNotFound", message, NOT_FOUND, request)
 
 
-class RackspaceComputeFlavor(Flavor):
+@attributes(
+    ["tenant_id", "region_name", "clock",
+     Attribute("flavors_store", default_factory=list)]
+)
+class RegionalFlavorCollection(object):
     """
-    A Rackspace compute flavor object representatione
+    A collection of flavors, in a given region, for a given tenant.
     """
-    flavors = {"3.75 GB Compute v1": {"id": "compute1-4", "ram": 3840, "vcpus": 2, "rxtx_factor": 312.5,
-                                      "disk": 0},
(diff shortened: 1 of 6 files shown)


gabecase:
ServerKeypairs story needs this as well(using same pattern of collections). I have reviewed this pr and it LGTM but will help to get some +1's from mimic core team as well since its a significant improvement.

ksheedlo:
I want to review this, but don't have enough expertise in Nova to feel comfortable reviewing a PR this large. @glyph @lekhajee can you take this on or suggest an owner who can?

tawalton:
@ksheedlo This PR is not as daunting as it first appears.  The flavor classes and flavor routes are the same as the master branch. This is just a refactor to introduce a RegionalFlavorCollection for flavors instead of having a flavor_store attribute on the RegionalServerCollection. See issue https://github.com/rackerlabs/mimic/issues/416

mimic/model/flavor_objects.py
```diff
@@ -1,279 +1,107 @@
...
+Model objects for the flavors.
 """
 
-from characteristic import attributes
+from characteristic import attributes, Attribute
+from json import dumps
+from mimic.model.flavors import (
+    RackspaceStandardFlavor, RackspaceComputeFlavor, RackspaceMemoryFlavor,
+    RackspaceOnMetalFlavor, RackspaceIOFlavor, RackspaceGeneralFlavor,
+    RackspacePerformance1Flavor, RackspacePerformance2Flavor)
```
ksheedlo: I find it awkward that this module is still named `flavor_objects` and now there's a new module called `flavors`. Perhaps this module could be renamed to `flavor_collections`?

mimic/model/flavor_objects.py
```diff
@@ -1,279 +1,107 @@
...
-               "30 GB Memory v1": {"id": "memory1-30", "ram": 30720, "vcpus": 4, "rxtx_factor": 1250,
-                                   "disk": 0},
-               "60 GB Memory v1": {"id": "memory1-60", "ram": 61440, "vcpus": 8, "rxtx_factor": 2500,
-                                   "disk": 0},
-               "120 GB Memory v1": {"id": "memory1-120", "ram": 122880, "vcpus": 16, "rxtx_factor": 5000,
-                                    "disk": 0},
-               "240 GB Memory v1": {"id": "memory1-240", "ram": 245760, "vcpus": 32,
-                                    "rxtx_factor": 10000, "disk": 0}}
+        if len(self.flavors_store) < 1:
+            self.create_flavors_list()
```
ksheedlo: non-blocking: I see this needs to be lazy loaded for some reason, and I see looking down at `create_flavors_list` it also depends on the tenant ID. Really what would be nice here is to have the attribute initializer set all this up, but `characteristic` isn't sophisticated enough to do that automatically. I wonder if `attrs` could configure it in such a way?

mimic/model/flavor_objects.py
```diff
@@ -1,279 +1,107 @@
...
+        for flavor in self.flavors_store:
+            if flavor.flavor_id == flavor_id:
+                return flavor
 
-    def extra_specs_json(self):
+    def create_flavors_list(self):
```
ksheedlo: This method should probably be private.

mimic/model/flavor_objects.py
```diff
@@ -1,279 +1,107 @@
...
-        return {
-            "class": "memory1",
-            "policy_class": "memory_flavor"
-        }
+        rackspace_flavors = [RackspaceStandardFlavor, RackspaceComputeFlavor,
+                             RackspacePerformance1Flavor, RackspaceOnMetalFlavor,
+                             RackspacePerformance2Flavor, RackspaceMemoryFlavor,
+                             RackspaceIOFlavor, RackspaceGeneralFlavor]
 
+        if len(self.flavors_store) < 1:
```
ksheedlo: If you check `len(self.flavors_store) < 1` in this function's callers, then you don't need to check it here. Or you could check it here, drop the other checks and call it `_create_flavors_list_if_empty` or similar. I feel checking it in both places is unnecessary repetition.

mimic/model/flavor_objects.py
```diff
@@ -1,279 +1,107 @@
...
+                    flavor_name = flavor
+                    flavor_id = flavor_spec['id']
+                    ram = flavor_spec['ram']
+                    vcpus = flavor_spec['vcpus']
+                    network = flavor_spec['rxtx_factor']
+                    disk = flavor_spec['disk']
+                    tenant_id = self.tenant_id
+                    flavor = flavor_class(flavor_id=flavor_id, tenant_id=tenant_id,
+                                          name=flavor_name, ram=ram, vcpus=vcpus,
+                                          rxtx=network, disk=disk)
```
ksheedlo: non-blocking: naming everything twice just to shove it into keyword args seems a bit repetitive. I would have had no idea what `rxtx_factor` is, but you have to turn around and call it `rxtx` again.

Anyway, I'm down with it - no need to change this if you don't want.

mimic/model/flavor_objects.py
```diff
@@ -1,279 +1,107 @@
...
-            "quota_resources": "instances=onmetal-compute-v1-instances,ram=onmetal-compute-v1-ram",
-            "class": "onmetal",
-            "policy_class": "onmetal_flavor"
-        }
+        if len(self.flavors_store) < 1:
+            self.create_flavors_list()
+        flavors = []
+        for flavor in self.flavors_store:
+            if self.region_name != "IAD" and isinstance(flavor, RackspaceOnMetalFlavor):
+                continue
```
ksheedlo: non-blocking, since I see this was here before: This seems like a piece of business logic awkwardly wrapped up in a container with very little business logic around it. I wonder if it could be handled more elegantly somewhere else.

mimic/model/flavor_objects.py
```diff
@@ -1,279 +1,107 @@
...
-
-    def extra_specs_json(self):
+    def get_flavor(self, http_get_request, flavor_id, absolutize_url):
         """
-        Create a JSON-serializable data structure describing
-        ``OS-FLV-WITH-EXT-SPECS:extra_specs`` for a performance1 flavor.
+        Return a flavor object if one exists from the list `/flavors` api,
+        else creates and adds the flavor to the :obj: `flavors_store`.
+        If the `flavor_id` is listed in `mimic.canned_responses.mimic_presets`,
+        then will return 404.
```
ksheedlo: Is this comment outdated and/or a typo? Looks like it returns 404 if the `flavor_id` is *not* listed in the `flavors_store`, where flavor types are defined in `mimic.model.flavors`.

mimic/model/flavor_objects.py
```diff
@@ -1,279 +1,107 @@
...
-               "120 GB Performance": {"id": "performance2-120", "ram": 122880, "vcpus": 32,
-                                      "rxtx_factor": 10000, "disk": 40}}
 
-    def extra_specs_json(self):
+    def collection_for_region(self, region_name):
         """
-        Create a JSON-serializable data structure describing
-        ``OS-FLV-WITH-EXT-SPECS:extra_specs`` for a performance 2 flavor.
+        Get a :obj:`RegionalFlavorCollection` for the region identified by the
+        given name.
```
ksheedlo: Does anything prevent you from passing an nonexistent `region_name` to this function? Presumably if it doesn't recognize the name it's going to create another `RegionalFlavorCollection`.

Forgive me if this has an obvious answer - I mainly work on MaaS which operates out of one region.

mimic/model/flavors.py
```diff
@@ -0,0 +1,270 @@
...
+
+    def extra_specs_json(self):
+        """
+        Create a JSON-serializable data structure describing
+        ``OS-FLV-WITH-EXT-SPECS:extra_specs`` for a performance1 flavor.
+        """
+        return {
+            "class": "performance1",
+            "policy_class": "performance_flavor"
+        }
```
ksheedlo: :100: this is all straightforward

mimic/model/flavor_objects.py
```diff
@@ -1,279 +1,107 @@
...
         """
-        return {
-            "class": "performance1",
-            "policy_class": "performance_flavor"
-        }
+        flavor = self.flavor_by_id(flavor_id)
+        if flavor is None:
+            return dumps(not_found("The resource could not be found.",
+                                   http_get_request))
+        return dumps({"flavor": flavor.detailed_json(absolutize_url)})
```
ksheedlo: Otherwise this looks a lot nicer than the old `get_flavor` function. :sparkles:

mimic/rest/nova_api.py
```diff
@@ -328,10 +328,10 @@ def get_flavor_details(self, request, tenant_id, flavor_id):
         """
         Returns a get flavor response, for any given flavorid
         """
-        return (
-            self._region_collection_for_tenant(tenant_id)
-            .get_flavor(request, flavor_id, absolutize_url=self.url)
-        )
+        flavor_collection = GlobalFlavorCollections(tenant_id=tenant_id,
+                                                    clock=self._session_store.clock)
```
ksheedlo: My gut feel here is that the class should be called `GlobalFlavorCollection` (strike the "s" at the end). Reading this as "flavor_collection is a new global flavor collections" and it feels off.

mimic/rest/nova_api.py
```diff
@@ -353,10 +353,10 @@ def get_flavor_list_with_details(self, request, tenant_id):
...
                Also be able to set different flavor types in the future.
         """
-        return (
-            self._region_collection_for_tenant(tenant_id)
-            .list_flavors(include_details=True, absolutize_url=self.url)
-        )
+        flavor_collection = GlobalFlavorCollections(tenant_id=tenant_id,
+                                                    clock=self._session_store.clock)
+        return (flavor_collection.collection_for_region(region_name=self._name)
+                .list_flavors(include_details=True, absolutize_url=self.url))
```
ksheedlo: otherwise these APIs read really clean, good job!

mimic/test/test_flavors.py
```diff
@@ -33,25 +33,23 @@ def get_server_flavor(self, postfix):
...
+        self.assertEqual(get_server_flavor['flavor']['id'], '2')
+
     def test_get_server_flavor_negative(self):
         """
         Test to verify :func:`get_flavor` when invalid flavor from the
         :obj: `mimic_presets` is provided.
         """
         get_server_flavor = request(self, self.root, "GET", self.uri +
-                                    '/flavors/1')
+                                    '/flavors/100')
```
ksheedlo: I don't understand this. Why is `2` a valid flavor ID and not `100`?

mimic/test/test_flavors.py
```diff
@@ -139,3 +137,12 @@ def test_get_flavor_list_after_list_flavors(self):
             get_server_flavor_response_body = self.get_server_flavor('/flavors/' + flavor_id)
             self.assertEqual(flavor_id, get_server_flavor_response_body['flavor']['id'])
             self.assertEqual(name, get_server_flavor_response_body['flavor']['name'])
+
+    def test_get_OnMetal_flavor(self):
```
ksheedlo: I'd prefer if this was a pure refactor that's able to maintain coverage without adding any new tests. I'll let it slide this time, though.

mimic/test/test_flavors.py
ksheedlo: OK, I see. Plain number IDs are used by standard flavors, but the ID can be any string. Can we use an ID that's more obviously not going to be found, such as `negative-test-1` or similar?

mimic/model/flavor_objects.py
ksheedlo: OK, looks like this comes in from the `NovaRegion` object and it would be difficult or impossible to set the name of this to whatever you want, so this code is chill.

ksheedlo:
### Actionable recommendations on this PR

1. Rename `mimic.model.flavor_objects` to `mimic.model.flavor_collections`
2. Make `RegionalFlavorCollection#create_flavors_list` private
3. DRY up if-guards for flavors store - https://github.com/rackerlabs/mimic/pull/415/files#r39714100
4. revise docstring for `RegionalFlavorCollection#get_flavor`
5. rename `GlobalFlavorCollections` to `GlobalFlavorCollection`
6. check for me - do we need the new OnMetal test case for coverage? Otherwise I'd prefer if it was filed as a separate PR; this PR would look better if it was possible without adding new tests.

tawalton:
Fixed recommendations on this PR

- [ ] Rename mimic.model.flavor_objects to mimic.model.flavor_collections
- [ ] Make RegionalFlavorCollection#create_flavors_list private
- [ ] DRY up if-guards for flavors store - https://github.com/rackerlabs/mimic/pull/415/files#r39714100
- [ ] revise docstring for RegionalFlavorCollection#get_flavor
- [ ] rename GlobalFlavorCollections to GlobalFlavorCollection
- [ ] check for me - do we need the new OnMetal test case for coverage? Otherwise I'd prefer if it was - [ filed as a separate PR; this PR would look better if it was possible without adding new tests.

ksheedlo:
Nice work! LGTM

PR #418, opened 2015-09-16
refactor(maas): use class models

ksheedlo:


ksheedlo:
Ready for review.

ksheedlo:
@glyph or @lekhajee can you look at this PR today? I want to hack on alarm states today, but I want this to get in first.

glyph:
Thanks @ksheedlo .

Looks like a straight refactoring: no direct tests, but all the code is hit by existing coverage, lint et. al. pass, so I will land.

PR #420, opened 2015-09-18
feat(maas_api): add support for alarm states

ksheedlo:
This change adds support for alarm states in the MaaS plugin. Since
Mimic can't parse and execute alarms, alarm states are created using the
control API for MaaS. They can be recalled using the
`/views/latest_alarm_states` API or the `/views/overview` API in the
normal MaaS routes.

### DIFF
--- a/mimic/rest/maas_api.py
+++ b/mimic/rest/maas_api.py
@@ -26,6 +26,7 @@ from mimic.canned_responses.maas_agent_info import agent_info
 from mimic.canned_responses.maas_monitoring_zones import monitoring_zones
 from mimic.canned_responses.maas_alarm_examples import alarm_examples
 from mimic.model.maas_objects import (Alarm,
+                                      AlarmState,
                                       Check,
                                       Entity,
                                       MaasStore,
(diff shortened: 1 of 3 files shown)


ksheedlo:
Update: There's some business logic around alarm state APIs that I still need to account for.

ksheedlo:
Update: Fixed and ready for review.

The bug was that in the first hack at this change, all alarm states were returned when requesting latest alarm states. "Latest" in this context means that it is the most recent alarm state for each entity for each alarm.

lekhajee:
@ksheedlo : needs rebase

ksheedlo:
Don't let me forget to harden this against bad keys like in #422.

ksheedlo:
Update: Added hardening. Missing `state` or `status` keys to the create-alarm-state API will return 400.

mimic/rest/maas_api.py
```diff
@@ -1299,6 +1304,31 @@ def multiplot(self, request, tenant_id):
         self._audit('rollups', request, tenant_id, status, content)
         return json.dumps({'metrics': metrics_replydata})
 
+    @app.route('/v1.0/<string:tenant_id>/views/latest_alarm_states', methods=['GET'])
+    def latest_alarm_states(self, request, tenant_id):
+        """
+        Gets entities grouped with their latest alarm states.
```
lekhajee: Can we add a link to the docs for this api, if one exists.

Same for the one below.

lekhajee:
+1 modulo the nit.

mimic/rest/maas_api.py
ksheedlo: There are docs for the `/views/overview` API, but not this one. The API for creating alarm states is part of the Mimic control plane. If you have a preferred way of writing up the APIs that we create for Mimic, let me know and I'll send a follow-up PR.

PR #421, opened 2015-09-19
rackspace images

tawalton:


lekhajee:
@tawalton : missing labels. Is this ready for review

tawalton:
@lekhajee this is a work in progress. We ran into a blocker that we are trying to work out.  We can't get data_for_api function in session.py to return the correct collection object.

tawalton:
These changes have been implemented in https://github.com/rackerlabs/mimic/pull/434

PR #422, opened 2015-09-20
fix(maas_api): drop unrecognized keys when creating

ksheedlo:
If keys that are not part of the MaaS business domain are POSTed to the
MaaS endpoints, a 500 would have been caused. This change fixes the bug.
In real MaaS, the extra properties would be dropped, so Mimic does the
same thing now.

### DIFF
--- a/mimic/model/maas_objects.py
+++ b/mimic/model/maas_objects.py
@@ -130,6 +130,7 @@ class Alarm(object):
                 'label': self.label,
                 'criteria': self.criteria,
                 'check_id': self.check_id,
+                'entity_id': self.entity_id,
                 'notification_plan_id': self.notification_plan_id,
                 'created_at': self.created_at,
                 'updated_at': self.updated_at,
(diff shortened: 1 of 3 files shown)


mimic/model/maas_objects.py
```diff
@@ -130,6 +130,7 @@ def to_json(self):
                 'label': self.label,
                 'criteria': self.criteria,
                 'check_id': self.check_id,
+                'entity_id': self.entity_id,
```
ksheedlo: This is a bit of a rider, because it turns out Intelligence tests are still broken without it.

mimic/rest/maas_api.py
```diff
@@ -154,7 +161,12 @@ def create_entity(clock, params):
         ``bool``, ``dict`` or ``NoneType``.
     """
     current_time_milliseconds = int(1000 * clock.seconds())
-    params_copy = dict(params)
+    params_copy = _only_keys(params, ['agent_id',
```
lekhajee: Non blocker: I feel like the keys list for each object should be in a different file, so one could see all the keys supported for each of these calls. I know from using the ele docs that this information was hard to find and having this here makes it lost in the midst of all this code.

lekhajee:
+1. Can merge when tests pass modulo the suggestion which can be fixed in a separate PR.

PR #423, opened 2015-09-20
fix(maas_api): return 400 for missing properties

ksheedlo:
If the user tries to create a Check without a `type` or an Alarm without
a `check_id` or a `notification_plan_id`, Mimic threw a 500. This change
fixes Mimic to return 400 with a properly formatted JSON response.


lekhajee:
+1

PR #424, opened 2015-09-21
refactor(maas_api): list spec'able keys for types

ksheedlo:
Addresses https://github.com/rackerlabs/mimic/pull/422#discussion_r40003112.

Each MaaS business object has a set of keys that can be set by the API
user. This change formalizes the concept and uses it when filtering keys
during updates, as well as when creating new instances of each type.


lekhajee:
+1

PR #425, opened 2015-09-22
feat(maas_api): list metrics for all check types

ksheedlo:
This change expands the list of metrics Mimic returns from just
`available` and `average` on `remote.ping` to all known metrics on all
known check types. If a remote check has a `monitoring_zones_poll`, all
metrics are returned from each monitoring zone. If a check has a type
that Mimic doesn't know about, it returns no metrics in the metrics list
call, as it did before.


lekhajee:
+1

PR #431, opened 2015-09-23
refactor(lbaas): move store methods to store

### LINKED ISSUE #430: Move mimic.canned_responses.loadbalancer. _verify_and_update_lb_state
This component appears at first glance to be a private function in the `mimic.canned_responses.loadbalancer` module. In fact, it's used as a private method of the `mimic.model.clb_objects.RegionalCLBCollection` object. It needs to be moved.

ksheedlo:
The methods `_verify_and_update_lb_state` and `_delete_node` are
actually private methods of the
`mimic.model.clb_objects.RegionalCLBCollection` class. They were defined
in the `mimic.canned_responses.loadbalancer` module, which caused
confusion. This change moves them into the class where they belong.

Fixes #430.

### DIFF
--- a/mimic/canned_responses/loadbalancer.py
+++ b/mimic/canned_responses/loadbalancer.py
@@ -4,8 +4,6 @@ Canned response for add/get/list/delete load balancers and
 add/get/delete/list nodes
 """
 from random import randrange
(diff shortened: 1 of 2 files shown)


lekhajee:
+1, thanks much :)

PR #432, opened 2015-09-23
feat(maas_api): use mimic metrics in multiplot

ksheedlo:
This PR ditches the old way of doing multiplot, which was to look for a ping check and spit out random metrics unless the name of the check was "squarewave", in which case it did other things.

With this change, multiplot consults the collection of MaaS metrics in the `maas_store` to see if anyone has specified an override for the requested metric. If so, the override (specified as a function taking a timestamp) generates metrics over the interval; otherwise, random values are generated appropriately for the metric type. The net of this is that anyone who uses MaaS and doesn't use `squarewave`* will get the same metrics they're used to, but now with additional types. Additionally, it will make it easy to rebuild `squarewave` using metric overrides and the control API.

*I believe nobody actually uses this. Toks wrote it, but Intelligence doesn't use it.

### DIFF
--- a/mimic/model/maas_objects.py
+++ b/mimic/model/maas_objects.py
@@ -388,21 +388,25 @@ class Metric(object):
             return random_string(random.randint(12, 30), selectable=(string.letters + string.digits))
         raise ValueError('No default data getter for type {0}!'.format(self.type))
 
+    def get_value(self, **kwargs):
+        """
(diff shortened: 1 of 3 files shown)


lekhajee:
@ksheedlo : needs to be rebased

lekhajee:
+1

Issue #416, opened 2015-09-15
Global and regional collections for nova


tawalton:
Currently the flavor_store and images_store live as an attribute on the Regional Server Collection, https://github.com/rackerlabs/mimic/blob/master/mimic/model/nova_objects.py#L655-#L656. The flavors and images should live on the session and have their own respective collections since they are not specific to the server collection. 

![image](https://cloud.githubusercontent.com/assets/5943782/9889872/bf4188dc-5bca-11e5-85b7-f4b80819bf9d.png)

Issue #414, opened 2015-09-15
Refactor nova tests


tawalton:
There are over 2300 lines of test code in https://raw.githubusercontent.com/rackerlabs/mimic/master/mimic/test/test_nova.py . This file needs to be refactored into multiple files maintaining code coverage.

PR #434, opened 2015-09-23
Support rackspace images

tawalton:
This PR addresses issue https://github.com/rackerlabs/mimic/issues/416


mimic/model/nova_objects.py
```diff
@@ -933,67 +931,6 @@ def request_action(self, http_action_request, server_id, absolutize_url):
...
-    def get_image(self, http_get_request, image_id, absolutize_url):
-        """
-        Return a image object if one exists from the list `/images` api,
-        else creates and adds the image to the :obj: `images_store`.
-        If the `image_id` is listed in `mimic.canned_responses.mimic_presets`,
-        then will return 404.
-        """
-        if (
-            image_id in get_presets['servers']['invalid_image_ref'] or
-            image_id.endswith('Z')
```
lekhajee: The image starts with Z has been removed above. This will cause tests to fail for our consumers.

mimic/test/test_nova.py
```diff
@@ -810,15 +749,86 @@ def test_reboot_server(self):
             treq.json_content(rebooted_server_response))
         self.assertEqual(rebooted_server_response_body['server']['status'], 'ACTIVE')
 
-    def test_change_password(self):
+    def test_rescue(self):
         """
```
lekhajee: There seem to be unrelated test changes in this file.

mimic/test/test_server_images.py
```diff
@@ -87,9 +89,9 @@ def test_get_image_list_with_details(self):
         for each_image in image_list:
             self.assertEqual(
                 sorted(each_image.keys()),
-                sorted(['id', 'name', 'links', 'status', 'progress', 'minRam',
-                        'OS-DCF:diskConfig', 'OS-EXT-IMG-SIZE:size', 'metadata',
-                        'minDisk', 'created', 'updated']))
+                sorted(['id', 'name', 'links', 'minRam',
```
lekhajee: Change this test class to load IAD instead of ORD so that this validation applies to onmetal images as well.

Also, just adding IAD to the list of regions will not help cause currently the APIMockHelper just fetched the endpoint for the first region :/

mimic/test/test_nova.py
tawalton: @lekhajee this is not a test change. It looks like a change in spacing / placement of tests in this file. I  will see about moving them back to original place

mimic/test/test_server_images.py
```diff
@@ -112,3 +114,34 @@ def test_get_image_list_after_list_images(self):
                              get_server_image_response_body['image']['id'])
             self.assertEqual(each_image['name'],
                              get_server_image_response_body['image']['name'])
+
+    def test_get_OnMetal_server_image(self):
+        """
+        Test to verify :func:`get_image` on ``GET /v2.0/<tenant_id>/images/<image_id>``
+        """
+        helper = APIMockHelper(self, [NovaApi(['IAD'])])
```
lekhajee: Replacing ORD with IAD will help skip this step.

mimic/test/test_server_images.py
tawalton: The onMetal images test that cover this case are here: https://github.com/rackerlabs/mimic/blob/support_rackspace_images/mimic/test/test_images.py

mimic/test/test_server_images.py
```diff
@@ -112,3 +114,34 @@ def test_get_image_list_after_list_images(self):
...
+                onmetal_id = image['id']
+                break
+        response, body = self.successResultOf(json_request(
+            self, root, "GET", uri + '/images/' + onmetal_id))
+        self.assertEqual(200, response.code)
+        self.assertEqual(body['image']['id'], onmetal_id)
+        self.assertEqual(body['image']['metadata']['flavor_classes'], 'onmetal')
+
+    def test_OnMetal_image_list(self):
+        helper = APIMockHelper(self, [NovaApi(['IAD'])])
```
lekhajee: How about adding a negative test to verify that ORD does not have onmetal images.

mimic/test/test_images.py
```diff
@@ -0,0 +1,113 @@
+"""
+Tests for :mod:`nova_api` and :mod:`nova_objects` for images.
+"""
```
lekhajee: Why is this in a separate file. We could merge this with the existing test_server_images.py file as this seems redundant.

Also, as mentioned below, if the idea to validate IAD and non IAD regions, lets have separate test classes with the same file.

tawalton: @lekhajee Can we just pull all the nova images tests into this file and remove them from test_nova.py since it is really big and it will start to address issue https://github.com/rackerlabs/mimic/issues/414

mimic/test/test_server_images.py
tawalton: Great idea.

mimic/test/test_images.py
lekhajee: sure, sounds good. I thought I had done that already, but if I had left some, please do so.

Also, lets please merge test_images and test_server_images.

lekhajee:
+1

PR #436, opened 2015-09-24
SSH Keypair Routes

gabecase:
Now able to create, read and delete SSH Keys for the NovaApi:
http://docs.rackspace.com/servers/api/v2/cs-devguide/content/ServersKeyPairs-d1e2545.html

Verified it works in reach here:
reach

### DIFF
--- a/mimic/model/keypair_objects.py
+++ b/mimic/model/keypair_objects.py
@@ -74,6 +74,9 @@ class RegionalKeyPairCollection(object):
         Remove a :obj:`KeyPair` from the list of keypairs
         """
         kp_to_remove = self.keypair_by_name(name)
+        if kp_to_remove is None:
+            raise ValueError("Keypair Not Found")
+
         self.keypairs.remove(kp_to_remove)
 
 
--- a/mimic/model/keypair_objects.py
+++ b/mimic/model/keypair_objects.py
@@ -34,19 +34,19 @@ class KeyPair(object):
 class RegionalKeyPairCollection(object):
     """
     A :obj:`ReionalKeyPairCollection` is a collection of
-    :obj:`KeyPair` objects owned by a given tenant for a region.
+    :obj:`KeyPair` objects owned by a given tenant for a  region.
     """
 
     def create_keypair(self, keypair):
         """
-        Add a :obj:`KeyPair` to the list of keypairs
+        Add a :obj:`KeyPair` to the list of  keypairs
         """
         self.keypairs.append(keypair)
         return self.keypair_by_name(keypair.name).key_json()
 
     def keypair_by_name(self, name):
         """
-        Return a :obj:`KeyPair` by name from the current keypairs list
+        Return a :obj:`KeyPair` by name from the current keypairs  list
         """
         for keypair in self.keypairs:
             if keypair.name == name:
@@ -54,7 +54,7 @@ class RegionalKeyPairCollection(object):
 
     def json_list(self):
         """
-        JSON List of all :obj:`KeyPair` for a region
+        JSON List of all :obj:`KeyPair` for a  region
         """
         result = {"keypairs": []}
         if len(self.keypairs) > 0:
@@ -71,7 +71,7 @@ class RegionalKeyPairCollection(object):
 
     def remove_keypair(self, name):
         """
-        Remove a :obj:`KeyPair` from the list of keypairs
+        Remove a :obj:`KeyPair` from the list of  keypairs
         """
         kp_to_remove = self.keypair_by_name(name)
         self.keypairs.remove(kp_to_remove)
--- a/mimic/model/keypair_objects.py
+++ b/mimic/model/keypair_objects.py
@@ -1,0 +1,100 @@
+"""
+Keypair objects for mimic
+"""
+
+import json
+
+from characteristic import attributes, Attribute
+
+
+@attributes(['name', 'public_key'])
+class KeyPair(object):
+    """
+    A KeyPair object
+    """
+    fingerprint = "aa:aa:aa:aa:aa:aa:aa:aa:aa:aa:aa:aa:aa:aa:aa:aa"
+
+    def key_json(self):
+        """
+        Serialize a :obj:`Keypair` into JSON
+        """
+        return {
+            "keypair": {
+                "name": self.name,
+                "public_key": self.public_key,
+                "fingerprint": self.fingerprint
+            }
+        }
+
+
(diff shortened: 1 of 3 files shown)


tawalton:
This PR addresses issue https://github.com/rackerlabs/mimic/issues/416

lekhajee:
Thank you for the screenshot!

mimic/rest/nova_api.py
```diff
@@ -216,6 +217,16 @@ def _region_collection_for_tenant(self, tenant_id):
         return (self._api_mock._get_session(self._session_store, tenant_id)
                 .collection_for_region(self._name))
 
+    def _keypair_collection_for_tenant(self, tenant_id):
```
lekhajee: missing docstring

mimic/rest/nova_api.py
```diff
@@ -399,6 +410,40 @@ def perform_action(self, request, tenant_id, server_id):
...
+        """
+        try:
+            content = json.loads(request.content.read())
+        except ValueError:
+            request.setResponseCode(400)
+            return json.dumps(bad_request("Malformed request body", request))
+
+        keypair = content["keypair"]
+        keypair_from_request = KeyPair(
+            name=keypair["name"], public_key=keypair["public_key"])
```
lekhajee: This will cause an exception if the `name` or `public_key` keys do not exist. Can we please handle this.

mimic/rest/nova_api.py
```diff
@@ -399,6 +410,40 @@ def perform_action(self, request, tenant_id, server_id):
...
+        keypair_response = self._keypair_collection_for_tenant(
+            tenant_id).create_keypair(keypair=keypair_from_request)
+        return json.dumps(keypair_response)
+
+    @app.route("/v2/<string:tenant_id>/os-keypairs/<string:keypairname>", methods=['DELETE'])
+    def delete_key_pair(self, request, tenant_id, keypairname):
+        """
+        Removes a key by its name
+        """
+        self._keypair_collection_for_tenant(
```
lekhajee: Please handle the case of 404 when the `name` does not exist.

mimic/test/test_keypairs.py
```diff
@@ -0,0 +1,87 @@
...
+
+    def test_delete_keypair(self):
+        resp = self.successResultOf(request(
+            self, self.helper.root, "DELETE", self.helper.uri +
+            '/os-keypairs/' + self.keypair_name
+        ))
+
+        self.assertEqual(resp.code, 202)
+        resp, body = self.get_keypairs_list()
+        self.assertTrue(len(body['keypairs']) < 2)
```
lekhajee: Instead of the length, can we verify that the deleted `key_name` no longer exists in the list.

mimic/model/keypair_objects.py
```diff
@@ -0,0 +1,100 @@
...
+        if len(self.keypairs) > 0:
+            keypairs_json = {}
+            for keypair in self.keypairs:
+                keypairs_json.update(keypair.key_json())
+            result = {
+                "keypairs": [
+                    keypairs_json
+                ]
+            }
+
```
lekhajee: you could replace the `for ` loop with `result["keypairs"] = [keypairs_json.update(keypair.key_json()) for keypair in self.keypairs]`

mimic/rest/nova_api.py
gabecase: I actually had this change on a different branch, updating to include that now

mimic/model/keypair_objects.py
gabecase: When I added the above code it started breaking. It would always return:
`{'keypairs': []}` even with keypairs in the collection.

Is it alright if I leave this loop as is? Reach is expecting a certain nested json structure from the api to be returned and I was able to achieve it with this loop.

mimic/rest/nova_api.py
```diff
@@ -216,6 +217,19 @@ def _region_collection_for_tenant(self, tenant_id):
         return (self._api_mock._get_session(self._session_store, tenant_id)
                 .collection_for_region(self._name))
 
+    def _keypair_collection_for_tenant(self, tenant_id):
+        """
```
gabecase: This is the doc string correct?

mimic/rest/nova_api.py
gabecase: I believe the doc string is below

mimic/rest/nova_api.py
```diff
@@ -399,6 +413,44 @@ def perform_action(self, request, tenant_id, server_id):
...
+
+    @app.route("/v2/<string:tenant_id>/os-keypairs", methods=['POST'])
+    def create_key_pair(self, request, tenant_id):
+        """
+        Returns a newly created key pair with the specified name.
+        """
+        try:
+            content = json.loads(request.content.read())
+            keypair = content["keypair"]
+            keypair_from_request = KeyPair(name=keypair["name"], public_key=keypair["public_key"])
```
lekhajee: Does not look like `public_key` is a required fields for this API call from looking at http://docs.rackspace.com/servers/api/v2/cs-devguide/content/CreateKeyPair.html. Also, looks like we are missing the attributes `user_id` and `private_key` in the response object.

Also, could you please add the links to the docs in each of the docstrings of the APIs.

gabecase: That route isn't being covered in this PR since its a rare use case(asking for the API to generate a pub/private key pair for you) since most of the time end users are uploading there own public key.

The route I'm mocking is here:
http://docs.rackspace.com/servers/api/v2/cs-devguide/content/UploadKeyPair.html

The api generating  a keypair for you route was out of the scope of this pull-request and doesn't have a use case for mimic yet so it hasn't been included yet.

I will update this pull-request to include a fake `user_id` though.

gabecase: I would like to add the following missing route:
http://docs.rackspace.com/servers/api/v2/cs-devguide/content/CreateKeyPair.html

In a separate PR if that is alright.

lekhajee:
@gabecase : Please fix conflicts

mimic/rest/nova_api.py
lekhajee: Ok, sounds good. Can you please add a issue for this, if you do not intend on working on this immediately

lekhajee:
+1. Feel free to merge this one the tests pass.

gabecase:
Why is this failing patch now, I didnt add any code

PR #438, opened 2015-09-25
feat(maas_api): deleting missing object throws 404

ksheedlo:


mimic/rest/maas_api.py
```diff
@@ -518,9 +518,21 @@ def delete_entity(self, request, tenant_id, entity_id):
         Delete an entity, all checks that belong to entity, all alarms that belong to those checks
         """
         entities = self._entity_cache_for_tenant(tenant_id).entities_list
+        previous_size = len(entities)
         entities[:] = [entity for entity in entities
                        if entity.id != entity_id]
 
+        if len(entities) == previous_size:
```
lekhajee: This seems non intuitive. Why compare the lengths of the list instead of trying to find the item in `self._entity_cache_for_tenant(tenant_id).entities_list`.

ksheedlo: Intuitively, if an entity was deleted the size of the list must have changed.

The convention for deletes in this plugin is to write a list comprehension for the new contents of the list and assign it to a slice of itself, like you see above. This ensures we don't iterate over a list while modifying it, which complicates reasoning and causes bugs. A slight downside of this approach is that you have to check the size of the list before and after if you want to know if the expression actually deleted anything.

lekhajee: @ksheedlo : I am aware that the length of the list changes when an item is deleted. (even without having to use intuition :) )

Has accessing an element when modifying the list or vice versa been an issue so far. I think we are over thinking this. All the plugins use the approach of finding an element from the list and we have even run performance tests and not had an issue.

ksheedlo: I just took a look through #412 again. It is true, we have not had a serious bug with iterating and modifying the list, even with the old way the code was written. However, it can lead to confusing code. For instance, you might see:

```python
# Delete an entity
for i in range(len(entities)):
    if entities[i].id == entity_id:
        del entities[i]
        break

for check in checks:
    if check.entity_id == entity_id:
        del checks[checks.index(check)]
```

There are 2 obvious problems with this code that I can think of off the top of my head:

1. It's very easy to look at this code, perhaps from the angle of copy and pasting it into some other delete function, and do this:

    ```python
    for i in range(len(checks)):
        if checks[i].entity_id == entity_id:
            del checks[i] # BUG!
    ```

2. Deleting a check becomes an O(N^2) operation, by having to lookup the index every single time a check that needs to be deleted is found.

ksheedlo: What if we implemented a custom equality class and simply called `list.remove`? For instance:

```python
try:
    checks.remove(Matcher(lambda check: check.entity_id == entity_id))
except ValueError:
    # handle not found case
```

lekhajee: sure, wfm

ksheedlo:
@lekhajee Updated.

lekhajee:
+1

PR #439, opened 2015-09-25
Including the IAD region to load on the nova-api for onmetal flavors

lekhajee:


ksheedlo:
LGTM to me

mimic/rest/nova_api.py
```diff
@@ -35,7 +35,7 @@ class NovaApi(object):
     Rest endpoints for mocked Nova Api.
     """
 
-    def __init__(self, regions=["ORD"]):
+    def __init__(self, regions=["ORD", "IAD"]):
```
tawalton: Can you add DFW also?

PR #441, opened 2015-09-26
feat(maas_api): square wave override metrics

ksheedlo:
This change adds back the square wave functionality that used to be in
the MaaS plugin. Users can specify that square wave metrics will be
returned for any entity, check and metric. They can also specify the
period, offset, and upper and lower values.

### DIFF
--- a/mimic/rest/maas_api.py
+++ b/mimic/rest/maas_api.py
@@ -2,6 +2,8 @@
 MAAS Mock API
 """
 
+from __future__ import division
+
 import json
 import collections
 import urlparse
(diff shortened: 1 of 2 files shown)


ksheedlo:
screenshot

screen shot 2015-09-25 at 8 57 07 pm

lekhajee:
@ksheedlo : just curious, is cloud intelligence not using this currently. Are there plans to use this in future

lekhajee:
+1

ksheedlo:
There's a long-standing issue to write tests for the actual content of graphs, which would require this. We don't use it currently.

PR #444, opened 2015-09-30
Added canned responses to Auth API

tpboudreau-rackspace:
Need to add a few canned responses to the Identity API that provide user data containing contact IDs that match those in the fixed permissions in the Valkyrie API plugin.

### DIFF
--- a/mimic/rest/auth_api.py
+++ b/mimic/rest/auth_api.py
@@ -345,6 +345,120 @@ class AuthApi(object):
                 {"id": "observer",
                  "description": "Global Observer Role.",
                  "name": "observer"}]
+
+        if token_id in get_presets["identity"]["non_dedicated_observer"]:
+                response["access"]["token"]["tenant"] = {
+                    "id": "135790",
+                    "name": "135790",
+                }
+                response["access"]["user"] = {
+                    "name": "OneTwo",
+                    "roles": [{"id": "1",
+                               "name": "monitoring:observer",
+                               "description": "Monitoring Observer"}]
+                }
+
+        if token_id in get_presets["identity"]["non_dedicated_admin"]:
+                response["access"]["token"]["tenant"] = {
+                    "id": "135790",
+                    "name": "135790",
+                }
+                response["access"]["user"] = {
+                    "name": "ThreeFour",
+                    "roles": [{"id": "1",
+                               "name": "monitoring:admin",
+                               "description": "Monitoring Admin"},
+                              {"id": "2",
+                               "name": "admin",
+                               "description": "Admin"}]
+                }
+
+        if token_id in get_presets["identity"]["dedicated_full_device_permission_holder"]:
+                response["access"]["token"]["tenant"] = {
+                    "id": "hybrid:123456",
+                    "name": "hybrid:123456",
+                }
+                response["access"]["user"] = {
+                    "id": "12",
+                    "name": "HybridOneTwo",
+                    "roles": [{"id": "1",
+                               "name": "monitoring:observer",
+                               "description": "Monitoring Observer"},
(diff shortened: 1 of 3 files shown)


mimic/rest/auth_api.py
```diff
@@ -345,6 +345,120 @@ def validate_token(self, request, token_id):
...
+                    "name": "HybridNineZero",
+                    "roles": [{"id": "1",
+                               "name": "monitoring:admin",
+                               "description": "Admin"},
+                              {"id": "2",
+                               "name": "admin",
+                               "description": "Admin"}],
+                    "RAX-AUTH:contactId": "90"
+                }
+
```
ksheedlo: I'm concerned that this pattern is going to introduce tight coupling between Mimic's Valkyrie mock and the MaaS test suite. Would it be possible to go through either a normal or control API and have MaaS's test driver specify to Mimic what kinds of responses it needs?

mimic/test/test_auth.py
```diff
@@ -1286,3 +1286,75 @@ def test_string_errors_as_well_as_json_errors(self):
...
+        url = "/identity/v2.0/tokens"
+        core, root = core_and_root([])
+        (response, content) = self.successResultOf(
+            json_request(self, root, "GET", url + "/HybridOneTwo"))
+        self.assertEqual(200, response.code)
+        self.assertTrue(content["access"])
+        self.assertTrue(content["access"]["token"])
+        self.assertTrue(content["access"]["token"]["tenant"])
+        self.assertTrue(content["access"]["token"]["tenant"]["id"])
+        self.assertEqual(content["access"]["token"]["tenant"]["id"], "hybrid:123456")
```
ksheedlo: IMO all the assertions leading up to this one are redundant, since this will throw an exception if the response is malformed.

mimic/test/test_auth.py
```diff
@@ -1286,3 +1286,75 @@ def test_string_errors_as_well_as_json_errors(self):
...
+        self.assertTrue(content["access"]["user"])
+        self.assertTrue(content["access"]["user"]["RAX-AUTH:contactId"])
+        self.assertEqual(content["access"]["user"]["RAX-AUTH:contactId"], "12")
```
ksheedlo: Likewise for this and down the page as far as asserting parent pieces of the response structure.

mimic/rest/auth_api.py
tpboudreau-rackspace: The concern is valid.  In fact I'd go as far as to say that removing fixtures entirely from both these APIs and having test suites load their own data would be preferable.  But that requires the work of adding ingestion endpoints to both API's (I strongly considered doing this), or learning/creating some other mechanism for loading data (does one exist? I'm not a Mimic expert).

For now I think I need this expediency.  If the Mimic maintainers reject this, I'll likely just have to run a MaaS fork of Mimic until I can find some time to do the job right.

mimic/test/test_auth.py
tpboudreau-rackspace: Was on the fence myself.  Removed.

mimic/rest/auth_api.py
ksheedlo: @tpboudreau-rackspace since I know what this is getting used for inside MaaS, I understand the need for expediency. I don't want you to have to run a fork. If MaaS can promise to help us iterate off these features once Mimic supports the proper ingestion endpoints, I'm happy to land this change as-is.

ksheedlo:
LGTM

@tpboudreau-rackspace let's not forget to do this right later. I'd like to have a discussion about what ingestion endpoints could be added to make this thing work without all the fixtures.

PR #445, opened 2015-09-30
Add get any permissions endpoint to Valkyrie API

tpboudreau-rackspace:
The Repose Valkyrie filter is being reworked to include a call to endpoint /account/{account_number}/permissions/contacts/any/by_contact/{contact_id}/effective, so this change simulates it.

### DIFF
--- a/mimic/rest/valkyrie_api.py
+++ b/mimic/rest/valkyrie_api.py
@@ -36,6 +36,20 @@ class ValkyrieApi(object):
         """
         return self.core.valkyrie_store.create_token(request)
 
+    effective_any_permissions_route = ('/account/<int:account_number>'
+                                       '/permissions/contacts/any'
+                                       '/by_contact/<int:contact_id>/effective')
+
+    @app.route(effective_any_permissions_route, methods=['GET'])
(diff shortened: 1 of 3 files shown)


mimic/rest/valkyrie_api.py
```diff
@@ -36,6 +36,20 @@ def login_user(self, request):
         """
         return self.core.valkyrie_store.create_token(request)
 
+    effective_any_permissions_route = ('/account/<int:account_number>'
+                                       '/permissions/contacts/any'
+                                       '/by_contact/<int:contact_id>/effective')
```
ksheedlo: Cool syntax. I wouldn't have thought this was valid python :mortar_board:

ksheedlo:
LGTM

This PR inherits my concerns from #444 about fixtures etc, but it's needed for MaaS and I trust @tpboudreau-rackspace to do the right thing and follow up.

mimic/rest/valkyrie_api.py
tpboudreau-rackspace: Yeah, flake8 forced be to get creative.

PR #446, opened 2015-10-01
Added API for ElasticSearch

ayersek64:


ksheedlo:
Hi @ayersek64 ! Thanks for the PR.

I see you're a Racker. Can you tell us a little bit about what project you're working on and why you're interested in using Mimic to mock Elasticsearch?

ayersek64:
I'm working on Abacus, a metric reporting tool for several Rackspace products like Cloud Servers, Cloud Load Balancers, and RackConnect.
Here is a link to the production page:
https://abacus.rax.io/

I'm currently working on changes to the Product Report Card in the Production menu. This view gets some of its data from an ElasticSearch instance running in production, but it is very slow for Development and difficult to control the response data to see different app behaviors. I would like to have a localized alternative and Mimic fits that need.

-Kevin Ayers
  BRDS SW Developer, IPA Shared Services
  Rackspace - The #1 Managed Cloud Company
  office: (210) 312-5888 | mobile: (512) 997-5372

ayersek64:
Can someone describe what I need to do to get the Travis CI build to pass?

I see from the details that the builds are failing during unit tests when executing the ElasticSearch unit test on 'import mock'.
I have updated requirements.txt to include 'mock 1.3.0', so I'm not sure why the CI environment does not have mock available for the unit test.

Is the requirements.txt file not used to setup the CI build environment? If not, what else do I need to update to ensure mock is included?

Thanks.

glyph:
@ayersek64 - No, `requirements.txt` is not used; it is installed from `setup.py`.  Since mimic is an application designed to be installed from PyPI, you always have to add dependencies to `setup.py` since that is the primary source of truth.

However, more importantly, we don't use the `mock` module in our tests on purpose.  `mock` leads to poor-quality tests that stub out important pieces of the implementation and don't provide a high degree of assurance that things aren't broken.  In fact Mimic's tag line has been, at times, 'testing without mocking' :).

glyph:
@ayersek64 - The best way to eliminate usage of `mock` in these tests would be to get rid of the `file_based_responses` module and replace it with responses generated from a set of model objects which represent elasticsearch indexes.  If you land the PR in the state that it's in now, where the ES responses come pre-initialized with data specific to Abacus, if your tests *using* mimic depend on this they will eventually break, when someone else who wants to use Mimic to test *their* Elasticsearch cluster either changes this data to be representative of their application or, more likely to get through code review, changes it so that it starts with an empty ElasticSearch cluster and must be populated.

glyph:
@ayersek64 - to make it easier to figure out if the tests are going to pass in travis-ci, you can run [`tox`](https://pypi.python.org/pypi/tox) (or [`detox`](https://pypi.python.org/pypi/tox)) locally and you should get test results that install Mimic the same way.

mimic/file_based_responses/elasticsearch.py
```diff
@@ -0,0 +1,74 @@
...
+
+            if response == None:
+                # something is wrong with the request; i.e. response is not
+                # available in the data file for requested index and/or type
+                response_code = 400
+                response = {'message':
+                            ('response for index/type not available in'
+                             ' ElasticSearch data file [elasticsearch.json]')}
+
+            log.msg("\n{'ES status': 'ok', 'indexes': " + ",".join(indexes) +
```
glyph: Rather than doing string manipulation to produce a log message, make use of Twisted's structured logging support.  This would be `from twisted.logger import Logger` rather than `from twisted.python import log`; more details [here](https://twistedmatrix.com/documents/15.4.0/core/howto/logger.html).

mimic/resource.py
```diff
@@ -22,7 +22,8 @@
 )
 from mimic.rest.noit_api import NoitApi
 from mimic.rest import (fastly_api, mailgun_api, customer_api,
-                        ironic_api, glance_api, valkyrie_api)
+                        ironic_api, glance_api, valkyrie_api,
+                        elasticsearch_api)
```
glyph: This is not a blocking comment since it's hardly *this* PR's fault, but this set of "plugin" imports is getting out of control.  We really need to address #108 soon.

mimic/test/test_elasticsearch.py
```diff
@@ -0,0 +1,230 @@
...
+
+
+class ElasticSearchAPITests(SynchronousTestCase):
+
+    """
+    Tests for the ElasticSearch API
+    """
+
+    # JSON data to be used by mocked JSON load function
+    unittest_json = [
```
glyph: As mentioned in earlier comment, this should be eliminated.  Either test the actual JSON as implemented in the service, or better yet, eliminate the canned JSON and synthesize a response based on some elasticsearch model objects.

README.md
```diff
@@ -1,6 +1,6 @@
 <img alt="MIMIC" src="https://i.imgur.com/eodn4M4.png" height="75">
 
-Mimic is an API-compatible mock service for  __Openstack Compute__ and __Rackspace's implementation of Identity and Cloud Load balancers__. It is backed by in-memory data structure rather than a potentially expensive database.
+Mimic is an API-compatible mock service for  __Openstack Compute__, __Rackspace's implementation of Identity and Cloud Load balancers__, and __more__. It is backed by in-memory data structure rather than a potentially expensive database.
```
glyph: Thanks for the doc update.  It might be better to land this in a separate branch, since this is not the first "and more" :)

ayersek64:
Wow. Interesting tag line.

I've been using mock since I've started at Rackspace 4 months ago on the Abacus project and it has been extremely helpful in isolation method/function execution in unit tests and I've actually discovered and resolved numerous bugs in the code. The mocking I did for Mimic actually helped me find numerous problems with my code as well. However, my experiences don't matter and the combined expertise of the project team do, so I will work on the modifications you recommended.

On another note, the JSON file based response was intended to provide dynamic control of return data  without having to restart the Mimic engine each time new data is needed. I do see your point about another team committing a modified JSON file that would break my use of Mimic. Originally, I wanted some way to specify the file to use when launching Mimic, but didn't see a clean way to do this.

My time for Mimic work is limited at this point, so updates will be slow in coming, but I will do my best to get things working they way the Mimic team prefers.

-Kevin Ayers
  BRDS SW Developer, IPA Shared Services
  Rackspace - The #1 Managed Cloud Company
  office: (210) 312-5888 | mobile: (512) 997-5372

glyph:
> Wow. Interesting tag line.

Well, it's also the current tag line that Mimic is "an API-compatible mock", so we have a fraught relationship with the word "mock" ;-).

> I've been using mock since I've started at Rackspace 4 months ago on the Abacus project and it has been extremely helpful in isolation method/function execution in unit tests and I've actually discovered and resolved numerous bugs in the code. The mocking I did for Mimic actually helped me find numerous problems with my code as well

`mock` can be a great *introduction* to the art of test doubles; and if you're not in the habit of requiring 100% test coverage and deterministic execution of every code path, then it can really elevate your testing.  However, every test has multiple purposes; `mock` is great for making tests more unit-test-ish (i.e. more isolated from irrelevant concerns), but it's horrible for making tests more regression-test-ish (i.e. useful for detecting future breakages).  There are a lot of techniques which can give you similar results to make tests just as isolated but more useful for testing the code as it's actually used.

You might find this video interesting: https://www.youtube.com/watch?v=Xu5EhKVZdV8

> However, my experiences don't matter and the combined expertise of the project team do, so I will work on the modifications you recommended.

I wouldn't say that your experiences don't matter at all.  You are after all a member of the project team as much as anybody :).

> On another note, the JSON file based response was intended to provide dynamic control of return data  without having to restart the Mimic engine each time new data is needed. I do see your point about another team committing a modified JSON file that would break my use of Mimic. Originally, I wanted some way to specify the file to use when launching Mimic, but didn't see a clean way to do this.

This is a discussion that periodically comes up, and we may need to find a way to address it.  Generally, Mimic's preference is to populate data by using the same "real" public APIs that you'd use to populate data in a real service, to make sure as little code as possible is specific to mimic.  For example, when testing Nova, you "boot servers" in Mimic just as you would against a real cloud to create new servers, rather than specifying servers that should exist on the command line.

However, many folks know what configuration they want to test against and it's not entirely clear when or how to issue those API requests to populate the data that they want.  So Mimic *is* deficient in that it doesn't provide a clear path to add test fixtures when you want to start it up with preconfigured.

> My time for Mimic work is limited at this point, so updates will be slow in coming, but I will do my best to get things working they way the Mimic team prefers.

This is totally understandable; Mimic is nobody's full-time job, so we all get this to a greater or lesser degree.

While I think we have a clear idea of an *ideal* state (don't include any test data, populate it with real elasticsearch methods) I don't think we should block you from integrating this until it's totally perfect.  But this comment is getting pretty long already so I'll elaborate more later :)

ayersek64:
Thank you for the reply and the video link.

I'm always looking to improve my testing skills and strategies and the video 'puts another tool in my toolbox'.

Thanks again.

-Kevin Ayers
  BRDS SW Developer, IPA Shared Services
  Rackspace - The #1 Managed Cloud Company
  office: (210) 312-5888 | mobile: (512) 997-5372

glyph:
hey @ayersek64 - I realize that you don't have a lot of time to spend on maintaining Mimic, but if you're using it for testing, I'm concerned that we may end up with divergent versions of Mimic deployed publicly.  Do you have a few cycles to maybe clean this up into something minimal that could be landed - perhaps stacking on top of the in-progress top-level mock plugin support in #500 - and then stuffing your test data into it via a control-plane API?

ayersek64:
Unfortunately, I'm finding it very difficult to find cycles outside my required workload.

I'm thinking I need to cancel the PR and try again in the future.

I'm not using Mimic for testing anymore. We scored QE resources and they stood up instances of ElasticSearch and populated it with deterministic data alleviating the need to use Mimic.

-Kevin Ayers
  BRDS SW Developer, IPA Shared Services
  Rackspace - The #1 Managed Cloud Company
  office: (210) 312-5888 | mobile: (512) 997-5372

glyph:
OK. If you're definitely not going to work on it, let's close this PR.

PR #448, opened 2015-10-03
chore(mimic): deprecate Python 2.6

ksheedlo:
Addresses https://github.com/rackerlabs/mimic/issues/318#issuecomment-145173987.

### DIFF
--- a/mimic/resource.py
+++ b/mimic/resource.py
@@ -5,6 +5,7 @@ Resources for Mimic's core application.
 from __future__ import unicode_literals
 
 import json
+import sys
 
 from io import BytesIO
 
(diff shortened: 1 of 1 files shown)


ksheedlo:
FYI, we won't be able to get full coverage on this because which branch runs depends on the Python version.

glyph:
I'm pretty sure @codecov-io is doing combined coverage based on *all* our travis builds.

glyph:
Is codecov not invoked after the py26 run?

ksheedlo:
@glyph I don't know how to tell if codecov is invoked or not.

ksheedlo:
@glyph I see Python 2.6 ran as Travis job 1756.1 and that number is not present in the Codecov report.

glyph:
@ksheedlo can you try merging master? The coverage invocation was tweaked to be somewhat simpler.  Nothing directly relevant to this, but there are fewer things to investigate if it's still failing with master's coverage configuration.

ksheedlo:
This was rebased a few hours ago, so it should be current.

glyph:
I did one more rebuild just to see, but honestly I have no idea why 2.6 isn't reporting coverage.  If we just can't make it work this way, put it in a function and have a test for each code path by calling the function with a version tuple of 2.6 and 2.7.

glyph:
We need to figure out how to make this pass codecov; taking it out of review for the time being ( @ksheedlo is on vacation and probably won't look at this for a while, but it's fine, it's not high priority)

ksheedlo:
Can we close this? I'm not sure it's necessary anymore since #501 has landed.

glyph:
Yeah. Our last known 2.6 user stopped using it, and lots of people were interested in py3 which conflicts in terms of dependencies.  Thanks for pointing this out.  Adios 2.6!

PR #451, opened 2015-10-06
feat(maas): support mocked agent host info

ksheedlo:
This change adds a control endpoint to create an agent on your mimicked
entity. Agents will return random values for agent metrics when queried
for host info.


ksheedlo:
I don't understand why codecov is failing.

tawalton:
@ksheedlo It looks like the coverage on this branch is different from master.  Appears to be here: https://codecov.io/github/rackerlabs/mimic/mimic/canned_responses/maas_agent_info.py?ref=2472f8825b20e38fcd12d056f57045db5f25b606

ksheedlo:
Ah, I killed that code. Good eye @tawalton !

mimic/model/maas_objects.py
```diff
@@ -484,80 +599,174 @@ def __init__(self, clock):
...
-                Metric(name='user_percent_average', type=METRIC_TYPE_NUMBER, unit='percent'),
-                Metric(name='wait_percent_average', type=METRIC_TYPE_NUMBER, unit='percent'),
-                Metric(name='sys_percent_average', type=METRIC_TYPE_NUMBER, unit='percent'),
-                Metric(name='idle_percent_average', type=METRIC_TYPE_NUMBER, unit='percent'),
-                Metric(name='irq_percent_average', type=METRIC_TYPE_NUMBER, unit='percent'),
-                Metric(name='usage_average', type=METRIC_TYPE_NUMBER, unit='percent'),
-                Metric(name='min_cpu_usage', type=METRIC_TYPE_NUMBER, unit='percent'),
-                Metric(name='max_cpu_usage', type=METRIC_TYPE_NUMBER, unit='percent'),
-                Metric(name='stolen_percent_average', type=METRIC_TYPE_NUMBER, unit='percent')]),
+                _agent_metric(name='user_percent_average', type=METRIC_TYPE_NUMBER, unit='percent'),
```
glyph: It seems to me that this `__init__` function is growing out of control.  There aren't tests for every single one of these checks and metrics, and we seem to keep adding to them.  Is there any automated way we could keep all these metrics in sync with the actual public API, some data list we could be importing and parsing, rather than procedurally generating said list on every `__init__`?

If this really has to be manually constructed, it might be good to start grouping things by common features to reduce the line-count bloat; for example, in this case, something like:

```python
MetricListBuilder(type=METRIC_TYPE_NUMBER, unit='percent').metrics(
    'user_percent_average', 'wait_percent_average', 'sys_percent_average', # ...
)
```

we could potentially make this a lot shorter and less repetitive that way.

glyph:
Other than my concern about the ever-growing `MaasStore.__init__` I think this looks OK.  It's already pretty big, so I think you can file a follow-up for that.

mimic/model/maas_objects.py
ksheedlo: I'll give that a shot.

ksheedlo:
@glyph 

I have a branch that implements a refactor of the `MaasStore`. You can see it at https://github.com/ksheedlo/mimic/commit/8d0678465e7fc208c352ced1fa1df84cf4a1e4d0. I was able to cut the line count in `MaasStore.__init__` by about half by implementing a stateful builder. However, the overall code size actually increased slightly and I feel it also made the code more brittle.

I'm not going to file a PR, but you're of course welcome to look in my repo and make comments or suggestions.

glyph:
How do you feel it's more brittle?  I'll play with it and see if I can propose a PR, perhaps you can review.

PR #453, opened 2015-10-07
refactor(py3): use range instead of xrange

ksheedlo:


glyph:
These are all pretty small, mostly fixed ranges, so it's likely not important, but in the future you might want to consider [`six.moves.range`](http://six.readthedocs.org/#module-six.moves), which provides consistent semantics - `xrange` on py2 and `range` on py3 - so it's an on-demand iterator in both cases.

PR #455, opened 2015-10-12
chore(py3): future import unicode_literals everywhere

ksheedlo:
Discussed offline with @glyph last week - this change is intended to help us reason through where we need to use byte strings vs where we can use Unicode strings. The hope is that this will smooth the process of adding python3 support.

With this change, the new policy for string literals (such as there is one) is

1. All string literals are `unicode`, except
2. explicit `b"bytestring literals"`. Use these for interfacing with Klein/Twisted, but not for generally storing human readable text or as dictionary keys.

### DIFF
--- a/mimic/canned_responses/auth.py
+++ b/mimic/canned_responses/auth.py
@@ -2,6 +2,9 @@
 """
 Canned response for get auth token
 """
+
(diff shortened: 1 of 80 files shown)


lekhajee:
+1

PR #456, opened 2015-10-12
Adding py3 support

lekhajee:
- adds py34 to tox
- allows the py34 build to fail without failing travis
- updates requirements to install twisted 15.4 
- removes references to old logging in twisted
- makes `random_hex_generator` py3 compatible
- Changes urllib to six.moves.urllib.parse to support both py27 and py34

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,17 +12,13 @@ env:
 matrix:
     include:
         - python: 2.6
-          env: TOXENV=py26-twisted_old_logging
-        - python: 2.6
-          env: TOXENV=py26-twisted_new_logging
+          env: TOXENV=py26
         - python: 2.7
-          env: TOXENV=py27-twisted_old_logging
-        - python: 2.7
-          env: TOXENV=py27-twisted_new_logging
+          env: TOXENV=py27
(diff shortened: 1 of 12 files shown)


glyph:
@lekhajee - Would you mind updating this so that it's an allow-fail build, so we can put this on trunk and at least be aware if other changes make it worse while we're working on it?

glyph:
This needs a `setup.py` version bump as well.

glyph:
For Twisted, I mean; right now it still says `>=14.0.2` but we are dropping those versions from our test matrix.

glyph:
I think we should land this before making too many other py3 changes and making it too hard to review - so please just adjust distutils metadata and land :)

ksheedlo:
Where does this change leave py26 support? #318

glyph:
@ksheedlo - we haven't killed it yet… ;). (Nor should *this* change kill it.  This should be just to get the tests bootstrapped.  We may well want to kill it in a subsequent PR before we get to move `py34` out of `allow_failures` though)

ksheedlo:
Sounds good to me.

PR #459, opened 2015-10-14
The image store needs to be linked to a session

tawalton:
Implementing the create image route for nova exposed an issue with the current image store. When an image was created, it persisted between sessions.  This PR adds the create image route and ties the image store to a particular session so a created image no longer persists.


mimic/test/test_nova.py
```diff
@@ -940,6 +941,41 @@ def test_rebuild(self):
             }
         })
 
+    def test_create_image(self):
```
glyph: We should really have test docstrings in the style of https://jml.io/pages/test-docstrings.html for all our unit tests.

mimic/rest/nova_api.py
```diff
@@ -223,13 +224,21 @@ def _region_collection_for_tenant(self, tenant_id):
...
-            lambda: GlobalImageCollection(tenant_id=tenant_id,
-                                          clock=self._session_store.clock))
+            "nova_image_collection",
+            lambda: GlobalNovaImageCollection(tenant_id=tenant_id,
+                                              clock=self._session_store.clock))
         image_region_collection = image_global_collection.collection_for_region(
             self._name)
         return image_region_collection
 
+    def _image_store_for_tenant(self, tenant_id):
```
glyph: I know this is a private method, but a docstring will help future maintainers understand what the purpose is.  It would be helpful to add one.

glyph:
Passing the image store as a separate object seems a bit weird to me.  Shouldn't the tenant's image collection object be taking the image store into account as an attribute of `self`?

mimic/model/rackspace_image_store.py
```diff
@@ -0,0 +1,60 @@
...
+                              RackspaceCentOSPVHMImage, RackspaceCoreOSImage, RackspaceDebianImage,
+                              RackspaceFedoraImage, RackspaceFreeBSDImage, RackspaceGentooImage,
+                              RackspaceOpenSUSEImage, RackspaceRedHatPVImage, RackspaceRedHatPVHMImage,
+                              RackspaceUbuntuPVImage, RackspaceUbuntuPVHMImage, RackspaceVyattaImage,
+                              RackspaceScientificImage, RackspaceOnMetalCentOSImage,
+                              RackspaceOnMetalCoreOSImage, RackspaceOnMetalDebianImage,
+                              RackspaceOnMetalFedoraImage, RackspaceOnMetalUbuntuImage)
+
+
+@attributes([Attribute("image_store", default_factory=list)])
```
glyph: If this is an `ImageStore`, but it also *has* an `image_store`, what is the difference in role between this object and that attribute?  I think you should add a `:var:` ReST field to the class docstring to document this attribute.

Also, `default_factory` says it's a `list`.  Perhaps it would be better named as `images_list`?

glyph:
If we're starting to have a lot of `rackspace_` modules, I think we should make a `rackspace/` package which contains them, to make it easier to separate out rackspace-specific and openstack-generic functionality.

mimic/model/nova_image_collection.py
```diff
@@ -6,7 +6,7 @@
 
 from characteristic import attributes, Attribute
 from json import dumps
-from mimic.model.rackspace_images import (ImageStore, OnMetalImage)
+from mimic.model.rackspace_images import OnMetalImage
```
glyph: If we're re-naming this module to be specifically `nova_image_collection`, it's a little distressing to me that it is still importing rackspace-specific images.  Good that it is importing fewer rackspace-specific things though.  Do we have a ticket for cleaning this up?

mimic/model/rackspace_image_store.py
tawalton: @glyph I will address this. The naming is a bit confusing.

glyph:
I think the most important thing here is just more verbiage in the docstrings explaining the relationships between all the new objects.  I'm not even 100% confident in my assessment that image store should be an attribute of image collection, but if it's not, the docs should explain clearly enough how they're related that I am 100% confident one way or the other :).

tawalton:
@glyph

glyph:
@tawalton - why'd you close?

tawalton:
@glyph I didn't... so strange

tawalton:
@glyph passing the image store object around does seem weird in the context of this PR. There will be a follow up PR for the glance api that needs to use the same image store.  The image collection will access the image store to list and get images and glance will have its own image collection that will also access the image store.  See diagram here https://github.com/rackerlabs/mimic/issues/416

tawalton:
@glyph I think we need to separate out rackspace-specific and openstack-generic functionality.  The nova image routes are dependent on the image store that contains only rackspace-specific images.  What do you think is the best way to decouple this?

glyph:
> @glyph passing the image store object around does seem weird in the context of this PR. There will be a follow up PR for the glance api that needs to use the same image store. The image collection will access the image store to list and get images and glance will have its own image collection that will also access the image store. See diagram here #416

It does make sense to me that the image store would be a separate object, and could be passed separately to other things; the part that I think should be changed is that in, i.e. `get_image`, it seems like it ought to be an *attribute* on the `RegionalNovaImageCollection` returned by `_image_collection_for_tenant`.  After all, `RegionalNovaImageCollection` is *already* associated with a session, it just doesn't have a back-reference *to* the session; so the API that creates it should just hand it a reference to the `ImageStore` it should use.

Does that make sense?

glyph:
> @glyph I think we need to separate out rackspace-specific and openstack-generic functionality. The nova image routes are dependent on the image store that contains only rackspace-specific images. What do you think is the best way to decouple this?

Thus far we have been able to let the defaults be the defaults; with this list of images, we are fast approaching a point where we will really need to have some kind of control-plane API which allows the user to globally select the default cloud profile (four immediately spring to mind: rackspace public, rackspace private, vanilla openstack, custom) that new tenants will get.

I don't think we need to resolve this tension to land this PR but we do need to have an issue to start planning and discussing it.

glyph:
@tawalton - can you resolve conflicts?  sorry that this got out of date, but with the py3 stuff our focus has been elsewhere for a little while…

tawalton:
@glyph done

mimic/model/nova_objects.py
```diff
@@ -818,7 +819,8 @@ def request_delete(self, http_delete_request, server_id):
         server.update_status(u"DELETED")
         return b''
 
-    def request_action(self, http_action_request, server_id, absolutize_url):
+    def request_action(self, http_action_request, server_id, absolutize_url,
+                       regional_image_collection, image_store):
```
glyph: `request_action` is a method on `RegionalServerCollection`.  `RegionalServerCollection` already has an attribute called `image_store`.  Why does this method also take a `regional_image_collection` and `image_store` rather than these being attributes on `RegionalServerCollection`?

mimic/rest/nova_api.py
```diff
@@ -315,18 +325,16 @@ def delete_server(self, request, tenant_id, server_id):
...
+                .request_delete(request, server_id))
 
     @app.route('/v2/<string:tenant_id>/images/<string:image_id>', methods=['GET'])
     def get_image(self, request, tenant_id, image_id):
         """
         Returns a get image response, for any given imageid
         """
-        return(self._image_collection_for_tenant(tenant_id)
-               .get_image(request, image_id, absolutize_url=self.url))
+        return (self._image_collection_for_tenant(tenant_id)
```
glyph: While this new formatting is correct, if these changes are irrelevant to this PR, could you remove them so the diff is smaller to review?

mimic/test/test_nova.py
```diff
@@ -4,6 +4,7 @@
 from __future__ import absolute_import, division, unicode_literals
 
 import json
+import random
```
glyph: I don't think that we should be selecting the image randomly in the test case.  If we need to test with multiple images, we should explicitly select the different images being tested with and parameterize `test_create_image` so that we can have multiple test cases that call it with different images.  If the test becomes sensitive to image selection and fails in the case of certain images but not others, this will be very hard to diagnose.

mimic/model/nova_objects.py
```diff
@@ -931,6 +933,37 @@ def request_action(self, http_action_request, server_id, absolutize_url):
...
+            image = image_store.get_image_by_id(image_ref)
+            image_json = regional_image_collection.get_image(http_action_request,
+                                                             image_ref, absolutize_url)
+            image_dict = loads(image_json)
+            flavor_classes = image_dict['image']['metadata']['flavor_classes']
+            os_type = image_dict['image']['metadata']['os_type']
+            os_distro = image_dict['image']['metadata']['org.openstack__1__os_distro']
+            vm_mode = image_dict['image']['metadata']['vm_mode']
+            disk_config = image_dict['image']['metadata']['auto_disk_config']
+            image_id = str(uuid.uuid4())
```
glyph: `str` is a red flag for python 3 compatibility.  Does this value represent bytes or text?  Since it's going to be used as an image ID, which is textual (and shows up in `JSON`), you should probably be using `six.text_type` here.

glyph:
Sorry @tawalton - still have a bit more review feedback.  Can you address the points above?

mimic/test/test_nova.py
tawalton: @glyph you want to specify in the test which server image you want to create?

mimic/rest/nova_api.py
tawalton: @glyph  I would have to disagree that a white space change really adds to the difficulty of reviewing this PR and would prefer not to address this

mimic/model/nova_objects.py
tawalton: @glyph  will do

mimic/test/test_nova.py
glyph: @tawalton - yes, perhaps creating one test case for each server image programatically.

mimic/rest/nova_api.py
glyph: I try to review each line carefully.  This actually makes cleanups that don't change anything else particularly hard to review, because I have to first realize that nothing has changed, then verify that nothing has changed.  It adds to the effort of each review.  Furthermore, it pollutes change history, by making it harder to read `git annotate` and diagnose why a certain line is a certain way.  Although this is a content-free reformat, it will appear alongside a merge commit message that suggests something more substantive is changing.  Given that Mimic has relatively light maintenance documentation and is in a constant state of evolving towards a better API, `annotate` is a particularly important and powerful tool for understanding how things got to where they are.

This doesn't mean we should never do cleanups, but rather, cleanup branches should be their own small changes.

Since you are in a bit of a time crunch here I can build on top of your PR and resubmit for review myself though.

glyph:
I'm going to try to merge master to sync the coverage numbers…

glyph:
Although codecov is OK with the coverage drop (I'm not sure why?) this apparently removes all test coverage of `mimic.model.rackspace_images.ImageStore` here.  In other words, although the *coverage on the diff* of the branch is 100%, the *coverage diff of* the branch is negative.  Can we just remove `ImageStore` entirely now?  Is it used?

glyph:
Oops - meant to link to the codecov report: https://codecov.io/github/rackerlabs/mimic/mimic/model/rackspace_images.py?ref=b280e2146077d8ce90d2ac61e4c6ea2023196213#l-828

tawalton:
@glyph good catch - image store has been removed

mimic/model/nova_objects.py
```diff
@@ -936,6 +938,37 @@ def request_action(self, http_action_request, server_id, absolutize_url):
                                          " while it is in task state other than active",
                                          http_action_request))
 
+        elif 'createImage' in action_json:
```
glyph: For future reference, we should probably be splitting this method up into different pieces based on these "action verbs".

mimic/model/rackspace_image_store.py
```diff
@@ -0,0 +1,66 @@
...
+                                          RackspaceOnMetalUbuntuImage)
+
+
+@attributes([Attribute("image_list", default_factory=list)])
+class RackspaceImageStore(object):
+    """
+    A store for images to share between nova_api and glance_api
+    :var image_list: list of Rackspace images
+    """
+    def create_image_store(self, tenant_id):
```
glyph: This method's documentation is really misleading and I find myself somewhat confused reading it.

It says "generates the data for each image...", but in fact it only does this if the data hasn't already been populated.  Or if some other data were not already attached.

mimic/model/rackspace_image_store.py
```diff
@@ -0,0 +1,66 @@
...
+        """
+        Generates the data for each image in each image class
```
glyph: This docstring also neglects to mention that the method has a return value, and what that return value is.

mimic/model/rackspace_image_store.py
```diff
@@ -0,0 +1,66 @@
...
+                    minRam = image_spec['minRam']
+                    minDisk = image_spec['minDisk']
+                    image_size = image_spec['OS-EXT-IMG-SIZE:size']
+                    image = image_class(image_id=image_id, tenant_id=tenant_id,
+                                        image_size=image_size, name=image_name, minRam=minRam,
+                                        minDisk=minDisk)
+                    if 'com.rackspace__1__ui_default_show' in image_spec:
+                        image.set_is_default()
+                    self.image_list.append(image)
+        return self.image_list
```
glyph: This method has a side effect _and_ it has a return value; the return value is not based upon what the side effect was, but just whatever the current state happens to be after the side effect has taken place.  This is somewhat counter to the normal design of APIs within python; for example, consider `list.sort` - it doesn't return the list itself, so that you don't develop the mistaken impression that you can do `list.sort().reverse()` and not affect the original `list`; instead, there's a separate `sorted` function that does not modify the list in place but instead creates a copy.

mimic/model/nova_image_collection.py
```diff
@@ -6,32 +6,23 @@
...
-        for image in images_store:
-            if image.image_id == image_id:
-                return image
-
     def list_images(self, include_details, absolutize_url):
         """
         Return a list of images.
         """
-        images_store = ImageStore.create_image_store(self.tenant_id)
+        images_store = self.image_store.create_image_store(self.tenant_id)
```
glyph: Again, within `nova_image_collections`, a module named in a way that suggests it's a general-purpose thing for all Nova clouds, we are calling a method - `create_image_store` - that is only defined on a class within our rackspace-specific implementation.

mimic/model/nova_image_collection.py
```diff
@@ -6,32 +6,23 @@
...
+from mimic.model.rackspace_images import OnMetalImage
 
 from mimic.model.nova_objects import not_found
 from mimic.canned_responses.mimic_presets import get_presets
 
 
 @attributes(
-    ["tenant_id", "region_name", "clock"])
-class RegionalImageCollection(object):
+    ["tenant_id", "region_name", "clock", "image_store"])
```
glyph: This adds a new `image_store` attribute, but there's no documentation of what type it is or where it needs to come from.

mimic/model/rackspace_images.py
```diff
@@ -98,6 +98,83 @@ def detailed_json(self, absolutize_url):
         return template
 
 
+@attributes(['image_id', 'tenant_id', 'name', 'minDisk', 'minRam', 'image_size', 'server_id', 'links',
+             'flavor_classes', 'os_type', 'os_distro', 'vm_mode', 'disk_config'])
+class RackspaceSavedImage(object):
+    """
+    A Rackspace saved image object representation
```
glyph: This class looks a *lot* like `mimic.model.glance_objects.Image`.  How are they different?  If we add this class it seems like a lot of duplicated code to generate very similar JSON data structures...

glyph:
The biggest remaining problem on this PR is the further duplication of more stuff within the Glance plugin.

We have to do something in the near term to populate a list of Rackspace images.  I'd like to see a Rackspace-specific function that operates on a generic Glance image store, and just adds images to it, but doesn't create any new Rackspace-specific classes, just populates Rackspace-specific data.

However, I may not be fully understanding what is being added here; is there a Rackspace images API somewhere that I'm not familiar with, independent of the Nova images API and glance?

glyph:
This PR makes a few issues worse, but does not really introduce any of them.  Rather than block on having them all fixed before it can be landed, I've filed #492 that we should work on soon.

PR #460, opened 2015-10-15
Fix ssh key list

tawalton:
There is a bug in the current implementation of listing ssh keys that causes only the last key created to be returned in the list call.  Changing the data structure from a dict to a list will fix the issue


glyph:
LGTM. Up to 99% now! :)

PR #462, opened 2015-10-19
point py34 to run against twisted, treq and klein's master

lekhajee:


glyph:
Looks good to me (for now - we do need to do releases of everything as soon as this is green!)

PR #463, opened 2015-10-20
include absolute_import and division for py3 porting

lekhajee:


glyph:
Hooray, no test failures :).

PR #464, opened 2015-10-20
fix usage of random_hex_genarator

lekhajee:
Fixes the `builtins.TypeError: Can't convert 'bytes' object to str implicitly` test failure


glyph:
Is there *anywhere* we're using `random_hex_generator` that we actually want `bytes`?  It seems like it might make more sense to make it return `unicode` rather than fixing all the call sites.

lekhajee:
@glyph : you caught me :) I thought of this too but after I had made the changes and submitted the PR :( and let it be.

lekhajee:
Addressed in #465

PR #465, opened 2015-10-20
change random_hex_genarator to return unicode

lekhajee:
Fixes the `builtins.TypeError: Can't convert 'bytes' object to str implicitly` test failure


glyph:
Looks better :).

PR #466, opened 2015-10-21
Reassign auth preset response to allow better MaaS test coverage.

tpboudreau-rackspace:
This slight change in the preset responses in the Identity API allows for the testing of an unusual class of dedicated quasi-users that carry the 'hybridRole' role.


ksheedlo:
LGTM

tpboudreau-rackspace:
Thanks @ksheedlo.

PR #470, opened 2015-10-23
fix(maas_api): return less awkward type in catalog

ksheedlo:
The MaaS type in the service catalog was "rax: monitor" where it should
have been "rax:monitor".


tawalton:
:+1:

glyph:
:sparkles: :clap:

PR #471, opened 2015-10-26
tests passing on python 3.4

glyph:
now that we've landed changes on twisted trunk, treq, and klein, we should be able to get py3 to work

### DIFF
--- a/mimic/model/mailgun_objects.py
+++ b/mimic/model/mailgun_objects.py
@@ -118,5 +118,5 @@ class MessageStore(object):
         Retrieve a :obj:`Message` object by its `to` address.
         """
         for each_msg in self.message_store:
-            if each_msg.to == to_address[0]:
+            if each_msg.to == to_address:
                 return each_msg
--- a/mimic/rest/fastly_api.py
+++ b/mimic/rest/fastly_api.py
@@ -11,6 +11,18 @@ from mimic.rest.mimicapp import MimicApp
 from mimic.canned_responses import fastly
 
 
+def text_urldata(urldata):
+    """
+    Convert ``urldata`` into text.
+
+    :param urldata: a dictionary mapping bytes to lists of bytes.
+
+    :return: a dictionary mapping text to lists of text.
+    """
+    return dict([(k.decode("utf-8"), [vv.decode("utf-8") for vv in v])
+                 for [k, v] in urldata.items()])
+
+
 class FastlyApi(object):
 
     """
@@ -53,7 +65,7 @@ class FastlyApi(object):
 
         https://docs.fastly.com/api/config#service_5
         """
-        url_data = request.args.items()
+        url_data = text_urldata(request.args).items()
         response = self.fastly_response.create_service(url_data)
         return json.dumps(response)
 
@@ -74,7 +86,7 @@ class FastlyApi(object):
 
         https://docs.fastly.com/api/config#service_3
         """
-        url_data = request.args.items()
+        url_data = text_urldata(request.args).items()
         data = dict((key, value) for key, value in url_data)
         service_name = data['name'][0]
 
@@ -90,7 +102,7 @@ class FastlyApi(object):
 
         https://docs.fastly.com/api/config#domain_4
         """
-        url_data = request.args.items()
+        url_data = text_urldata(request.args).items()
         response = self.fastly_response.create_domain(url_data,
                                                       service_id, version_id)
         return json.dumps(response)
@@ -117,7 +129,7 @@ class FastlyApi(object):
 
         https://docs.fastly.com/api/config#backend_2
         """
-        url_data = request.args.items()
+        url_data = text_urldata(request.args).items()
         response = self.fastly_response.create_backend(url_data,
                                                        service_id, version_id)
         return json.dumps(response)
@@ -131,7 +143,7 @@ class FastlyApi(object):
 
         https://docs.fastly.com/api/config#condition_3
         """
-        url_data = request.args.items()
+        url_data = text_urldata(request.args).items()
         response = self.fastly_response.create_condition(url_data,
                                                          service_id, version_id)
         return json.dumps(response)
@@ -145,7 +157,7 @@ class FastlyApi(object):
 
         https://docs.fastly.com/api/config#cache_settings_3
         """
-        url_data = request.args.items()
+        url_data = text_urldata(request.args).items()
         response = self.fastly_response.create_cache_settings(url_data,
                                                               service_id, version_id)
         return json.dumps(response)
@@ -159,7 +171,7 @@ class FastlyApi(object):
 
         https://docs.fastly.com/api/config#response_object_3
         """
-        url_data = request.args.items()
+        url_data = text_urldata(request.args).items()
         response = self.fastly_response.create_response_object(url_data,
                                                                service_id, version_id)
         return json.dumps(response)
@@ -173,7 +185,7 @@ class FastlyApi(object):
 
         https://docs.fastly.com/api/config#settings_2
         """
-        url_data = request.args.items()
+        url_data = text_urldata(request.args).items()
         response = self.fastly_response.create_settings(url_data,
                                                         service_id, version_id)
         return json.dumps(response)
--- a/mimic/rest/loadbalancer_api.py
+++ b/mimic/rest/loadbalancer_api.py
@@ -323,7 +323,7 @@ class LoadBalancerRegion(object):
         """
         Deletes multiple nodes from a LB.
         """
-        node_ids = map(int, request.args.get(b'id', []))
+        node_ids = [int(node_id) for node_id in request.args.get(b'id', [])]
         response_data = self.session(tenant_id).delete_nodes(lb_id, node_ids)
         request.setResponseCode(response_data[1])
         return json_dump(response_data[0])
--- a/mimic/rest/loadbalancer_api.py
+++ b/mimic/rest/loadbalancer_api.py
@@ -292,7 +292,8 @@ class LoadBalancerRegion(object):
         """
         try:
             content = json_from_request(request)
-            assert isinstance(content, dict) and content.keys() == ["node"]
+            assert (isinstance(content, dict) and
+                    list(content.keys()) == ["node"])
             content = content["node"]
             assert isinstance(content, dict)
         except (ValueError, AssertionError):
--- a/mimic/rest/maas_api.py
+++ b/mimic/rest/maas_api.py
@@ -387,7 +387,7 @@ class MaasMock(object):
                 )
 
     def _audit(self, app, request, tenant_id, status, content=''):
-        headers = dict([(k, v) for k, v in request.getAllHeaders().iteritems()
+        headers = dict([(k, v) for k, v in request.getAllHeaders().items()
                         if k != 'x-auth-token'])
 
         self._entity_cache_for_tenant(tenant_id).audits_list.append(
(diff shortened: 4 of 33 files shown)


glyph:
Locally I see this:

```
___________________________________ summary ____________________________________
  py26: commands succeeded
  py27: commands succeeded
  py34: commands succeeded
  pypy: commands succeeded
  docs: commands succeeded
  lint: commands succeeded
  congratulations :)
```

glyph:
Sorry for the enormous slog of a diff, but that is the nature of py3 porting :-\.  Given that all the work here *can* be done in <24 hours, I don't think we should bother with the incremental process Twisted has had…

mimic/model/mailgun_objects.py
```diff
@@ -118,5 +118,5 @@ def filter_message_by_to_address(self, to_address):
         Retrieve a :obj:`Message` object by its `to` address.
         """
         for each_msg in self.message_store:
-            if each_msg.to == to_address[0]:
+            if each_msg.to == to_address:
```
ksheedlo: What's going on here?

glyph: Previously `to_address` was just being pulled straight out of `request.args`, which left the signature of this method somewhat nonsensical; `to_address` was a `list` of `bytes`.  Now it's actually a `text_type`, which makes more sense for it to be called `to_address` (i.e. not plural) and to be compared against `each_msg.to`, which is also a `text_type`.  There's only one caller and it's been adjusted to do the work of extracting the argument from `request.args` fully (pulling it out of the list, decoding it).

glyph: Also: thanks for having a look at this :)

ksheedlo: Someday everything will be Python 3+ and there will be types and things
like this will be caught at the first code review.

mimic/rest/fastly_api.py
```diff
@@ -11,6 +11,18 @@
...
+def text_urldata(urldata):
+    """
+    Convert ``urldata`` into text.
+
+    :param urldata: a dictionary mapping bytes to lists of bytes.
+
+    :return: a dictionary mapping text to lists of text.
+    """
+    return dict([(k.decode("utf-8"), [vv.decode("utf-8") for vv in v])
+                 for [k, v] in urldata.items()])
```
ksheedlo: Any reason you unpacked into a list here instead of a tuple?

ksheedlo: non-blocking: I think I saw this idiom somewhere else in the PR, should we consider refactoring it into a helper?

mimic/rest/loadbalancer_api.py
```diff
@@ -321,7 +323,7 @@ def delete_nodes_from_load_balancer(self, request, tenant_id, lb_id):
         """
         Deletes multiple nodes from a LB.
         """
-        node_ids = map(int, request.args.get('id', []))
+        node_ids = [int(node_id) for node_id in request.args.get(b'id', [])]
```
ksheedlo: :100:

mimic/test/test_maas.py
```diff
@@ -98,11 +113,10 @@ def createAlarm(self, label, entity_id, check_id):
         postdata['notification_plan_id'] = 'npTechnicalContactsEmail'
         alarms_endpoint = '{0}/entities/{1}/alarms'.format(self.uri, entity_id)
         req = request(self, self.root, b"POST", alarms_endpoint,
-                      json.dumps(postdata))
+                      json.dumps(postdata).encode("utf-8"))
         resp = self.successResultOf(req)
         self.assertEquals(resp.code, 201)
-        alarm_id = resp.headers.getRawHeaders('x-object-id')[0]
-        location = resp.headers.getRawHeaders('location')[0]
+        alarm_id, location = id_and_location(resp)
```
ksheedlo: :100: Thanks for refactoring this!

mimic/test/test_nova.py
```diff
@@ -71,14 +73,14 @@ def create_server(helper, name=None, imageRef=None, flavorRef=None,
...
-        body = json.dumps({"server": data})
+        body = json.dumps({"server": data}).encode("utf-8")
 
     create_server = request_func(
         helper.test_case,
         helper.root,
         b"POST",
         '{0}/servers'.format(helper.get_service_endpoint(
-            "cloudServersOpenStack", region)),
+            "cloudServersOpenStack", region)).encode("ascii"),
```
ksheedlo: Why `ascii` here?

mimic/test/test_nova.py
```diff
@@ -245,7 +247,7 @@ def test_create_server(self):
         Test to verify :func:`create_server` on ``POST /v2.0/<tenant_id>/servers``
         """
         self.assertEqual(self.create_server_response.code, 202)
-        self.assertTrue(type(self.server_id), unicode)
+        self.assertTrue(type(self.server_id), text_type)
```
ksheedlo: This assertion looks weird. Is it supposed to be

```python
self.assertTrue(type(self.server_id) == text_type)
```

or perhaps (2.7+)

```python
self.assertIs(type(self.server_id), text_type)
```

ksheedlo: How about this:

```python
self.assertTrue(type(self.server_id) is text_type)
```

mimic/test/test_nova.py
```diff
@@ -757,33 +780,49 @@ def test_unrescue(self):
...
         http://docs.rackspace.com/servers/api/v2/cs-devguide/content/Reboot_Server-d1e3371.html
         """
-        no_reboot_type_request = json.dumps({"reboot": {"missing_type": "SOFT"}})
+        no_reboot_type_request = json.dumps(
+            {"reboot": {"missing_type": "SOFT"}}).encode("utf-8")
         response, body = self.successResultOf(json_request(
             self, self.root, b"POST",
-            self.uri + '/servers/' + self.server_id + '/action', no_reboot_type_request))
+            self.uri.encode("ascii") + b'/servers/' +
+            self.server_id.encode("ascii") +
```
ksheedlo: Encode `ascii` for URLs/URIs?

mimic/test/test_resource.py
```diff
@@ -37,7 +37,7 @@ def one_api(testCase, core):
 
     :return: 2-tuple of ``region`` and ``service_id``
     """
-    service_id, api = core._uuid_to_api.items()[0]
+    service_id, api = next(iter(core._uuid_to_api.items()))
```
ksheedlo: :100:

mimic/test/test_resource.py
```diff
@@ -284,4 +283,4 @@ def test_verbose_logging(self):
         self.assertNotEqual(None, response_match)
         self.assertEqual(url, response_match.group('url'))
         headers = json.loads(response_match.group('headers'))
-        self.assertEqual(['application/json'], headers.get('Content-Type'))
+        self.assertEqual(['close'], headers.get('Connection'))
```
ksheedlo: What's going on here?

ksheedlo:
Code mostly looks good - I'd appreciate if you addressed the questions and had a look at the assertion in `test_nova.py`. I'm going to run Intelligence tests against this branch under py2 and py3 and report back.

ksheedlo:
Intelligence tests looks good running against Python 2. I need to go and install some things to make py3 work.

ksheedlo:
Intelligence tests are passing against py3.

ksheedlo:
LGTM :+1:

We can fix the assertion later if necessary.

mimic/rest/maas_api.py
```diff
@@ -2,7 +2,7 @@
 MAAS Mock API
 """
 
-from __future__ import division, unicode_literals
+from __future__ import division, unicode_literals, print_function
```
lekhajee: nit: do we need the `print_function`

ksheedlo: We don't need `print_function`.

glyph: We definitely don't need it, but this is something I add every time I add a print statement and then tend to leave it behind after I remove the prints; if we *do* end up using a print statement, it makes the code more consistent between 2/3, and since it's a syntactic import it doesn't trigger a py3 warning.

glyph: Would you like me to take it out just to reduce noise?

ksheedlo: +1 for removing noise. It makes no difference to me whether you use print statements for debugging or not, but we shouldn't have print statements in our production code and I see `print_function` and immediately think *someone used a print in this module!*

Logging is better than printing.

glyph:
Depends on #474 and #475; #475 because otherwise the tests won't pass, and #474 because I am having to do a lot of parallel coverage testing with `detox` to get this to pass and it is really annoying when every tox invocation runs `coverage erase` first.

mimic/rest/fastly_api.py
glyph: I didn't want to get into refactoring this stuff too much because really we ought to be using the shiny new `URL` object available in Twisted 15.5+ for this sort of thing rather than manually dealing with URL encoding ourselves; you're right that it's repeated though.

mimic/test/test_maas.py
glyph: no problem, it was just too tedious to retype again at the numerous call sites :)

mimic/test/test_nova.py
glyph: To be honest, this decision was somewhat arbitrary.  However, all these URLs are more or less guaranteed to be ASCII so it would be good to explode if something started surprising us by inserting non-ASCII code points in there.

mimic/test/test_nova.py
glyph: Good catch.  I think it should be `assertIsInstance`.

mimic/test/test_resource.py
glyph: Weird. It consistently fails on py27, pypy, and py34 if I put it back. I can't imagine how this ever got set before.  Will investigate.

glyph: NICE catch, @ksheedlo !  I typo'd `requestHeaders` in the implementation where it should have said `responseHeaders` and hence we were logging the wrong headers.  Really glad you were able to review this! :smile:

mimic/rest/fastly_api.py
glyph: Oh also, re: the "unpacked into a list" thing; unpacking is just into variables, so:

```python
x, y = 1, 2
[x, y] = 1, 2
(x, y) = 1, 2
```

are all semantically equivalent.  The reason I often choose square brackets rather than no brackets or parentheses is that they behave more consistently when you shrink the unpack to a size of 1, like so:

```python
unpackable = [1, 2]
[x, y] = unpackable
(x, y) = unpackable
# both x=1, y=2. but:
unpackable = [1]
[x] = unpackable # x = 1
(x) = unpackable # x = [1]
```
So especially when I'm bulk editing lots of code (as in this branch) I try to stick with square brackets in case I accidentally robotically remove the last element from some tuple I'm unpacking by accident; I either want to get an error (if the length is now wrong) or a consistent unpacking result (if I'm unpacking other 2-tuples and 3-tuples nearby).

Hopefully this makes sense? :)

mimic/test/test_nova.py
glyph: See above about "arbitrary"

mimic/test/test_resource.py
glyph: yeah this is a good idiom to know; you get a more sensible error, and you don't have to reify everything into a list before taking a value from it.

glyph:
So this is basically ready to land as per @ksheedlo's :+1: earlier; however, there are some simpler (quick to review) build configuration changes included here which would be nice to have someone just glance at separately in #475 and #474 respectively, so I'm going to hold off until those have landed.

glyph:
All dependencies merged, we've got a +1… landing!

glyph:
Thanks ever so much to @ksheedlo for all the reviews to get this in :).

PR #473, opened 2015-10-29
feat(AuthApi): list tenants for auth token

ksheedlo:
This Identity API is used in a specific login flow to Intelligence when a user needs to bypass SSO and log in directly by POSTing us a token. In this case, we need to know their tenant so that we can request their service catalog from Identity. This change adds the auth API that Intelligence needs to log in with just a token.

The API is documented here: https://developer.rackspace.com/docs/cloud-identity/v2/developer-guide/#list-tenants

### DIFF
--- a/mimic/rest/auth_api.py
+++ b/mimic/rest/auth_api.py
@@ -569,6 +569,33 @@ class AuthApi(object):
             prefix_for_endpoint=prefix_map.get)
         )
 
+    @app.route('/v2.0/tenants', methods=['GET'])
+    def list_tenants(self, request):
+        """
+        List all tenants for the specified auth token.
+
+        The token for this call is specified in the X-Auth-Token header,
+        like using the services in the service catalog. Mimic supports
+        only one tenant per session, so the number of listed tenants is
+        always 1 if the call succeeds.
+
+        For more information about this call, refer to the `Rackspace Cloud
+        Identity Developer Guide
+        <https://developer.rackspace.com/docs/cloud-identity/v2/developer-guide/#list-tenants>`_
+        """
+        try:
+            sess = self.core.sessions.existing_session_for_token(
+                request.getHeader(b'x-auth-token').decode('utf-8'))
+            return json.dumps({'tenants': [{'id': sess.tenant_id,
+                                            'name': sess.tenant_id,
+                                            'enabled': True}]})
+        except KeyError:
+            request.setResponseCode(401)
+            return json.dumps({'unauthorized': {
+                'code': 401,
+                'message': ("No valid token provided. Please use the 'X-Auth-Token'"
+                            " header with a valid token.")}})
+
 
 def base_uri_from_request(request):
     """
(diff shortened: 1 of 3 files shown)


ksheedlo:
Build failed due to another pypy/crypto fight. I thought we fixed that?

glyph:
@ksheedlo - we did fix it. Last time this happened there were a lot of interactions with caching; I am wondering if we have somehow triggered one of those again - have you restarted the build on travis or cleared the cache and gotten the same result?

glyph:
@ksheedlo - working on pypy/crypto fight in https://github.com/rackerlabs/mimic/pull/475

glyph:
Yeah there was a problem. It's fixed now.  Not sure how that landed; the build cache might have been a bad state for the build that broke it, I guess.  You can try merging #475 in here and it should make the build go green.

ksheedlo:
This PR is passing Python 3 but Travis is still allowing Python 3 builds to fail; we should merge #478 as soon as we can so that checking Python 3 compatibility isn't something we have to do manually.

glyph:
Sorry for the review lag here.

This looks mostly okay, but I do have a couple of feedback points:

I tend to dislike methods whose names begin with `get_`.  Any method or function with a return value is "getting" something, which makes it somewhat vacuously true and therefore inconsistently applied (unless you apply it to absolutely everything, in which case it's just ugly).  Also, we now have 3 methods: `session_for_api_key`, `session_for_username_password`, and `get_or_create_session_for_token` which all retrieve an existing session or create a new one, only the last of which has the `get_or_create_` prefix.

However, It looks like the semantic change to `session_for_token` was already done outside of this branch (when, I wonder?) so it's great that you're changing the docs.

This also leaves `lookup_session_for_token` the odd method out; rather than calling it "lookup", how about just `existing_session_for_token`?

If you agree with my suggestion for naming here – to be clear, that would be to remove the re-naming of `session_for_token` and to change `lookup_session_for_token` to `existing_session_for_token` – then go ahead and merge.  If not, feel free to push back and state your reasoning, I'm open to being convinced otherwise.

If it were just the question of this one name, I'd say that this is just aesthetic feedback and go ahead and merge either way, but I actually think that keeping a consistent naming convention is a pretty important thing; plus, this is going to be part of the public API of mimic when we open up to external plugins, which will probably be quite soon, so I think we should have a pretty good reason to break consistency and change the name.

ksheedlo:
> If you agree with my suggestion for naming here – to be clear, that would be to remove the re-naming of session_for_token and to change lookup_session_for_token to existing_session_for_token

:+1:

> It looks like the semantic change to session_for_token was already done outside of this branch (when, I wonder?)

Answering this for historical interest. Intelligence needs the real semantics because our tests often integrate Mimic against a real SSO Identity Provider. IIRC, this went back and forth a couple of times before we established integration tests for this case, since probably no one else is going to POST a real Identity token to Mimic and expect to get a session. The docs are most likely a holdover from an earlier version of the method.

PR #474, opened 2015-10-31
Improve tox configuration for better behavior with respect to coverage

glyph:
In this PR I plan to:

- upgrade coverage.py
- make coverage reports accurate when tests are run in parallel with [`detox`](https://pypi.python.org/pypi/detox)
- move coverage reporting out of `tox` itself to the travis script so that it's easier for users to get local combined coverage with `coverage combine` and `coverage html`.
- stop running the tests in the source directory, so that we can tell if the `setup.py` is correct

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -51,7 +51,7 @@ script:
   - ./.travis/run.sh
 
 after_success:
-  - "source ~/.venv/bin/activate && codecov"
+  - "source ~/.venv/bin/activate && coverage combine && codecov"
 
 notifications:
   irc:
--- a/.travis/install.sh
+++ b/.travis/install.sh
@@ -50,6 +50,7 @@ else
     python -m virtualenv ~/.venv
     source ~/.venv/bin/activate
     pip install tox codecov
+    coverage erase
     tox --recreate --notest
 
     # If "installdeps" fails, "tox" exits with an error, and the "set -e" above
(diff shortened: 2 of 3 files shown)


glyph:
OK it looks like running tests outside of the source directory, because `envsitepackagesdir` is not an available substitution value in the `changedir` field.

glyph:
So I've just removed coverage reporting entirely, which should make builds slightly faster (no need to do coverage work) because if you want to see coverage information you should use `coverage html` which is *much* nicer.

glyph:
This depends on #475 so please review that first.

glyph:
Apparently not RFR, since this apparently breaks reporting to codecov.  I thought it might but the green checkmark mislead me.  Let me try that again.

glyph:
That coverage report looks like a @codecov-io bug.  Clicking on the "coverage diff" link - https://codecov.io/github/rackerlabs/mimic/compare/b77965e571501646894736f23282df67eb081149...49cbfca038c2c760e13005593ce8d92d1f8b8fee - no diff shows up.  Plus, *the code didn't change* so I'm not sure what the problem could be.  Maybe someone landed something on `master` without looking at coverage and this is going off an older build?

glyph:
OK, back in review since now coverage reports are happening again.

ksheedlo:
:+1:

PR #475, opened 2015-10-31
fix cryptography/pypy version match on travis

glyph:
Wheel all the requirements, as specified in the requirements.txt, instead of treating cryptography specially.


glyph:
Added _all_ the "needed by" labels because this is blocking literally everything :)

ksheedlo:
:+1:

PR #478, opened 2015-11-01
chore(travis): remove python3.4 from allow_failures

ksheedlo:
Now that we have landed support for Python 3, it's important not to
regress. This change removes Python 3 from the `allow_failures` section
in Travis so that builds will not be green while failing Python 3.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,7 +42,6 @@ matrix:
           env: MACAPP_ENV=system
     allow_failures:
       - env: TOXENV=docs-linkcheck
-      - env: TOXENV=py34
 
 install:
   - travis_retry ./.travis/install.sh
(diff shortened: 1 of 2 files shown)


ksheedlo:
@glyph I wonder why this wasn't done in #471. Are you waiting on dependencies to cut a release?

glyph:
@ksheedlo - yeah, I don't feel like we can gate Mimic releases on trunk builds of other things; that means we'll get blocking test failures on unrelated changes if another project introduces a bug.  However, I guess we don't need to wait for a release - can you add git hashes to the deps in tox.ini so that we have our py34 support pinned to predictable hashes, then file separate tickets for upgrading to released versions when they come out?

ksheedlo:
Updated `tox.ini` and filed #479, #480, #481.

glyph:
Thanks for the fix and the issues.  LGTM.  Looking forward to turning those hashes into actual versions :).

PR #483, opened 2015-11-04
Start dns plugin

tawalton:
This is the beginning of the dns plugin.  It has one route for reverse dns, which addresses issue, https://github.com/rackerlabs/mimic/issues/485


glyph:
Looks pretty straightforward - although I hope more is coming :).

PR #484, opened 2015-11-05
chore(test_maas): speed up slow test

ksheedlo:
This test created 100 MaaS entities and listed them 100 times, for no
apparent reason.  Split and redone with a much faster strategy for
testing pagination and markers.


glyph:
LGTM, thanks for paying attention to test suite performance, everyone who has to wait for travis on rackerlabs thanks you ;)

PR #487, opened 2015-11-08
refactor(queue_api): use model classes

ksheedlo:


ksheedlo:
This is out ahead of some feature work I'm doing in Intelligence that may benefit from having more parts of Queues to test with, but not strictly needed by anyone that I know of right now.

In particular I want to add the ability to post and claim messages. I'm also thinking about having messages and claims ttl out similar to how they do in the real Queues API, but with Mimic's clock.

ksheedlo:
> ERROR: InvocationError: '/Users/travis/build/rackerlabs/mimic/.tox/py27/bin/coverage run -p /Users/travis/build/rackerlabs/mimic/.tox/py27/bin/trial --rterrors mimic'

Looks like coverage crashing. Note this only failed on the OSX Python 2.7 job. It passes Python 2.7 on my Mac.

mimic/model/queue_objects.py
```diff
@@ -0,0 +1,61 @@
...
+    """
+    name = attr.ib(validator=attr.validators.instance_of(text_type))
+    id = attr.ib(validator=attr.validators.instance_of(text_type),
+                 default=attr.Factory(lambda: random_hex_generator(4)))
+    _messages = attr.ib(default=attr.Factory(list))
+
+    def brief_json(self):
+        """
+        A brief representation of this queue that can be serialized
+        via json.dumps.
```
glyph: This generates "brief" JSON that shows up in a specific place in the API; it might be nice to document that.

glyph:
Definitely a net improvement so I'm not going to block on my minor docstring feedback; please feel free to submit a follow-up PR to insert more API references if you think that would be helpful.

Issue #485, opened 2015-11-05
Plugins needed for Reach to run its automation


tawalton:
- [ ] DNS plugin - rdns call

- [ ] Support plugin - support-accounts call

- [ ] Volumes plugin - get volumes call

- [ ]Glance plugin - multiple routes

- [ ] Networks plugin - just need the endpoint in the catalog (flavor does not load on 												             server details page)

core.py -  MimicCore init method need to remove the random_hex_generator and hard    			code the hex for internal apis

BenjamenMeyer:
Perhaps for the core.py entry, a better solution may be to convert from having the internal APIs use the random_hex_generator at all, and make a variant of the `ExternalApiStore` (from #614) for Internal APIs - or rather, just register all the internal APIs as templates with an `ExternalApiStore` and then enable them by default. There is no enforcement over what the Service-ID should be; docs suggest as UUID. Reach shouldn't be reliant on the values any way to look up services.

PR #489, opened 2015-11-09
start mock for block storage

tawalton:
This is the beginning of the cinder plugin. It has one route to get the list of volumes, which addresses issue, #485


lekhajee:
+1

Issue #488, opened 2015-11-09
travis build failing on mac 27 environment as of this weekend


glyph:
@tawalton discovered this with his networking PR https://github.com/rackerlabs/mimic/pull/486 - but rebuilding the last successful build of master fails in the same way.

glyph:
I believe we are seeing this issue: https://github.com/pyca/cryptography/issues/2474

glyph:
Yep that was the problem, rebuilding now works; master is green again.

glyph:
Cryptography fixed this upstream.

PR #490, opened 2015-11-10
bump pypy and cpython versions to catch up to date

glyph:
Update CPython and PyPy versions in travis configuration.

Hopefully this will do something to ameliorate #488 because I cannot imagine why that is happening right now...


glyph:
OK this is not a good way to do this, probably.  And #488 is fixed separately.

PR #491, opened 2015-11-11
feat(queue_api): list and post messages

ksheedlo:
This change adds the ability to post messages to a queue and list them
back. It also expires messages after their TTL. Making claims and
deleting messages is not supported yet.

### DIFF
--- a/mimic/model/queue_objects.py
+++ b/mimic/model/queue_objects.py
@@ -6,8 +6,44 @@ from __future__ import absolute_import, division, unicode_literals
 
 import attr
 from six import text_type
+from twisted.internet.interfaces import IReactorTime
 
 from mimic.util.helper import random_hex_generator
+
+
+@attr.s
+class Message(object):
+    """
+    A Message object in Cloud Queues.
+    """
+    ttl = attr.ib(validator=attr.validators.instance_of(int))
+    body = attr.ib(validator=attr.validators.instance_of(dict))
+    queue_name = attr.ib(validator=attr.validators.instance_of(text_type))
+    posted_by = attr.ib(validator=attr.validators.instance_of(text_type))
+    posted_at = attr.ib(validator=attr.validators.instance_of(int))
+    id = attr.ib(validator=attr.validators.instance_of(text_type),
+                 default=attr.Factory(lambda: random_hex_generator(12)))
+
+    def to_json(self, current_time):
+        """
+        A representation of this message that can be serialized via json.dumps.
+        """
+        return {'body': self.body,
+                'age': current_time - self.posted_at,
+                'href': self.href(),
+                'ttl': self.ttl}
+
+    def href(self):
+        """
+        Returns the URL path representing this message.
+        """
+        return '/v1/queues/{0}/messages/{1}'.format(self.queue_name, self.id)
+
+    def is_expired_at(self, current_time):
+        """
+        Returns True if the message is expired at the given time.
+        """
+        return (self.posted_at + self.ttl) < current_time
 
 
 @attr.s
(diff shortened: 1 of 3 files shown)


ksheedlo:
This is breaking on my local Python 3, but I'm not sure why.

glyph:
@ksheedlo The problem is that you're passing text to `json_request`.  `json.dumps` returns `str` on both pythons, so `bytes` on py2 and `six.text_type` on py3.  `json_request` serializes anything passed to it that isn't bytes.  So on the py3 builder you're getting a double-serialized JSON string.

Generally, if you're using `json_request`, don't cal `json.dumps` yourself; let it do the serialization for you.

glyph:
(This is all in `test_queue.py`)

ksheedlo:
Still broken on py3, but it's down to 1 erroring test now. Here's the stack trace:

```
Traceback (most recent call last):
  File "/Users/ken/rackerlabs/mimic/mimic/test/test_queue.py", line 126, in test_same_client_message_shows_with_echo
    headers={b'Client-ID': [b'client-1']}))
  File "/Users/ken/rackerlabs/mimic/.tox/py34/lib/python3.4/site-packages/twisted/trial/_synctest.py", line 704, in successResultOf
    deferred, result[0].getTraceback()))
twisted.trial.unittest.FailTest: Success result expected on <Deferred at 0x10dd58a90 current result: None>, found failure result instead:
Traceback (most recent call last):
  File "/Users/ken/rackerlabs/mimic/mimic/test/helpers.py", line 210, in get_body
    body_d = treq.json_content(response)
  File "/Users/ken/rackerlabs/mimic/.tox/py34/lib/python3.4/site-packages/treq/content.py", line 100, in json_content
    d.addCallback(json.loads)
  File "/Users/ken/rackerlabs/mimic/.tox/py34/lib/python3.4/site-packages/twisted/internet/defer.py", line 317, in addCallback
    callbackKeywords=kw)
  File "/Users/ken/rackerlabs/mimic/.tox/py34/lib/python3.4/site-packages/twisted/internet/defer.py", line 306, in addCallbacks
    self._runCallbacks()
--- <exception caught here> ---
  File "/Users/ken/rackerlabs/mimic/.tox/py34/lib/python3.4/site-packages/twisted/internet/defer.py", line 588, in _runCallbacks
    current.result = callback(current.result, *args, **kw)
  File "/Users/ken/.pyenv/versions/3.4.3/lib/python3.4/json/__init__.py", line 318, in loads
    return _default_decoder.decode(s)
  File "/Users/ken/.pyenv/versions/3.4.3/lib/python3.4/json/decoder.py", line 343, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
  File "/Users/ken/.pyenv/versions/3.4.3/lib/python3.4/json/decoder.py", line 361, in raw_decode
    raise ValueError(errmsg("Expecting value", s, err.value)) from None
builtins.ValueError: Expecting value: line 1 column 1 (char 0)
```

glyph:
Did you rebase or something?  Pulling from this PR is failing for me.

ksheedlo:
Amended and force pushed. Try `git branch -f <local-ref> <remote-ref>`

glyph:
The problem you're having now is that the values in `request.args` are bytes, and `list_messages_for_queue` is comparing `request.args.get` to `b'true'`, which is effectively a type error but unfortunately not an exception.  So you're getting an empty response_body with a 204 response code.

ksheedlo:
Updated, should work now.

glyph:
Tests are passing but coverage is down; if you could address that when you get back that would be great.

ksheedlo:
Update: Fixed coverage.

glyph:
Looking good!

PR #494, opened 2015-11-30
newest twisted no longer supports python 2.6.

glyph:


tawalton:
:ship:

glyph:
thanks @tawalton

PR #496, opened 2015-12-02

glyph:


glyph:
Closing and deleting so @requires will resync with the better dependency organization on master.

PR #497, opened 2015-12-03
reorganize everything into requirements .txt files

glyph:
also:

- update the documentation

- encourage the use of 'tox' everywhere for a uniform development experience


glyph:
This should allow @requires to submit more well-formed PRs so we can just merge them if they pass.

.travis/install.sh
```diff
@@ -3,60 +3,70 @@
...
-    # temporary pyenv installation to get pypy-2.6 before container infra upgrade
-    if [[ "${TOXENV}" == "pypy" ]]; then
-        git clone https://github.com/yyuu/pyenv.git ~/.pyenv
-        PYENV_ROOT="$HOME/.pyenv"
-        PATH="$PYENV_ROOT/bin:$PATH"
-        eval "$(pyenv init -)"
-        pyenv install pypy-2.6.1
-        pyenv global pypy-2.6.1
+if [[ -n "${PYENV_PYTHON}" ]]; then
+    if [[ "$DARWIN" = true ]]; then
```
radix: It seems weird that sometimes you protect some `brew` commands with a check for `$DARWIN == true` and sometimes with a check for `"${MACAPP_ENV}" == "system"`.

ALSO it seems weird to me that the `true` here isn't in doublequotes. does that mean something special?

glyph: this weirdness was all here before the change, it's somewhat less inconsistent after the change :).  The mac app env is a particularly weird edge case that has to work around some bugs (you can see the app-building shell script for an explanation).

I think the `"true"` thing is maybe a problem though, I'll look at that.

glyph: eh, it's not actually a problem but it is a syntax inconsistency so I'll fix it.

radix:
LGTM. Feel free to merge after the build passes.

glyph:
Thanks for the review @radix ! :)

PR #498, opened 2015-12-03
upgrade code quality tools

glyph:
linter, documentation, etc.

this depends on #497 and is an example of how upgrading dependencies will be more straightforward once that lands.


glyph:
a more useful diff to look at, in the meanwhile: https://github.com/glyph/mimic/compare/requirements-sanity...glyph:better-linter

glyph:
The correct diff is https://github.com/rackerlabs/mimic/compare/master...glyph:better-linter

tawalton:
just looking at the the link provided above -- LGTM

glyph:
thanks @tawalton !

PR #499, opened 2015-12-09

glyph:


glyph:
Awesome that the tests passed!  However, I think this indicates a need for an additional check: that the requirements files are not stomped on by `setup.py`.  In the case of this PR, `testtools` is pinned by `setup.py` and upgrading it in the `production.txt` has no effect.

glyph:
Looks like no test failures, so, let's do it!

glyph:
Oops.  My previous comment about `testtools` is still the case.

glyph:
Closing to ask requires.io to try again.

PR #500, opened 2015-12-18
Support mocks for top-level domains.

### LINKED ISSUE #108: Support for non Openstack services

glyph:
Fixes #108


ksheedlo:
I like this! Will try to look in detail at some point this week.

glyph:
Cool, thanks.  I should really investigate why the builds are failing...

glyph:
Now that tests and coverage are on par, time to land :).  Please review!

glyph:
@ksheedlo bump

mimic/imimic.py
```diff
@@ -9,7 +9,8 @@
 
 class IAPIMock(Interface):
     """
-    An :obj:`IAPIMock` provides an API.
+    An :obj:`IAPIMock` provides an API via an entry in a tenant's service
+    catalog.
```
ksheedlo: Depending on how general the use case becomes it may make sense soonish to rename this such that it explicitly references Openstack Identity. Lots of APIs exist that don't have the same notion of a tenant or service catalog.

mimic/imimic.py
```diff
@@ -47,3 +48,36 @@ def get_session(session_store):  # pragma:nocover
...
+        """
+        The resource which would appear at the root of the domain.  For
+        example, if you were mocking an API typically present at
+        ``https://api.foo.example.com/v2/``, this should return a resource
+        whose ``.getChild("v2", ...)`` returns the root of the API.
+
+        :return: The resource for the root of the API implemented by this
+                 :obj:`IAPIDomainMock`.
+        :rtype: :obj:`twisted.web.iweb.IResource`
+        """
```
ksheedlo: :100:

ksheedlo:
:+1: to me

mimic/imimic.py
glyph: This is a good point; I think I'll have a PR for a final pass before we open up 3rd-party plugins, and this renaming should be part of it.

glyph:
Thanks!  I'll go ahead and land.

mimic/imimic.py
glyph: (I am assuming you meant "lots of APIs exist that *do* have the same notion…"?)

PR #501, opened 2015-12-18
prep for 2.0 release

### LINKED ISSUE #318: drop python 2.6 support

glyph:
- add python 3.4 and 3.5 support officially, using released versions
- drop python 2.6 support (fixes #318)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,6 +15,8 @@ matrix:
           env: TOXENV=py27
         - python: 3.4
           env: TOXENV=py34
+        - python: 3.5
+          env: TOXENV=py35
         - python: pypy
           env: TOXENV=pypy
         - python: 2.7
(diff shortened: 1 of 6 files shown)


lekhajee:
+1

PR #506, opened 2016-01-21
Add stack update, preview, and action endpoints

dragorosson:


mimic/rest/heat_api.py
```diff
@@ -75,6 +75,9 @@ class HeatRegion(object):
     Rest endpoints for mocked Heat API.
     """
 
+    HEAT_URL = '/v1/<string:tenant_id>'
```
manishtomar: Non-blocking: Would you consider renaming this to `TENANT_URL`? I would consider `HEAT_URL` to be `.../stacks` which you've already correctly named as `STACK_URL` :)

dragorosson: I like that idea. Done.

mimic/test/test_heat.py
```diff
@@ -84,6 +84,31 @@ def create_stack(self, stack_name, tags=None):
...
         return data
 
+    def update_stack(self, stack_name, stack_id, resp_code=202):
+        """
+        Request stack update and assert that the response matched the one
+        provided.
+        """
+        req_body = {'foo': 'bar'}
+        req = request(self, self.root, b"PUT",
+                      '%s/stacks/%s/%s' % (self.uri, stack_name, stack_id),
```
manishtomar: Non-blocking: How about using `str.format`?

mimic/model/heat_objects.py
```diff
@@ -149,6 +151,38 @@ def default_create_behavior(collection, request, body, absolutize_url):
     return json.dumps(response)
 
 
+@stack_update.declare_default_behavior
+def default_update_behavior(collection, request, stack_name, stack_id):
+    """
+    Successfully update a stack as long as it exists. Updates the stacks status.
+    """
+    stack = collection.stack_by_id(stack_id)
```
manishtomar: Non-blocking suggestion: Code to check stack based on id and returning 404 can be in a decorator.
```python
def stack_exists(f):
    def decorator(collection, request, stack_name, stack_id):
        stack = collection.stack_by_id(stack_id):
        if not stack:
            request.setResponseCode(404); return b''
        return f(collection, request, stack)
    return decorator

@stack_exists
@stack_update.declare_default_behavior
def default_update_behavior(collection, request, stack):
...
```
This is merely a suggestion. Feel free to ignore :)

mimic/test/test_heat.py
```diff
@@ -84,6 +84,31 @@ def create_stack(self, stack_name, tags=None):
         data = self.get_responsebody(resp)
         return data
 
+    def update_stack(self, stack_name, stack_id, resp_code=202):
+        """
+        Request stack update and assert that the response matched the one
+        provided.
```
glyph: Given that this is not a `@route` (and therefore its parameters are not documented by external documentation) would you mind adding `:param:` fields for `stack_name`, `stack_id`, and `resp_code`?  I'd really like Mimic to have a higher standard of internal API docs.

mimic/test/test_heat.py
```diff
@@ -84,6 +84,31 @@ def create_stack(self, stack_name, tags=None):
...
+                      body=json.dumps(req_body).encode("utf-8"))
+        resp = self.successResultOf(req)
+        self.assertEqual(resp.code, resp_code)
+
+    def check_stack(self, stack_name, stack_id, resp_code=201,
+                    req_body={'check': None}):
+        """
+        Request stack check and assert that the response matched the one
+        provided.
+        """
```
glyph: See above re: `:param:`

mimic/rest/heat_api.py
```diff
@@ -138,9 +141,39 @@ def extract_tags():
...
+        """
+        region_collection = self._region_collection_for_tenant(tenant_id)
+        return region_collection.request_update(request, stack_name, stack_id)
+
+    @app.route(STACK_URL + '/actions', methods=['POST'])
+    def perform_action(self, request, tenant_id, stack_name, stack_id):
+        """
+        Performs an action on a stack. Only action-check is enabled at
+        Rackspace.
+        See http://api.rackspace.com/api-ref-orchestration.html#stack_update
```
manishtomar: This links to stack update doc. Moreover, I couldn't find actions doc in api.rackspace.com or developer.rackspace.com. I wonder if you should link to http://developer.openstack.org/api-ref-orchestration-v1.html#stack-actions.

mimic/test/test_heat.py
```diff
@@ -159,6 +184,68 @@ def test_list_stacks_two(self):
         self.assertEqual(set(stack['stack_name'] for stack in two_stack_list),
                          set(['foostack', 'barstack']))
 
+    def test_check_stack(self):
+        """
+        Stack updates with correct check status.
+        """
+        foo_resp = self.create_stack('foostack')
+        self.check_stack('barstack', foo_resp['stack']['id'])
```
manishtomar: Why is this done with `barstack` instead of `foostack`? I am guessing it should fail but thats not a concern since its not really checking.

mimic/test/test_heat.py
```diff
@@ -159,6 +184,68 @@ def test_list_stacks_two(self):
...
+        Supplying more than one action in stack-action requests generate 400.
+        """
+        foo_id = self.create_stack('foo')['stack']['id']
+        self.check_stack('foo', foo_id,
+                         req_body={'check': None, 'cancel_update': None},
+                         resp_code=400)
+
+    def test_disabled_actions(self):
+        """
+        Stack updates with correct check status.
```
manishtomar: The doc doesn't seem to match. How about "Performing unimplemented actions results in 405"

mimic/test/test_heat.py
```diff
@@ -84,6 +84,31 @@ def create_stack(self, stack_name, tags=None):
...
+        provided.
+        """
+        req_body = {'foo': 'bar'}
+        req = request(self, self.root, b"PUT",
+                      '%s/stacks/%s/%s' % (self.uri, stack_name, stack_id),
+                      body=json.dumps(req_body).encode("utf-8"))
+        resp = self.successResultOf(req)
+        self.assertEqual(resp.code, resp_code)
+
+    def check_stack(self, stack_name, stack_id, resp_code=201,
```
manishtomar: This is called below to perform other actions. How about renaming it to `perform_stack_action` instead?

manishtomar:
LGTM modulo comments.

Thank you soo much for doing this :)

mimic/test/test_heat.py
dragorosson: Done.

mimic/rest/heat_api.py
dragorosson: Done. It's true, there's no documentation for stack actions in api.rackspace.com yet, so I linked to the OpenStack docs.

mimic/test/test_heat.py
dragorosson: My mistake when copy-pasting. Done.

mimic/test/test_heat.py
dragorosson: Another copy-paste mistake. Done.

mimic/test/test_heat.py
dragorosson: Good idea, I should have renamed it after I refactored what it did. Done.

mimic/test/test_heat.py
dragorosson: Done, for the other helper methods too.

mimic/test/test_heat.py
dragorosson: Done.

dragorosson:
@glyph @lekhajee @manishtomar Looks like the build succeeded except for the  MACAPP_ENV=system test, which doesn't look related to the PR and is failing on other PRs as well.

glyph:
@dragorosson - yes, working on the mac app failures as we speak. Something must have come un-pinned.

dragorosson:
Excellent, thanks @glyph!

mimic/model/heat_objects.py
```diff
@@ -9,6 +9,27 @@
 from mimic.model.behaviors import BehaviorRegistryCollection, EventDescription
 
 
+def get_stack_or_404(f):
+    """
+    Decorator which retrieves a stack from a collection and adds it to the
+    function's arguments if it exists. If the stack does not exist, a 404
```
glyph: This is slightly misleading, since it modifies the signature and doesn't document the modified signature.

`do_something_behavior` takes `collection, request, stack`, and it says It "retrieves a stack from a collection and adds it to the function's arguments".  but it doesn't say that it *removes* the stack name and stack ID from the argument signature.

I generally don't like decorators which mangle function signatures this way, because they break one of the more basic ways that Python programmers read code: you look at a function and you call it with the arguments you see in the argument list.  I think this might make sense as a utility function.

mimic/model/heat_objects.py
```diff
@@ -9,6 +9,27 @@
...
+
+    Use like so::
+
+        @get_stack_or_404
+        def do_something_behavior(collection, request, stack):
+            ...
+    """
+    def decorator(collection, request, stack_name, stack_id):
+        stack = collection.stack_by_id(stack_id)
+        if not stack:
```
glyph: Should this really be `stack is None` rather than `not stack`?  are there other untruthy values other than `None` we might compare to here?

mimic/model/heat_objects.py
glyph: If you're going to leave it as a decorator - not a dealbreaker for me, just not my preference - make sure the docstring shows *both* how the function is declared *and* how it is called, and explain the distinction.

glyph:
@dragorosson - now that the test issues are dealt with, I've popped the label off and reassigned to you. Please go ahead and switch the labels again when you've dealt with the remaining feedback.

mimic/model/heat_objects.py
dragorosson: I don't really like how it changes the function's signature either. I reverted it because I have another idea of how to do it, but I'll have to try it later.

mimic/model/heat_objects.py
dragorosson: Done. No, it should either be a instance or None.

mimic/model/heat_objects.py
glyph: I guess I gave the anti-suggestion to this afterward :)

mimic/model/heat_objects.py
```diff
@@ -135,6 +135,8 @@ def creation_response_json(self, absolutize_url):
 
 
 stack_creation = EventDescription()
+stack_check = EventDescription()
```
glyph: I don't like the fact that we have these new event descriptions with no explanation of what they mean or what their behavior signatures are.  However, I don't want to block on that, given that there's no clear existing convention for doing this, there are other events to document, and it's not clear how to document the control-plane endpoint associated with them.  I think probably `:var:` fields in the module docstring would do it.

So would you mind opening a new issue for documenting these event descriptions in the heat mimic?

glyph:
Thanks for doing this work, @dragorosson .  If you could open that new documentation issue that would be great.

mimic/model/heat_objects.py
dragorosson: Honestly, after seeing how sparingly EventDescriptions are used elsewhere in mimic and that default behaviors are the only ones used, I think that refactoring them out of the heat mimic would be better.

glyph: There are definitely other places within mimic (the nova mock springs to mind) where non-default behaviors are used.  We want *more* usage of behaviors, generally, not less…

dragorosson: Oh, I didn't mean that they shouldn't be used or are not useful, just that they are not needed here yet.

glyph: It feels to me like removing the behaviors here would be moving backwards, and we should instead be adding in error behaviors and a control plane.

Issue #505, opened 2016-01-21
docker build failed


nyasukun:
this Dockerfile is from python:2.7-onbuild.
However python:2.7-onbuild requires "requirements.txt."
It seems requirements.txt was deleted at this commit 76175faef4f5ec0385bd3bef8c6c2005ec2d98c9
Since this, docker build always failed.

In my environment, it is successful to build with creating a link "requirments.txt" to "requirements/production.txt."

glyph:
Thanks for reporting this issue.  Clearly we need to be putting the docker build into our CI...

nyasukun:
Is there any reason not to add requiments.txt as a link at root?

glyph:
We have three options, as I see it:

1. A symlink, which would cause problems for Windows users,
2. A copy, which would work fine for Windows users, but create duplication,
3. Move the file, which means we manage requirements files inconsistently (some in `requirements/`, some not)

I guess 3. is probably the least-bad option.

nyasukun:
I have another option:
* Modify Dockerfile to use the file under the ```requirments/```

see:
https://github.com/rackerlabs/mimic/compare/master...nyasukun:dockerfile_without_onbuild?expand=1

glyph:
@nyasukun That looks pretty reasonable. Want to submit a PR?

nyasukun:
@glyph Yes. I submitted a PR #507.

derwolfe:
Would adding a docker build step on travis be sufficient for adding this to CI?

glyph:
@derwolfe - those two things are bascially synonyms, so yes.

derwolfe:
@glyph sorry, that question was a bit unclear on my part. What I meant was, is the only concern that the image builds or should there be some tests run against the image?

glyph:
@derwolfe - oh. well it would be nice to just get the build, for starters :). but ideally it would be just like the Mac app build; build the image, run the tests in the context of the thing that was just built

nyasukun:
@glyph It seems to be merged. Can you close this issue?

glyph:
Oops, I guess the PR didn't have a "fixes" in its description.  The word "solves" is not similarly magic :)

nyasukun:
oh. sorry. I'm new in github and travis.

lvh:
... It seems like this should have been closed, so I'll go ahead and close it :)

glyph:
Yep! Thanks for noticing, @lvh .

PR #507, opened 2016-01-23
Modify Dockerfile in order to build docker image successfully

nyasukun:
to solve #505.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,3 +1,12 @@
-FROM python:2.7-onbuild
+FROM python:2.7
+
+RUN mkdir -p /usr/src/app
+WORKDIR /usr/src/app
+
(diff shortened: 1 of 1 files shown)


nyasukun:
All checks failures seem unrelated things with this change.

glyph:
I can't force travis to rebuild this branch, unfortunately (I've reported the issue to travis support).  Can you push an empty commit so we can get a green checkmark and proceed?  Thanks!

glyph:
> I can't force travis to rebuild this branch, unfortunately (I've reported the issue to travis support). Can you push an empty commit so we can get a green checkmark and proceed? Thanks!

Looks like my re-build request went through after waiting about an hour. No need to push that empty commit!

PR #508, opened 2016-01-26
Correct mac app build failures.

glyph:
`setuptools` was automatically upgraded on our build infrastructure, which caused the Mac app build to start failing.

The root cause here is that `pkg_resources`, within setuptools, started doing some dynamic-importing tricks in `pkg_resources._vendor`, which in turn caused these modules to stop getting detected as necessary by `modulegraph`.  `treq` depends on these modules, because it uses `pkg_resources.resource_string`, which does the aforementioned dynamic importing.

The solution is to add some static imports of these internal modules to the main script, which informs `modulegraph` of their necessity, causing them to be included in the app bundle properly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -182,4 +182,3 @@ docker build -t mimic . && docker run --restart=no --rm=true -p 8900:8900 mimic
 ```
 
 This will expose Mimic on port 8900, so you can access it directly from the host. The default port exposure is intended for communication between containers; see the Docker documentation for more information. If you're using `boot2docker`, run `boot2docker ip` to find the right IP.
-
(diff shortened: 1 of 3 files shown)


glyph:
Looks like the issue is that `setuptools` isn't pinned, because it comes along with `pip` which upgrades automatically on travis / homebrew.  This adds some hackery to the main script to deal with the current version, and pins it to the current version as well.

glyph:
I'm going to set this as RFR while we wait for travis, since I reproduced the problem locally and this fix totally fixed it.

glyph:
Since this is blocking all development I'm just going to toss all the "needed by" labels on there :)

tawalton:
LGTM :shipit:

derwolfe:
LGTM as well :sunglasses:

glyph:
Thanks for the +1s. In the future you can just hit the merge button yourselves :)

PR #514, opened 2016-01-30

ziadsawalha:


glyph:
Superseded by #515 - I thought this would be automatically closed when the tip of that branch was merged, but I guess I don't know how github really works.

The changes in #515 should also make it possible to automatically merge future mechanical changes from requires.io, since it does the extra pinning required for the new, broken version of testtools, and now has checks to ensure `setup.py` is satisfied by `requirements.txt`.

PR #515, opened 2016-01-30
Upgrade requirements and fix docs build failure

glyph:
This should address both the packages that are out of date, and also the potential mismatch between `setup.py` and `requirements.txt`.

### DIFF
--- a/.travis/install.sh
+++ b/.travis/install.sh
@@ -83,7 +83,9 @@ done;
 # https://github.com/pypa/pip/issues/3414
 
 for fn in .wheels/*pp27*; do
-    mv -v "${fn}" "$(echo "${fn}" | sed -e s/pp27/pp240/g)";
+    if [ "${fn}" != '.wheels/*pp27*' ]; then
+        mv -v "${fn}" "$(echo "${fn}" | sed -e s/pp27/pp240/g)";
+    fi;
 done;
 
 # If "installdeps" fails, "tox" exits with an error, and the "set -e" above
--- a/.travis/install.sh
+++ b/.travis/install.sh
@@ -72,6 +72,7 @@ tox --version;
 export PIP_WHEEL_DIR=".wheels";
 export PIP_FIND_LINKS=".wheels";
 
+pip install wheel;
 for req in requirements/*.txt; do
     pip wheel -r "${req}";
 done;
--- a/.travis/install.sh
+++ b/.travis/install.sh
@@ -80,8 +80,6 @@ for req in requirements/*.txt; do
     ./.tox/"${TOXENV}"/bin/pip wheel -r "${req}" || true;
 done;
 
-PIP_NO_INDEX="yes" tox --recreate --notest;
-
 # If "installdeps" fails, "tox" exits with an error, and the "set -e" above
 # causes it to retry.  If "inst" fails, however, no error is reported for some
 # reason.  The following line causes "grep" to exit with error (and thanks to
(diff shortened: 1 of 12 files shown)


glyph:
shout out to some great pip issues we hit on this one:

- https://github.com/pypa/pip/issues/3414
- https://github.com/pypa/pip/issues/988

.travis/install.sh
```diff
@@ -62,8 +62,32 @@ hash -r;
 
 coverage erase;
 tox --version;
+
 tox --recreate --notest;
 
+# Pip has no dependency resolver: https://github.com/pypa/pip/issues/988.  This
+# means that if requirements.txt conflicts with setup.py, we won't find out
+# about it from tox alone.  Therefore, let's ensure that we give tox *only* the
+# verisons of the wheels specified in the pinned dependencies, and ensure that
```
ksheedlo: nit: versions

.travis/install.sh
```diff
@@ -62,8 +62,32 @@ hash -r;
...
+    ./.tox/"${TOXENV}"/bin/pip wheel -r "${req}" || true;
+done;
+
+# https://github.com/pypa/pip/issues/3414
+
+for fn in .wheels/*pp27*; do
+    if [ "${fn}" != '.wheels/*pp27*' ]; then
+        mv -v "${fn}" "$(echo "${fn}" | sed -e s/pp27/pp240/g)";
+    fi;
+done;
```
ksheedlo: whut

.travis/run.sh
```diff
@@ -13,4 +13,6 @@ if [[ "${MACAPP_ENV}" == "system" ]]; then
 fi;
 
 source ~/.venv/bin/activate;
-tox --develop -- $TOX_FLAGS;
+
+ls .wheels;
```
ksheedlo: Why do we need this in the run script? Is it left in for debugging?

docs/development/getting-started.rst
```diff
@@ -115,11 +115,11 @@ tests.
    $ brew install python
    $ cd /dir/where/mimic/lives/
 
-   # build a virtualenv using the brewed python
+   $ # build a virtualenv using the brewed python
```
ksheedlo:
```
sh: build: command not found
```

:100:

.travis/install.sh
glyph: :computer:

glyph: The link above the `for` loop is the explanation.

.travis/run.sh
glyph: Yeah this was just debugging. But it might be useful to have it in future build logs, so I left it in.

.travis/install.sh
ksheedlo: I tried to understand the linked bug, but I don't know enough about wheels and why installing one would get blocked. Someone else should take a look at this.

glyph: Sadly I doubt that any other mimic reviewer will fully understand the nuance here.  @dstufft? @fijal? want to do a spot check?

dstufft: It's me, the global packaging reviewer!

That looks accurate to me. The only thing I'd say is if PyPy 4.1 is released you'll need to update this for that.

ksheedlo:
LGTM

.travis/install.sh
glyph: The PyPy version is effectively pinned above by the pyenv invocation, so we'll have an opportunity to do that without breaking a random build.

However, hopefully by the time PyPy 4.1 is out this will have been fixed in wheel and we can just delete it...

PR #516, opened 2016-02-01

ziadsawalha:


glyph:
Hooray! First time I can just click the button and not worry about it.  All that work on #515 was totally worth it :)

PR #517, opened 2016-02-02
refactor(maas): use OrderedDict for things

ksheedlo:


glyph:
Adding WIP label due to the coverage issue.

ksheedlo:
ಠ_ಠ

ksheedlo:
Ready for review.

glyph:
I generally like this. I think I'm going to just merge it, since it makes a lot of things less abstract.

However, the one general pattern I don't like here is that it's heavily "look before you leap".  In a follow-up change, it would be great to replace `_assert_object_exists` with `_get_object`, that would actually do the work of returning the object, and raise an exception if it's not found.  (This design pattern has a few names, the one on the c2 wiki being ["couple leaping with looking"](http://c2.com/cgi/wiki?CoupleLeapingWithLooking) although I prefer "with your shield or on it".)

One thing that I'd like to ask for pre-merge is a better PR title for a better commit message - this adds a bunch of new error conditions and stuff, for which there are tests, and that's definitely more interesting for a mimic user reading the changelog than the fact that `OrderedDict` was used to implement it.

PR #526, opened 2016-02-10
refactor(maas_api): with your shield or on it

ksheedlo:
Addresses comments from #517.

PR #527, opened 2016-02-10
refactor(maas_api): with your shield or on it

glyph:
Addresses comments from #517.


glyph:
travis having been poked, this PR is now unnecessary. Back to #526...

PR #528, opened 2016-02-11
feat(maas_api): list agent connections

ksheedlo:
This PR adds a new API call that recently landed in MaaS and
Intelligence is starting to use. It's likely to become our new preferred
way to test if an agent is connected, since the real API has superior
performance when testing many agent connections at once.

### DIFF
--- a/mimic/model/maas_objects.py
+++ b/mimic/model/maas_objects.py
@@ -12,7 +12,7 @@ from uuid import uuid4
 from characteristic import attributes, Attribute
 from six import text_type
 
-from mimic.util.helper import random_hex_generator, random_string
+from mimic.util.helper import Matcher, random_hex_generator, random_port, random_string
 
 METRIC_TYPE_INTEGER = 'i'
 METRIC_TYPE_NUMBER = 'n'
@@ -568,6 +568,9 @@ class MultiHostInfoType(object):
 
 
 @attributes([Attribute('id', instance_of=text_type, default_factory=lambda: text_type(uuid4())),
+             Attribute('connection_guid',
+                       instance_of=text_type,
+                       default_factory=lambda: text_type(uuid4())),
              Attribute('counts',
                        instance_of=dict,
                        default_factory=lambda: {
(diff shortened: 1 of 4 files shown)


mimic/model/maas_objects.py
```diff
@@ -600,6 +603,55 @@ def get_host_info(self, available_types, requested_types, entity_id, clock):
...
+        """
+        return [AgentConnection(id=self.id + '_' + dc,
+                                guid=self.connection_guid,
+                                agent_id=self.id,
+                                endpoint=dc,
+                                datacenter=dc)
+                for dc in ('dfw1', 'ord1', 'lon3')]
+
+
+@attributes([Attribute('id', instance_of=text_type),
```
glyph: In the future, we should be moving to `attrs`, rather than `characteristic`.  There's nothing *wrong* with Characteristic, and it is still maintained, but attrs has a few subtle advantages and eventually we'd like to be on just one utility library of this type.  (You can really think of attrs as characteristic NG, it's a newer library by the same author for the same purpose.)

mimic/model/maas_objects.py
```diff
@@ -600,6 +603,55 @@ def get_host_info(self, available_types, requested_types, entity_id, clock):
...
+                       default_factory=lambda: text_type(random_port()))])
+class AgentConnection(object):
+    """
+    Models an agent connection, which can be queried from MaaS.
+    """
+    def to_json(self):
+        """
+        Serializes this AgentConnection to a JSON-encodable dict.
+        """
+        return {'id': self.id,
```
glyph: One nice thing about `attrs` is that this functionality is built in: `attr.asdict(self)`:

https://attrs.readthedocs.org/en/stable/api.html?highlight=asdict#attr.asdict

mimic/model/maas_objects.py
ksheedlo: I'm actually working on a PR upstream to redo this module in terms of `attrs`, but wanted to get the feature out first.

glyph: Yep. Not a blocking comment.

mimic/rest/maas_api.py
```diff
@@ -1158,6 +1158,35 @@ def view_agent_host_info(self, request, tenant_id):
                          'next_marker': None,
                          'next_href': None}})
 
+    @app.route('/v1.0/<string:tenant_id>/views/connections', methods=['GET'])
+    def view_connections(self, request, tenant_id):
+        """
+        Lists agent connections.
```
glyph: Can you insert an API reference link here?

mimic/test/test_maas.py
```diff
@@ -1135,6 +1135,35 @@ def test_agenthostinfo_missing_include_returns_400(self):
...
+        (resp, data) = self.successResultOf(
+            json_request(self, self.root, b"GET",
+                         '{0}/views/connections'.format(self.uri)))
+        self.assertEquals(resp.code, 400)
+        self.assertEquals(data['message'], 'Validation error for key \'agentId\'')
+
+    def test_view_connections(self):
+        """
+        An existing agent returns a list of connections, and a non-existing agent
+        returns an empty list.
```
glyph: It strikes me as odd that "no agent ID" is a 400, but "invalid agent ID" is a 200.  Is this the behavior of the actual upstream API?

mimic/rest/maas_api.py
ksheedlo: I don't have one yet as it's a new API call. It's similar to https://developer.rackspace.com/docs/cloud-monitoring/v1/developer-guide/#list-agent-connections , but allows you to list connections for many agents at once.

mimic/test/test_maas.py
ksheedlo: Yep.

```
┌─[ken][Kens-MacBook-Pro][±][ksheedlo/fix-2790 ✗][2.1.2][~/racker/sage]
└─▪ ele GET /views/connections
{
  "type": "badRequest",
  "code": 400,
  "message": "Validation error for key 'agentId'",
  "details": "You must specify an agentId",
  "txnId": ".rh-dBH2.h-api1.dfw1.prod.cm.k1k.me.r-EtXiYLc5.c-5122892.ts-1455231353984.v-d067ae1"
}
┌─[ken][Kens-MacBook-Pro][±][ksheedlo/fix-2790 ✗][2.1.2][~/racker/sage]
└─▪ ele GET '/views/connections?agentId=NOPE'
{
  "values": [
    {
      "agent_id": "NOPE",
      "connections": []
    }
  ],
  "metadata": {
    "count": 1,
    "limit": 100,
    "marker": null,
    "next_marker": null,
    "next_href": null
  }
}
```

mimic/model/maas_objects.py
```diff
@@ -797,3 +849,13 @@ def latest_alarm_states_for_entity(self, entity_id):
         for state in alarm_states_for_entity:
             latest_alarm_states_by_alarm[state.alarm_id] = state
         return latest_alarm_states_by_alarm.values()
+
+    def list_connections_for_agent(self, agent_id):
+        """
+        Lists connections for an agent, or returns empty list if there is no such agent.
+        """
+        try:
+            agent = self.agents[self.agents.index(Matcher(lambda agent: agent.id == agent_id))]
```
glyph: Looking at this line of code sparked a thought.  We have a general pattern throughout Mimic where we have data structures that involve lookups on one or more attributes.  There's almost always a "lookup by ID", but there's also often a "lookup by name" or "lookup by relationship to some other object".  It might be interesting to write a small general utility for this sort of structure; that would also help a lot if we wanted to start persisting things across mimic restarts because it would be a single point to go update.

Anyway, not really relevant to this particular PR, but worth thinking about for the future.

ksheedlo: This structure should really be a simple dict; in fact, #529 makes it so.

glyph:
Whoops. See immediately preceding referenced comment.

mimic/rest/maas_api.py
glyph: OK just file an issue to do so when docs are available then.

mimic/test/test_maas.py
glyph: :ok:

ksheedlo:
Filed #531 .

PR #529, opened 2016-02-11
feat(maas_api): add failure case for bad agent ID

ksheedlo:
When an entity is assigned an agent ID that does not map to an existing
agent, trying to fetch the host info for that entity throws a 400.

### DIFF
--- a/mimic/model/maas_objects.py
+++ b/mimic/model/maas_objects.py
@@ -12,7 +12,7 @@ from uuid import uuid4
 from characteristic import attributes, Attribute
 from six import text_type
 
-from mimic.util.helper import Matcher, random_hex_generator, random_port, random_string
+from mimic.util.helper import random_hex_generator, random_port, random_string
 
 METRIC_TYPE_INTEGER = 'i'
 METRIC_TYPE_NUMBER = 'n'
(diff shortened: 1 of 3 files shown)


ksheedlo:
Slight conflicts with #528 as that one looks up agents from a data structure which changes in this PR.

glyph:
This looks pretty good to me.  The one change I'd like is just to add a reference link to the docs.  It's a little repetitive but consistently including those makes it easier for maintainers to just jump in to one spot in the code. Other than that, just have a look at the comments and double-check the behavior of invalid agent IDs, and if that's correct, go ahead and land.

glyph:
Whoops. Meant to make those changes to #528

glyph:
OK, now comments *actually* intended for this PR: 

:+1: , please land this when you have landed #528 and resolved any conflicts.

glyph:
Apropos of the comment on #528 as well:

This is definitely an improvement, but I still think we may want to consider a helper data structure, because (as with many other locations in Mimic) there is manual work and imperative code associated with keeping the `agent_id`-as-key-in-dictionary and `agent_id`-as-attribute-of-object in sync.  Of course changing agent IDs is not a valid operation, but it would be nice to just specify it once and have it automatically used because the indexing structured just specified `agent_id` as the attribute to index by.

glyph:
Of course, having it automatically work for attributes which *are* mutable would just be a bonus :).

PR #532, opened 2016-02-12
refactor(maas_objects): better validation with attrs

ksheedlo:
This change rewrites objects in the `mimic.model.maas_objects` module to
use the `attr` library for defining fields and validation, from
`characteristic` before.

### DIFF
--- a/mimic/model/maas_objects.py
+++ b/mimic/model/maas_objects.py
@@ -9,33 +9,42 @@ import random
 import string
 from uuid import uuid4
 
-from characteristic import attributes, Attribute
+import attr
+from attr.validators import instance_of, optional, provides
 from six import text_type
+from twisted.internet.interfaces import IReactorTime
 
-from mimic.util.helper import random_hex_generator, random_port, random_string
+from mimic.util.helper import one_of_validator, random_hex_generator, random_port, random_string
 
 METRIC_TYPE_INTEGER = 'i'
 METRIC_TYPE_NUMBER = 'n'
 METRIC_TYPE_STRING = 's'
 
 
-@attributes([Attribute('agent_id', default_value=None),
-             Attribute('alarms', default_factory=collections.OrderedDict),
-             Attribute('created_at', instance_of=int),
-             Attribute('checks', default_factory=collections.OrderedDict),
-             Attribute('id',
-                       default_factory=(lambda: u'en' + random_hex_generator(4)),
-                       instance_of=text_type),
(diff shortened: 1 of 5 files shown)


mimic/model/maas_objects.py
```diff
@@ -9,33 +9,46 @@
...
-             Attribute('id',
-                       default_factory=(lambda: u'en' + random_hex_generator(4)),
-                       instance_of=text_type),
-             Attribute('ip_addresses', default_factory=dict, instance_of=dict),
-             Attribute('label', default_value=u'', instance_of=text_type),
-             Attribute('managed', default_value=False, instance_of=bool),
-             Attribute('metadata', default_factory=dict, instance_of=dict),
-             Attribute('updated_at', instance_of=int),
-             Attribute('uri', default_value=None)])
+def _is_alarm_state(instance, attribute, value):
```
glyph: Shouldn't this just be `_is_alarm_state = one_of_validator("OK", "WARNING", "CRITICAL", "UNKNOWN")`?

If not for some reason (perhaps some tests depend on a specific `ValueError`?), a more succinct way to express this would be `if value not in ["OK", "WARNING", "CRITICAL", "UNKNOWN"]:`.

```
    if value != 'OK' and value != 'WARNING' and value != 'CRITICAL' and value != 'UNKNOWN':
```

ksheedlo: Good eye. I don't remember why I thought I had to do that.

mimic/model/maas_objects.py
```diff
@@ -189,32 +183,29 @@ def update(self, **kwargs):
...
-             Attribute('label', default_value=u'', instance_of=text_type),
-             Attribute('metadata', default_factory=dict, instance_of=dict),
-             Attribute('type', default_value='email', instance_of=text_type),
-             Attribute('updated_at', instance_of=int)])
+@attr.s
 class Notification(object):
     """
     Models a MaaS Notification.
     """
+    created_at = attr.ib(validator=attr.validators.instance_of(int))
```
glyph: These lines look pretty repetitive. Perhaps it would be better to `from attr.validators import instance_of` rather than re-typing that whole identifier each time? Or `an_int = attr.validators.instance_of(int)` up at the top so each line only says `validator=an_int` ?

glyph:
This looks good; address the comment about brevity (or not, it's a taste thing, so discretion is with you as the author), and go ahead and merge.  Thanks for paying down some of this tech debt :).

PR #538, opened 2016-02-17
refactor(*): migrate characteristic to attrs

ksheedlo:


ksheedlo:
Needs: `characteristic` removed from project's dependencies, but I haven't done one of these before for Mimic so I'll wait and see what I need to do about that.

glyph:
This is totally a step in the right direction; we can do the dependency removal, and any necessary subsequent cleanups, in a separate PR.

PR #540, opened 2016-02-18
refactor(*): rewrite 2.6 dict comprehensions

ksheedlo:
This change rewrites some dict comprehension expressions from when 2.6 compatibility was a requirement. In particular we don't have to

```python
dict((k, v) for k, v in comprehension)
```

in favor of the 2.7 syntax for dict comprehensions

```python
{k: v for k, v in comprehension}
```


glyph:
I suppose it is a bit more readable this way.  And we should be discouraging 2.6 aggressively.  So. :shipit:

PR #541, opened 2016-02-18
refactor(maas_objects): serialize by excluding fields

ksheedlo:
Instead of enumerating all the fields we want to serialize in
`Entity.to_json` and `AlarmState.to_json`, we can use a comprehension to
skip the fields we don't want to serialize. The result is less
boilerplate code.

### DIFF
--- a/mimic/model/maas_objects.py
+++ b/mimic/model/maas_objects.py
@@ -56,15 +56,7 @@ class Entity(object):
         """
         Serializes the Entity to a JSON-encodable dict.
         """
-        return {'label': self.label,
-                'id': self.id,
-                'agent_id': self.agent_id,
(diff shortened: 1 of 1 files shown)


glyph:
Did you consider using https://attrs.readthedocs.org/en/stable/api.html?highlight=asdict#attr.asdict for this?

ksheedlo:
No. I think the filter is what I was looking for. Will update whenever I
get back to the keyboard.

glyph:
Cool. A definite step in the right direction.  I have this vague sense we could do something much cooler with serialization (schemas for parameters or return types, somehow? annotations on `attr.ib` objects indicating their serializability?); but, in the meanwhile, this is going to delete a lot of code :).

PR #542, opened 2016-02-18
stop working around fixed bug in 'wheel'

glyph:
It should now generate the correct identifiers for pypy wheels.


ksheedlo:
:+1: to me

PR #544, opened 2016-02-18
refactor(*): rewrite 2.6 dict comprehensions

ksheedlo:
Same code as #540, just kicking travis.


glyph:
Thusly kicked we may now close.

PR #546, opened 2016-02-19

ziadsawalha:


glyph:
Looks like 20.2 got pulled?

glyph:
https://pypi.python.org/pypi/setuptools shows 20.1.1 right now

PR #548, opened 2016-02-21
chore(clb_objects): remove dependency on toolz

ksheedlo:
`attrs` has a built in functionality for filtering out keys when
serializing to a dict. We used `toolz.dicttoolz.dissoc` in one place for
doing this, but we can use a filter to `attr.asdict` and remove the
dependency.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -34,7 +34,6 @@ def setup_options(name, version):
             "attrs>=15.1.0",
             "testtools>=1.7.1,<1.8.0",
             "iso8601>=0.1.10",
-            "toolz>=0.7.4"
         ],
         package_dir={"mimic": "mimic"},
         packages=find_packages(exclude=[]) + ["twisted.plugins"],
--- a/mimic/model/clb_objects.py
+++ b/mimic/model/clb_objects.py
@@ -14,8 +14,6 @@ import attr
 
 from six import text_type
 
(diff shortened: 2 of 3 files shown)


ksheedlo:
whoops, careful! I manually edited `requirements/production.txt` against the instructions in the file. Going to check that real quick.

ksheedlo:
Running `pip install .; pip install coverage; pip freeze` in the virtualenv didn't change `requirements/production.txt` in any interesting way, so this PR is good to review.

setup.py
```diff
@@ -34,8 +34,7 @@ def setup_options(name, version):
             "xmltodict>=0.9.1",
             "attrs>=15.1.0",
             "testtools>=1.7.1,<1.8.0",
-            "iso8601>=0.1.10",
-            "toolz>=0.7.4"
+            "iso8601>=0.1.10"
```
glyph: For future reference, the convention in Python (I am surprised to see we weren't doing it here!) is to always put a trailing comma at the end of the line to avoid spurious diffs like this.

glyph:
LGTM. Thanks for finding this.

glyph:
(Assigning to you so you can put the comma back :))

ksheedlo:
Update: Fixed comma.

PR #549, opened 2016-02-21
chore(requirements): remove jsonschema

ksheedlo:
`jsonschema` is an unused dependency that this change removes.


glyph:
Wow. Really nice to get rid of that functools dependency which required the weird annotation. Thanks.

PR #551, opened 2016-02-24
Load mocks implementing IAPIDomainMock

derwolfe:
This PR adds 
- logic to load mocks implementing `IAPIDomainMock`
- a test showing that this logic works

### DIFF
--- a/mimic/core.py
+++ b/mimic/core.py
@@ -10,7 +10,7 @@ from twisted.python.urlpath import URLPath
 from twisted.plugin import getPlugins
 from mimic import plugins
 
-from mimic.imimic import IAPIMock
+from mimic.imimic import IAPIMock, IAPIDomainMock
 from mimic.session import SessionStore
 from mimic.util.helper import random_hex_generator
 from mimic.model.mailgun_objects import MessageStore
@@ -37,6 +37,9 @@ class MimicCore(object):
 
         :param apis: an iterable of all :obj:`IAPIMock`s that this MimicCore
             will expose.
+
+        :param domains: an iterable of all :obj:`IAPIDomainMock`s that this
+            MimicCore will expose.
         """
         self._uuid_to_api = {}
         self.sessions = SessionStore(clock)
@@ -55,9 +58,12 @@ class MimicCore(object):
     @classmethod
     def fromPlugins(cls, clock):
         """
-        Create a :obj:`MimicCore` from all :obj:`IAPIMock` plugins.
+        Create a :obj:`MimicCore` from all :obj:`IAPIMock` and
+        :obj:`IAPIDomainMock` plugins.
         """
-        return cls(clock, list(getPlugins(IAPIMock, plugins)))
+        all_plugins = getPlugins(IAPIMock, plugins)
+        domain_plugins = getPlugins(IAPIDomainMock, plugins)
+        return cls(clock, all_plugins, domain_plugins)
 
     def service_with_region(self, region_name, service_id, base_uri):
         """
--- a/mimic/test/test_core.py
+++ b/mimic/test/test_core.py
@@ -62,7 +62,7 @@ class CoreBuildingTests(SynchronousTestCase):
     def test_load_domain_plugin_includes_all_domain_plugins(self):
         """
         Using the :func:`MimicRoot.fromPlugin` creator for a
-        :class:`MimicCore`, domains mocks implementing `class`:`IAPIDomainMock`
+        :class:`MimicCore`, domain mocks implementing `class`:`IAPIDomainMock`
         are included.
         """
         # build a fake plugins package with a domain plugin
--- a/mimic/test/test_core.py
+++ b/mimic/test/test_core.py
@@ -1,7 +1,10 @@
 from __future__ import absolute_import, division, unicode_literals
 
+import sys
+
 from twisted.internet.task import Clock
 from twisted.trial.unittest import SynchronousTestCase
+from twisted.python.filepath import FilePath
 
 from mimic.core import MimicCore
 from mimic.plugins import (nova_plugin, loadbalancer_plugin, swift_plugin,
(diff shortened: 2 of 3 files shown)


mimic/core.py
```diff
@@ -37,6 +37,9 @@ def __init__(self, clock, apis, domains=()):
 
         :param apis: an iterable of all :obj:`IAPIMock`s that this MimicCore
             will expose.
+
+        :param domains: an iterable of all :obj:`IAPIDomainMock`s that this
+            MimicCore will expose.
```
ksheedlo: Thanks for the documentation!

mimic/test/test_core.py
```diff
@@ -55,3 +59,6 @@ def test_from_plugin_includes_all_plugins(self):
             len(plugin_apis),
             len(list(core.entries_for_tenant('any_tenant', {},
                                              'http://mimic'))))
+        self.assertEqual(
+            len(domain_apis),
+            len(list(core.domains)))
```
ksheedlo: I think the dummy plugin wasn't intended to be included in the default core. If we had any real domain plugins in this repo we could do this test. I have an idea for one.

derwolfe: @ksheedlo - I've created (but not yet pushed or finished implementing) a test that writes a plugin implementing IAPIDomainMock, loads it, and messes with the path to make it visible. I'm most interested in testing that plugins implementing the interface are indeed loaded by `fromPlugins`.

ksheedlo: nice! Look forward to seeing the updated version.

mimic/test/test_core.py
```diff
@@ -55,3 +58,43 @@ def test_from_plugin_includes_all_plugins(self):
...
+domain_plugin = ExampleDomainAPI()
+__all__ = [domain_plugin]
+"""
+        self.package.child('_domain.py').setContent(plugin)
+        sys.path.insert(1, self.root.path)
+
+        import fakeplugins
+        import mimic.plugins
+
+        self.original = mimic.plugins
```
derwolfe: I really dislike how I'm doing this. I'm working on trying to use `sys.modules["mimic.plugins"] = fakeplugins`, yet I've not gotten this work yet.

mimic/test/test_core.py
```diff
@@ -55,3 +58,40 @@ def test_from_plugin_includes_all_plugins(self):
...
+__all__ = [domain_plugin]
+"""
+        self.package.child('_domain.py').setContent(plugin)
+        sys.path.append(self.root.path)
+
+        # monkey patch mimic.plugins package with fakeplugins
+        import fakeplugins
+        import mimic.plugins
+        mimic.plugins = fakeplugins
+
```
derwolfe: I am not a fan of how I am reassigning `mimic.plugins` to `fakeplugins`. That being said; I've tried using `t.t.u.patch` and `sys.modules['mimic.plugins'] = fakeplugins` and have not been able to get the test to pass. If anyone has a suggestion to handle this more elegantly, I'd be happy to change it :-).

glyph: The important thing here is to put `mimic.plugins` *back* by using [`addCleanup`](https://docs.python.org/2.7/library/unittest.html#unittest.TestCase.addCleanup).  As it is, this corrupts state for all subsequent tests in the process.

derwolfe: I'll add that back in - thanks :smile:

mimic/test/test_core.py
```diff
@@ -55,3 +58,47 @@ def test_from_plugin_includes_all_plugins(self):
...
+        init = b"""from fakeplugins._domain import domain_plugin
+__all__ = [domain_plugin]
+"""
+        self.package.child('__init__.py').setContent(init)
+        plugin = b"""from mimic.test.dummy import ExampleDomainAPI
+domain_plugin = ExampleDomainAPI()
+__all__ = [domain_plugin]
+"""
+        self.package.child('_domain.py').setContent(plugin)
+        sys.path.append(self.root.path)
```
glyph: this `sys.path` entry sticks around since no `addCleanup` removes it...

mimic/test/test_core.py
```diff
@@ -55,3 +58,47 @@ def test_from_plugin_includes_all_plugins(self):
...
+        """
+        # build a fake plugins package with a domain plugin
+        self.root = FilePath(self.mktemp())
+        self.root.createDirectory()
+        self.package = self.root.child('fakeplugins')
+        self.package.createDirectory()
+        init = b"""from fakeplugins._domain import domain_plugin
+__all__ = [domain_plugin]
+"""
+        self.package.child('__init__.py').setContent(init)
```
derwolfe: Drop `init.py`. Instead, extend `mimic.plugins__path__` to include the new fake domain plugin. Then cleanup `mimic.plugins.__path__` by remove the plugin's path.

glyph: You will also have to clean up `mmic.plugins.fake_plugin` (or whatever you end up changing `.child("...")` to point at instead of `__init__.py`

glyph: This clean-up should consist of `sys.modules.pop("mimic.plugins.fake_plugin")` and `del mimic.plugins.fake_plugin`.

mimic/core.py
```diff
@@ -55,9 +58,12 @@ def __init__(self, clock, apis, domains=()):
     @classmethod
     def fromPlugins(cls, clock):
         """
-        Create a :obj:`MimicCore` from all :obj:`IAPIMock` plugins.
+        Create a :obj:`MimicCore` from all :obj:`IAPIMock` and
+        :obj:`IAPIDomainMock` plugins.
         """
-        return cls(clock, list(getPlugins(IAPIMock, plugins)))
+        all_plugins = getPlugins(IAPIMock, plugins)
```
glyph: Minor nitpick: `all_` is a bad prefix, it should be like, `service_catalog_` or something, since it clearly isn't *all* plugins.  That won't stop me from merging, though.

glyph:
Looks good. Thanks for fixing this issue.

PR #553, opened 2016-02-25
Fix for new setuptools

glyph:
Looks like `pkg_resources` picked up a new dynamic import.

### DIFF
--- a/bundle/start-app.py
+++ b/bundle/start-app.py
@@ -27,8 +27,8 @@ from twisted.python import log
 
 # The following are required by pkg_resources.resource_string, which is used by
 # treq; so trick modulegraph into including it.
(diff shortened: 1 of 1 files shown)


glyph:
I think that we might want to land this even though `setuptools` has un-released 20.2.1 just as it un-released 20.2.0; it seems clear that a new release of setuptools is going to start depending on new packages from `pkg_resources._vendor` whenever a non-buggy version of 20.2.x comes out.

glyph:
I've merged master to trigger a new build, to ensure that this works on "old" setuptools as well.

glyph:
@manishtomar - if you could do a quick review once the tests have passed, that would be great

manishtomar:
:+1:

PR #554, opened 2016-02-25
add mac standard pip cache directory

glyph:


glyph:
From Travis support:

> … on .org OS X builds, caches are not available …

so I guess this is a lost cause. Oh well! Perhaps some of these changes might be useful for Linux too, since I think our usage of `PIP_FIND_LINKS` is a bit wonky, but that is for a esparate PR.

PR #556, opened 2016-02-27
feat(yo_api): implement new yo domain mock

ksheedlo:
this could be a fun thing to put in the test for #551 so that we don't have to try to include the dummy plugin.

I'm currently stuck trying to find the helper URI to send test requests against.

### DIFF
--- a/mimic/model/yo_objects.py
+++ b/mimic/model/yo_objects.py
@@ -1,0 +1,51 @@
+"""
+Yo API data model
+"""
+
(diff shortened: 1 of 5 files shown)


ksheedlo:
update: can demo the new api locally, but tests won't pass.

ksheedlo:
Update: Fixed tests. will check coverage and mark when RFR

glyph:
I'm going to take your word for it that this is how the Yo API works; otherwise, this looks great :).

PR #560, opened 2016-03-02
feat(maas_api): control endpoint for loading fixtures

ksheedlo:


ksheedlo:
TODO: tests, coverage.

I also have a problem where certain objects in the fixture may need to refer to other objects in such a way that the creator of the fixture doesn't know object IDs ahead of time. This is because IDs in MaaS are randomly chosen for a low probability of collisions, and it feels weird to allow the user to choose them. But still alarms need to reference a notification plan and suppressions need to reference entities, checks, alarms, and notification plans.

For now this is how it is because I need it for a feature I'm developing in Intelligence; I'm open to suggestions as to how to improve it.

glyph:
Right now it seems like this is abandoned.  If anyone wants to fix it again, please feel free to reopen!

ksheedlo:
👌

PR #563, opened 2016-03-09
s/all_plugins/service_catalog_plugins

derwolfe:
`all_plugins` is incorrect, these are all of the `service_catalog plugins`


glyph:
Thanks!

PR #568, opened 2016-03-16
Remove a repeated fetch of entity.

ddunlop:
Entity is fetched in the statement above this was a duplicate fetch

### DIFF
--- a/mimic/rest/maas_api.py
+++ b/mimic/rest/maas_api.py
@@ -1136,7 +1136,6 @@ class MaasMock(object):
             request.setResponseCode(e.code)
             return json.dumps(e.to_json())
 
-        entity = entities[entity_id]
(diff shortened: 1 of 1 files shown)


glyph:
1. tests pass
2. deletes code

:shipit:

ksheedlo:
@ddunlop @glyph can you guys make sure to add the blue "ready for review" label for changes like this when it's ready and before merging?

Reason why is I'm experimenting with a new dashboard and workflow for open source maintainers. We're working with a model that requires a PR to transition into a ready state (signified by the label) and then a maintainer gets credit for the review when it's closed or merged in or the label comes off.

glyph:
Thanks for the heads up, @ksheedlo .  All the more motivation for me to write that label-bot :).

PR #578, opened 2016-03-28
Allow builds that read stdin to fail.

glyph:
PyObjC 3.1 made a small change that causes its setup.py to read from
stdin and thereby block forever on pypy.  Allow that to fail quickly
like it used to, instead of blocking.


glyph:
This is designed to unblock #577 .

derwolfe:
Thanks @glyph; this LGTM (pending travis passing). 

@glyph has also reported the [bug](https://bitbucket.org/ronaldoussoren/pyobjc/issues/146/install-hangs-forever-on-pypy) upstream to the pyobjc maintainers.

PR #585, opened 2016-04-10

ziadsawalha:


glyph:
Since this one's a setuptools bump I'm going to wait 48h or so before merging it, given the last couple of version clawbacks we've seen on PyPI.

PR #598, opened 2016-04-26

ziadsawalha:


glyph:
It appears that codecov had a minor outage; the branch coverage delta that the previous coverage report pointed at was an `if bytes is str:` in `swift_api` which suggests that python 3 test coverage didn't actually get reported.  I'm re-running the travis build in the hopes that we get a coverage status and don't need to make any weird exceptions for this PR.

glyph:
The outage is ongoing and I have no reason to believe this coverage gap is real so I'm going to merge.

PR #600, opened 2016-04-28
identity-admin: Update against master

BenjamenMeyer:
Merge to update the identity-admin branch with the latest from master

Tested using tox for py27 and py34 and all tests passed.


glyph:
Wow. that's a lot of commits.  Did you break Travis somehow? :)

glyph:
Oh, I guess changing travis.yml to a super old incompatible configuration would break it, yeah. :-|.  It looks like this merge went poorly and we should abandon this attempt.

Rather than merging `master` into `identity_admin`, could you create a new branch off of `master` and merge `identity_admin` into it, instead?  This would probably make GitHub's dysfunctional history viewer freak out less, and also keep the diff and commit list a lot smaller.

Thanks for trying to catch this PR up to date!

glyph:
(To be clear: when I say "this merge went poorly" I mean "the merge commit contains lots of accidental reverts of old commits".)

PR #603, opened 2016-04-29
Refactor: Rename AuthApi to IdentityApi

BenjamenMeyer:
AuthApi is going to start taking on more non-auth functionality,
and begin to reflect more of Rackspace Identity/OpenStack Keystone.
Therefore it is only fitting to rename it from AuthApi to IdentityApi.


glyph:
Re-ran the Travis build in the hopes that @codecov will get itself together for this run...

glyph:
This is definitely a better name.

PR #606, opened 2016-05-02
Enhancement: API Management

BenjamenMeyer:
- Add a layer in mimic core to enable the differentiation between internally hosted and external provided services

### DIFF
--- a/mimic/core.py
+++ b/mimic/core.py
@@ -20,6 +20,150 @@ from mimic.model.glance_objects import GlanceAdminImageStore
 from mimic.model.valkyrie_objects import ValkyrieStore
 
 
+class _MimicCoreApiWrapperEntry(object):
+    """
+    Wrap an API entry to be able to distinguish between internally
+    hosted implemented as :obj:`IAPIMock` and externally hosted APIs
+    that do not live within Mimic.
+    """
+
+    def __init__(self, api, external):
+        """
+        Create a wrapped :obj:`IAPIMock` entry
+
+        :param api: the :obj:`IAPIMock` instance to wrap
+        :param external: boolean value of whether the :obj:`IAPIMock`
+            is internally or externally hosted.
+        """
+        self.api = api
+        self.base_uri = None
+        self.external = external
+        self.api_uuid = ((api.__class__.__name__) + '-' +
+                         random_hex_generator(3))
+
+    def uri_for_service(self, region, service_id, base_uri):
+        """
+        Generate a URI prefix for a given region and service ID.
+
+        Each plugin loaded into mimic generates a list of catalog entries; each
+        catalog entry has a list of endpoints.  Each endpoint has a URI
+        associated with it, which we call a "URI prefix", because the endpoint
+        will have numerous other URIs beneath it in the hierarchy, generally
+        starting with a version number and tenant ID.  The URI prefixes
+        generated for this function point to the top of the endpoint's
+        hierarchy, not including any tenant information.
+
+        :param unicode region: the name of the region that the service resource
+            exists within.
+        :param unicode service_id: the UUID for the service for the specified
+            region
+        :param str base_uri: the base uri to use instead of the default - most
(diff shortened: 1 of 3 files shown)


BenjamenMeyer:
@glyph if there is a better way per mimic's design and twisted, then I'm all ears. Modifying IAPIMock didn't seem to be quite appropriate and this provides a nice clean interface that doesn't require changing everything in Mimic to support it.

glyph:
I like where you're going with this in general; composition FTW.  However:

1. You're trying to emulate a mapping interface, but:
 1. it's not exposed publicly (so API clients outside of `mimic.core` don't care)
 2. coverage issues in `__getitem__`, `set`, etc, suggest it is over-engineered and we have a lot more code than we need here to get the job done
 3. you have odd sticky-outy bits like `values_unwrapped` which are _not_ part of the mapping interface; I feel like you should either jump in with both feet on one idiom or the other. Option 1: just have nothing-but-mapping, in which case, have separate "wrapped" and "unwrapped" objects that act as façades to the different modes.  Option 2: don't have any pretense at all towards a mapping API, and just have `with_wrapping` and `without_wrapping` methods which lookup the key one way or the other.
2. Since no `_MimicCoreApiWrapperEntry` here exists with `external` set to `True` it's not clear what the public API for this *would* be, so it's still sort of an open question.  Could you add one example, at least in the tests? (This would also address other coverage issues.)
3. Clearly you don't have enough tests for `__repr__` either, because `'%s(%s)'.format` is not how one does that :).

Point 2 seems like the most serious concern to me.  One way to gradually expand an API like this is to add a *new* interface, `IExternalAPIMock`, and do a [`.providedBy` check](http://docs.zope.org/zope.interface/README.html#declaring-implemented-interfaces) to see if the plugins provide that interface as well, and if they do, treat them differently.

glyph:
@BenjamenMeyer please feel free to ask for clarification on any of those points - most importantly though appease the github status gods ::)

BenjamenMeyer:
@glyph Thanks.

I basically re-implemented the dictionary interface to try to keep as much code working the same as it was before to basically hide the wrapper level. There's two functions that can be used to work with the wrapper object just in case, but as I said - if there's a better way for how Mimic works so that it builds off of IAPIMock more appropriately then I'm all ears, I just don't want to have to force every existing API mock to be modified.

I wanted to push this up now to get an idea of whether I'm heading in the right direction or if there's a simpler direction that fits the code base better. As such, I kept it minimal so this only adds the wrapping layer and not the additional functionality to start modifying the contents - e.g adding an external API to the list, or modifying them, etc. I didn't want to get too far down this path before finding out it was the wrong path.

Per point 2, the interface could happen one of two ways: (i) the external entity directly modifying MimicCore's `_uuid_to_api` dictionary which would create the wrapper and enable it to set `external` appropriately, or (ii) a function to do the same more officially (so the external doesn't access the "private" _uuid_to_api attribute).

I like the IExternalAPIMock idea. That might work. I'll look into it.

Also, is there a way I can get the code coverage data from the tox runs? I don't see any information on in my local tox runs for py27,py34. Thanks!

glyph:
> Also, is there a way I can get the code coverage data from the tox runs? I don't see any information on in my local tox runs for py27,py34. Thanks!

It is running `coverage`, but it's running it with the `-p` flag, so that parallel runs don't stomp on each other (as with detox).  So you have to do:

```console
$ detox -e py27 -e py34
# spin, spin, spin
$ coverage combine
```

and then at that point you can run `coverage report` or `coverage html` or whatever you like to see a combined coverage report.

If you'd prefer it if tox could do this for you, then please [vote for the issue on tox](https://bitbucket.org/hpk42/tox/issues/275/allow-running-commands-before-and-after) to allow this to be expressed within `tox.ini` :).

BenjamenMeyer:
I'll close this PR if #608 is accepted as they're both infrastructure PRs for the same thing, just different implementations.

PR #607, opened 2016-05-02
Increase version number to 2.1.0

glyph:
Prepare for 2.1.0 release.


glyph:
Release is already on PyPI: https://pypi.python.org/pypi/mimic/2.1.0

glyph:
already released! so, let's do this

PR #608, opened 2016-05-03
Enhancement: IExternalAPIMock support

BenjamenMeyer:
Infrastructure for adding an external service to the service catalog.

Better version of #606 based on feedback in that PR from @glyph


BenjamenMeyer:
Note: This won't support doing End-Point Templates as I think we'll need another mechanism to do so. For some basic information see https://github.com/mseknibilel/OpenStack-Install-and-Understand-Guide/blob/master/OpenStack_Essex/Keystone/keystone-2012.1/doc/source/old/endpoints.rst. I'll have to think on that side more; it'll be good for a different PR.

That said, this PR should support managing specific End-Points. In a sense, that'll be sort of like the template version any way since we'll be mucking with the tenantid portion of the field any way

mimic/core.py
```diff
@@ -51,9 +52,25 @@ def __init__(self, clock, apis, domains=()):
...
 
         for api in apis:
-            this_api_id = ((api.__class__.__name__) + '-' +
-                           random_hex_generator(3))
-            self._uuid_to_api[this_api_id] = api
+            # Gate check the API to make sure it implements one of the
+            # supported interfaces
+            required_interfaces = (IAPIMock, IExternalAPIMock)
+            provided_interfaces = list(zope.interface.providedBy(api))
+            if True in [i in provided_interfaces
```
glyph: So, this is just a general stylistic comment, and you can do with it what you like, it's definitely not a blocker.

`zope.interface` has some really interesting calculus to decide whether an interface is provided or not, and it's generally best to let it do its job.  As far as I know this implementation is correct, but I always end up forgetting one or two weird edge cases.

But what you're really trying to do here, if you implement it yourself, is to decide if `provided_interfaces` and `required_interfaces` have a non-empty common subset, which can be expressed more gracefully as:

```python
required_interfaces = set([IAPIMock, IExternalAPIMock])
provided_interfaces = set(zope.interface.providedBy(api))
if required_interfaces & provided_interfaces:
    # ...
```

But: better yet would be to avoid implementing implementation-calculus at all on your own, even if it is probably undeniably correct, and instead express this as:

```python
if IExternalAPIMock.providedBy(api):
    this_api_id = ((api.name_key) + '-' + random_hex_generator(3))
elif IAPIMock.providedBy(api):
    this_api_id = ((api.__class__.__name__) + '-' + random_hex_generator(3))
else:
    raise TypeError(
        api.__class__.__module__ + '/' +
        api.__class__.__name__ +
        " does not implement IAPIMock or IExternalAPIMock"
    )
self._uuid_to_api[this_api_id] = api
```

mimic/core.py
```diff
@@ -82,11 +99,19 @@ def service_with_region(self, region_name, service_id, base_uri):
...
-                self.sessions,
-            )
+            if IExternalAPIMock.providedBy(api):
+                # External APIs only provide a catalog entry, nothing more
+                raise RuntimeError(
+                    api.__class__.__module__ + '/' +
+                    api.__class__.__name__ +
+                    " is an external service to Mimic."
+                )
+            else:  # IAPIMock.providedBy(api)
```
glyph: I think the comment here should be a bit more expository; what I believe it's trying to express is "we already validated that it implements one or the other of these in `__init__`"?

mimic/core.py
```diff
@@ -128,9 +153,17 @@ def entries_for_tenant(self, tenant_id, prefix_map, base_uri):
...
-                    )
-                yield entry
+            if IExternalAPIMock.providedBy(api):
+                for entry in api.catalog_entries(tenant_id):
+                    for endpoint in entry.endpoints:
+                        prefix_map[endpoint] = api.uri_for_service(
+                            endpoint.region, service_id
+                        )
+                    yield entry
+            else:  # IAPIMock.providedBy(api)
```
glyph: Maybe instead of having a comment here we could just add private methods, `_all_external_mocks` and `_all_service_catalog_mocks`, which explain this?  Or even an `iter_api_mocks` that yields triples of `service_id, api, external`, where `external` is just a boolean, if we don't want the overhead of having to iterate twice?

mimic/test/dummy.py
```diff
@@ -91,3 +91,38 @@ def resource(self):
...
+
+@implementer(IExternalAPIMock, IPlugin)
+class ExampleExternalAPI(object):
+    """
+    Example external API the return nothing
+    """
+
+    def __init__(self, name=u"example",
+                 regions_and_versions=[
+                     ('ORD', 'v1',
```
glyph: Could we pick a region other than ORD for testing here so it's easier for readers of the tests to recognize it immediately as distinct from other places where "ORD" is a placeholder? Maybe just `"EXTERNAL"`?

glyph:
None of this feedback looks super significant, but if you feel the implementation changes require further review, feel free to resubmit.  Otherwise, feel free to hit the merge button.

mimic/core.py
BenjamenMeyer: I don't think we're iterating the loop twice here. Each entry in `self._uuid_to_api` each object can only be either an Internal API or an External API but not both. If we actually had an `_all_external_mocks` and `_all_internal_mocks` then yes, we would have to iterate over the data twice. However, if we were going to do that then I would split them into separate storage specifically to keep from looping over all the data twice.

I can adjust the comment if you like.

mimic/core.py
BenjamenMeyer: I'll actually be refactoring this a little in another PR. I think the final proposal makes sense to do.

The two things that will happen later is the entire body of the loop will become an `add` function call, and then the interface check might also become a `validate` function used by the `add` function (not sure yet on the validate function part).

mimic/test/dummy.py
BenjamenMeyer: Will do.

BenjamenMeyer:
@glyph overall I think it's probably good, but I did do a slight refactor while incorporating the changes so I'm going to re-open for review just to make sure.

mimic/core.py
glyph: > I don't think we're iterating the loop twice here.

No, I just proposed two solutions, the first of which would double-iterate.

glyph: > Not that I'm expecting a lot of external mocks, but it can reduce branching if the storage is split

Exactly, I'm just looking to simplify the code, not address issues of scale.

mimic/test/dummy.py
glyph: Thanks :)

mimic/core.py
glyph: OK.  As I said, stylistic; hopefully educational for you for future reference, maybe a future PR for cleanup, but we can proceed with this PR as-is.

mimic/core.py
```diff
@@ -35,13 +35,16 @@ def __init__(self, clock, apis, domains=()):
...
-        :param apis: an iterable of all :obj:`IAPIMock`s that this MimicCore
-            will expose.
+        :param apis: an iterable of all :obj:`IAPIMock` and
+            :obj:`IAPIExternalAPI` mocks that this MimicCore will expose.
 
         :param domains: an iterable of all :obj:`IAPIDomainMock`s that this
             MimicCore will expose.
         """
-        self._uuid_to_api = {}
+        self._uuid_to_api = {
```
glyph: This is also a style thing, but I'd prefer you block on it: I *really* don't like the practice of pretending that raw dictionaries are structured objects.  It's hard to document; a simple `:ivar:` ReST or epytext field can document what `_uuid_to_api` is, but there isn't even any syntax for explaining what `_uuid_to_api["internal"]` is.  It is harder on IDEs, because they can't "see" the structure of this attribute.  It's harder on grep, because you could use single or double quotes, or a constant.  It's harder on linters because they can't verify that attribute names match; it's a dict, so any key of the right type should be valid!  It's harder to catch versioning issues, because we're preserving the name `self._uuid_to_api` but giving its keys and values totally different types.

I could go on :).

For these reasons, and more, please replace `self._uuid_to_api['external']` with `self._uuid_to_external` and add `:ivar:`s for each of these, then go ahead and land on master; this is my last bit of feedback.

Thanks for participating in Mimic's development 😄!

glyph: Ahem. Implied in previous comment:

- replace `_uuid_to_api['internal']` with just `_uuid_to_api`, since this will make for a smaller change and preserve the type of the variable through history better.
- "both of these" for `:ivar:` on `MimicCore` meaning both the `_uuid_to_api` ivar and the `_uuid_to_external` ivar.

Issue #219, opened 2015-04-24
Inject entries into the service catalog


lvh:
Two use cases:

1. I am writing tests for something that orchestrates APIs but don't want to hit production Nova. I hook otter up to Mimic, then tell Mimic about that instance so that it shows up in the service catalog.
2. I am writing things that test some other complex orchestration system. I want to run all my tests, but not everything has Mimic mocks yet. I don't want to hard code the things that don't use Mimic; my tests should just grab things from the service catalog. Injecting entries into the service catalog means that I can start introducing more and more Mimic mocks piecemeal as necessary.

Ideally this would eventually happen by copying the identity admin API.

lekhajee:
Link to the identity admin API to add endpoints to a service catalog: http://developer.openstack.org/api-ref-identity-v2-ext.html#os-kscatalog-ext

BenjamenMeyer:
I'm basically working this now for Cloud Backup; taking over from @lvh

BenjamenMeyer:
PRs #608 and #614 are steps towards this.

BenjamenMeyer:
I'm presently only addressing external APIs; but we could probably do something similar for internal APIs too and therefore address #230 in the process. Let's get the external API work done first.

BenjamenMeyer:
Once #614 is merged, then I'll submit another PR with all the end-points for the OS-KSCATALOG extension (/endpointTemplates and /tenantid/endpointPointeTemplates) and a few endpoints for OS-KSADM extension (/services).

PR #614, opened 2016-05-11
Enhancement iexternalapitemplatemock

BenjamenMeyer:
Next stage in the infrastructure for supporting external APIs within mimic

Second PR in series for addressing #219


BenjamenMeyer:
Marked `work-in-progress` to see if I have sufficient test coverage. I think I do, but unsure as local detox/coverage does not report fractions of a percent.

BenjamenMeyer:
Woohoo...raised coverage from 99.51% to 99.52% :smile_cat:

mimic/core.py
```diff
@@ -161,9 +189,14 @@ def entries_for_tenant(self, tenant_id, prefix_map, base_uri):
...
-                    prefix_map[endpoint] = api.uri_for_service(
-                        endpoint.region, service_id
-                    )
+                    try:
+                        prefix_map[endpoint] = api.uri_for_service(
+                            endpoint.region, service_id
+                        )
+                    except IndexError:  # pragma:nocover
+                        # ignore APIs that don't support the given region
+                        pass
```
BenjamenMeyer: @glyph I tried to create a test for the above try/except and then remove the pragma:nocover, but I'm not sure exactly how since it requires calling api.uri_for_service() with a different region than expected. I believe the try/except is necessary; but could be wrong. I'd love to write a test if possible. Advice/tips/etc appreciated.

BenjamenMeyer:
FWIW, last PR ( #608) created IExternalAPIMock; the branch name for this PR came from an intermediate thought of doing a mock API version for the EndPoint Templates (IExternalApiTemplateMock), but part way through I realized it could be merged back into IExternalAPIMock so you won't actually see IExternalApiTemplateMock in the code in the final product, just the early commits on the branch.

mimic/model/identity_objects.py
```diff
@@ -0,0 +1,305 @@
...
+        """
+        Initialize an :obj:`ExternalApiStore` for a given service.
+
+        :param text_type service_uuid: unique identifier for the API
+        :param text_type service_name: name of the service being provided,
+            f.e Cloud Files, Identity
+        :param text_type service_type: type of the service being provided,
+            f.e object-store, auth
+        :param iterable api_templates: iterable of templates to during
+            initialization
```
derwolfe: It looks like `api_templates` is missing a word :)

mimic/imimic.py
```diff
@@ -28,13 +28,37 @@ def resource_for_region(region, uri_prefix,
         """
 
 
+class IEndPointTemplate(Interface):
+    """
+    AN :obj:`IEndPointTemplate` provides an interface for the OS-KSCATALOG
+    Admin end-point templates, defining the fields.
```
derwolfe: It looks like there is a mixture of `end-point` and `endpoint`. Would you mind using `endpoint` in the docstrings? It looks to be the chosen spelling convention in mimic to date.

mimic/model/identity_objects.py
```diff
@@ -0,0 +1,305 @@
...
+            raise TypeError(
+                endpoint_template.__class__.__module__ + "/" +
+                endpoint_template.__class__.__name__ +
+                " does not implement IEndPointTemplate"
+            )
+
+    def remove_template(self, template_id):
+        """
+        Add a new template for the external API.
+
```
derwolfe: I believe this docstring should read `Remove a templated endpoint from the external API` or something similar.

mimic/test/test_catalog.py
```diff
@@ -0,0 +1,116 @@
...
+        self.assertEqual(self.tenant, endpoint.tenant_id)
+        self.assertEqual(self.region, endpoint.region)
+        self.assertEqual(self.endpointid, endpoint.endpoint_id)
+        self.assertIsNone(endpoint.prefix)
+        self.assertFalse(endpoint.external)
+        self.assertIsNone(endpoint.complete_url)
+
+    def test_internal_endpoint(self):
+        """
+        Typical creationg of :obj:`Endpoint` for an internal API
```
derwolfe: creationg/creation

mimic/test/test_identity_objects.py
```diff
@@ -0,0 +1,302 @@
...
+        Temporary for code-coverage until end-points are
+        implemented that will actually use these.
+        """
+        class reqMock(object):
+            def setResponseCode(self, code):
+                pass
+
+        bad_request("testing bad request", reqMock())
+        not_found("testing not found", reqMock())
+        forbidden("testing forbidden", reqMock())
```
derwolfe: The `YetToBeDone` test should probably be removed, unless I'm misunderstanding why it is here :-)

mimic/test/test_core.py
```diff
@@ -92,3 +102,236 @@ def cleanup():
...
+            for entry in core.entries_for_tenant(
+                'some-tenant',
+                prefix_map,
+                base_uri
+            )
+        ]
+
+        self.assertEqual(len(catalog_entries), 2)
+
+    def test_entries_for_tenant_external_invalid(self):
```
derwolfe: It doesn't seem like this test is named correctly.

mimic/test/test_identity_objects.py
BenjamenMeyer: This is a couple things that I know I'll need in the next PR and had added. I'd prefer to keep them in this PR and promise they will be used in the next one or removed. In fact, I'll likely be removing `forbidden` in the next PR, but `bad_request` and `not_found` are both utilized; and `YetToBeDone` will be removed in the PR.

The test is there just to ensure sufficient code coverage instead of adding them and letting them just hang out. As noted, it's just temporary.

mimic/test/test_core.py
BenjamenMeyer: So this test was my attempt at trying to testing [this little block of code](https://github.com/BenjamenMeyer/mimic/blob/enhancement_iexternalapitemplatemock/mimic/core.py#L196), though unsuccessfully. See [this comment](https://github.com/rackerlabs/mimic/pull/614#discussion_r62910112) for details. I'm all :ear:'s for solutions.

BenjamenMeyer: Looking at it more closely, I think the try/catch I was concerned about isn't necessary; I'm still going to keep this test, but renaming it as it does provide some value in that there are multiple regions for the external APIs so it still exercises the loop a little differently than the previous few tests.

mimic/core.py
```diff
@@ -164,6 +192,7 @@ def entries_for_tenant(self, tenant_id, prefix_map, base_uri):
                     prefix_map[endpoint] = api.uri_for_service(
                         endpoint.region, service_id
                     )
+
```
derwolfe: It doesn't seem like this newline is necessary; if not, would you mind removing? :-)

mimic/imimic.py
```diff
@@ -46,7 +70,51 @@ def catalog_entries(tenant_id):  # pragma:nocover
...
+        """
+        Update an existing template for the external API.
+
+        :param unicode templates: a :obj:`IEndpointTemplate` to add to the
+            :obj:`IExternalAPIMock` instance
+        """
+
+    def remove_template(template_id):  # pragma:nocover
+        """
+        Add a new template for the external API.
```
derwolfe: This docstring should probably `remove an existing template from the external API` or something similar.

mimic/model/identity_objects.py
```diff
@@ -0,0 +1,305 @@
...
+    """
+    return _identity_error_message("forbidden", message, FORBIDDEN, request)
+
+
+@implementer(IExternalAPIMock, IPlugin)
+class ExternalApiStore(object):
+    """
+    A :obj:`ExternalApiStore` is an object which provides the functionality to list
+    an API that is external to Mimic in the service catalog provided by Mimic.
+    """
```
derwolfe: I'm not sure I understand what the `ExternalApiStore`'s function is based on the docstring - would you mind taking look at the docstring and possibly rewording? If I understand correctly, the `ExternalApiStore` allows users to create new entries in the service catalog for externally hosted services for a specific tenant.

mimic/imimic.py
```diff
@@ -28,13 +28,37 @@ def resource_for_region(region, uri_prefix,
         """
 
 
+class IEndpointTemplate(Interface):
+    """
+    AN :obj:`IEndpointTemplate` provides an interface for the OS-KSCATALOG
+    Admin endpoint templates, defining the fields.
+    """
+    id_key = Attribute("Unique Service ID")
+    region_key = Attribute("Region the End-Point services")
```
derwolfe: Could you change all of the spellings of `End-Point` to `Endpoint`? :)

mimic/test/dummy.py
```diff
@@ -93,36 +96,103 @@ def resource(self):
         return example_resource
 
 
-@implementer(IExternalAPIMock, IPlugin)
-class ExampleExternalAPI(object):
+@implementer(IEndpointTemplate)
+class ExampleEndpointTemplate(object):
     """
-    Example external API the return nothing
+    Example End-Point Template
```
derwolfe: It seems like this is creating an `Example external endpoint template`.

mimic/test/dummy.py
```diff
@@ -93,36 +96,103 @@ def resource(self):
...
+
+    :param text_type name: user-visible name of the service.
+    :param list endpoint_templates: list of endpoint templates to
+        initialize the API store with.
+    :param boolean or None set_enabled: If none, the endpoint templates
+        are used AS-IS. If a boolean type, then it sets all the templates
+        to have the same default accessibility for all tenants.
+
+    Note: The service-type of the first endpoint template is used as the
+        service type for the entire Api Store, and is enforced that all
```
derwolfe: Is `Api Store` capitalized here to reference the `ExternalApiStore` object? If so, it might be clearer to use :obj:`ExternalApiStore`. :-)

mimic/test/dummy.py
```diff
@@ -93,36 +96,103 @@ def resource(self):
...
+        self.versionInfo = (versionInfoURL
+                            if versionInfoURL is not None
+                            else url + '/versionInfo')
+        self.versionList = (versionListURL
+                            if versionListURL is not None
+                            else url + '/versions')
+
+
+def make_example_external_api(name=u"example",
+                              endpoint_templates=[ExampleEndpointTemplate()],
```
derwolfe: Could using mutable default arguments for `endpoint_templates` be an issue here? I don't *think* so, because we are creating a new instance of `ExternalApiStore` each time, but thought I'd ask :-).

mimic/test/fixtures.py
```diff
@@ -37,21 +37,37 @@ def __init__(self, test_case, root, username, password):
...
-            listed in the service catalog
-        :param unicode region: The service catalog region of the desired endpoint
+        Return the publicURL for the given service and region. Note that if
+        there are multiple endpoints for a given region, the first will be
+        returned, and if no region is specified, the first endpoint will be
+        returned.
+        :param unicode service_name: The name of the service for which to get
+            an endpoint as listed in the service catalog
+        :param unicode region: The service catalog region of the desired
+            endpoint
```
derwolfe: Could you amend the docstring to describe the conditions under which a `KeyError` might be raised?

mimic/test/test_catalog.py
```diff
@@ -0,0 +1,116 @@
+from __future__ import absolute_import, division, unicode_literals
+
+from twisted.trial.unittest import SynchronousTestCase
+
+from mimic.catalog import Endpoint, Entry
+
+
+class CatalogEntry(SynchronousTestCase):
+    """
+    Tests for creating a :class:`Entry` object
```
derwolfe: What would you think of modifying the styles of the test docstring to contain assertions instead of descriptions? This can be helpful to other maintainers and users as it should make it clear what the test is intending to test. For example, instead of staying "Absolute mimial creation of :obj:`Entry`", the docstring could say what conditions one should expect to be fulfilled by created a new `Entry` object.

mimic/test/test_core.py
```diff
@@ -92,3 +102,243 @@ def cleanup():
             fake_plugin.dummy_domain_plugin,
             core.domains
         )
+
+
+class CoreApiBuildingTests(SynchronousTestCase):
```
derwolfe: This isn't blocking feedback, but it seems like many of the docstrings in this class could be reflowed to 80 columns, instead of taking multiple lines.

mimic/test/test_core.py
```diff
@@ -92,3 +102,243 @@ def cleanup():
...
+        catalog_entries = [
+            entry
+            for entry in core.entries_for_tenant(
+                'some-tenant',
+                prefix_map,
+                base_uri
+            )
+        ]
+
+        self.assertEqual(len(catalog_entries), 1)
```
derwolfe: Is checking the length of the catalog entries a sufficiently strict test? The concern I have is that this only validates that there is a plugin in the catalog, not that an instance of ExampleAPI is provided by the catalog.

derwolfe: The same question applies to the two following tests as well.

mimic/test/test_external_api.py
```diff
@@ -5,28 +5,30 @@
...
-        self.helper = APIMockHelper(self, [self.eeapi])
+        """
+        Validate that an external API does not provide a Resource
+        for Mimic to support that API within itself.
+        """
+        eeapi = make_example_external_api(
+            name=self.eeapi_name,
+            set_enabled=True
+        )
+        self.assertIsNotNone(eeapi)
```
derwolfe: Is it possible for the return value of `make_example_external_api` to be `None`?

mimic/test/test_external_api.py
```diff
@@ -47,8 +49,12 @@ class TestExternalApiMock(SynchronousTestCase):
...
     """
     def setUp(self):
-        self.eeapi_name = u"example-api"
-        self.eeapi = ExampleExternalAPI(name=self.eeapi_name)
+        self.eeapi_name = u"externalServiceName"
+        self.eeapi = make_example_external_api(
+            name=self.eeapi_name,
+            set_enabled=True
+        )
+        assert(self.eeapi is not None)
```
derwolfe: The question from https://github.com/rackerlabs/mimic/pull/614/files#r63302816 applies here as well. :-)

mimic/test/test_identity_objects.py
```diff
@@ -0,0 +1,302 @@
...
+
+
+class ValidateEndpointTemplates(SynchronousTestCase):
+    """
+    Validate on Endpoint Template Additions
```
derwolfe: I don't understand what this docstring means - would you mind rewording it?

mimic/test/test_identity_objects.py
```diff
@@ -0,0 +1,302 @@
...
+            region=new_region,
+            url=new_url
+        )
+        eeapi = make_example_external_api(
+            name=self.eeapi_name,
+            set_enabled=True,
+            endpoint_templates=[old_eeapi_template]
+        )
+
+        eeapi.update_template(new_eeapi_template)
```
derwolfe: It doesn't look like this test contains an assertion.

mimic/test/test_identity_objects.py
```diff
@@ -0,0 +1,302 @@
...
+            name=self.eeapi_name,
+            uuid=eeapi_template_id
+        )
+        eeapi = make_example_external_api(
+            name=self.eeapi_name,
+            set_enabled=True,
+            endpoint_templates=[eeapi_template]
+        )
+
+        eeapi.remove_template(eeapi_template_id)
```
derwolfe: What should this be asserting?

mimic/test/test_identity_objects.py
```diff
@@ -0,0 +1,302 @@
...
+            set_enabled=True,
+            endpoint_templates=[eeapi_template]
+        )
+        eeapi.enable_endpoint_for_tenant(
+            'some-tenant',
+            eeapi_template_id
+        )
+
+        eeapi.remove_template(eeapi_template_id)
+
```
derwolfe: It looks like this test is missing an assertion

mimic/test/test_identity_objects.py
```diff
@@ -0,0 +1,302 @@
...
+            'some-tenant',
+            eeapi_template_id
+        )
+        eeapi.enable_endpoint_for_tenant(
+            'some-other-tenant',
+            alternate_eeapi_template_id
+        )
+
+        eeapi.remove_template(eeapi_template_id)
+
```
derwolfe: It looks like this test is missing an assertion.

mimic/test/test_identity_objects.py
```diff
@@ -0,0 +1,302 @@
...
+            endpoint_templates=[new_eeapi_template]
+        )
+        eeapi.enable_endpoint_for_tenant(
+            'some-tenant',
+            new_eeapi_template_id
+        )
+        eeapi.disable_endpoint_for_tenant(
+            'some-tenant',
+            new_eeapi_template_id
+        )
```
derwolfe: It looks like this test is missing an assertion.

mimic/test/test_identity_objects.py
derwolfe: On rereading this, it looks like these are intending to test the new functionality introduced by the new `[mimic/model/identity.py`](https://github.com/BenjamenMeyer/mimic/blob/510d39179b8c940e61724e331a1b689cfb90a704/mimic/model/identity_objects.py) file. It seems like it might be more helpful/informative to add tests for the behavior demonstrated by these functions, instead of calling them to generate coverage. What do you think?

mimic/core.py
```diff
@@ -95,6 +102,27 @@ def add_api(self, api):
...
+    def get_external_api(self, api_name):
+        """
+        Access an API instance for an external API.
+
+        Note: Internally hosted APIs are not modifiable at run-time
+            so this only returns access to the Externally hosted API.
+
+        :param text_type api_name: the name of the API instance
+            e.g Cloud Files
+        :returns: The :obj:`IExternaAPIMock` instance supporting the
```
derwolfe: `/IExternaAPIMock/IExternalAPIMock`

mimic/test/test_identity_objects.py
```diff
@@ -0,0 +1,302 @@
...
+
+
+class EndpointTemplateOperations(SynchronousTestCase):
+    """
+    Tests for creating a :class:`MimicCore` object with apis
+    """
+    def setUp(self):
+        self.eeapi_name = u"externalServiceName"
+
+    def test_uri_for_service_with_invalid_region(self):
```
derwolfe: Could you explain to me what this test covers? (I ask because it isn't immediately obvious to me why it isn't in mimic/test/test_core.py, but I could definitely be missing something :-) )

derwolfe:
@BenjamenMeyer Thanks for working on this! Please resubmit for review once you've had a chance to either answer the questions I raised or address the comments :)

mimic/test/test_external_api.py
derwolfe: I ask because there are several test assertions and python `asserts` where this is being checked, yet it doesn't seem like returning None should be a possibility. :-)

mimic/test/test_identity_objects.py
BenjamenMeyer: I'll have that in the next PR; this PR doesn't contain the actual RESTful API endpoints, it's just infrastructure. The next PR will have the endpoints that all this infrastructure supports. Thus why I'd like to just leave it as is now, and then use or remove in the next PR.

mimic/imimic.py
BenjamenMeyer: Will correct.

mimic/model/identity_objects.py
BenjamenMeyer: The `ExternalApiStore` represents a service (e.g Cloud Files)  and hosts all the endpoint templates for that service (e.g ORD vs IAD vs SYD...). By default the templates are disabled and *can* be enabled on a per-tenant basis, or they can be enabled *globally* for all tenants.

I'll update the description.

mimic/imimic.py
BenjamenMeyer: Will do.

mimic/test/dummy.py
BenjamenMeyer: No, it doesn't cause an issue. If you care what the name or templates should be, then they should be specified. If not, then use the defaults.

mimic/test/dummy.py
BenjamenMeyer: ATM, all endpoint templates are only for external services; but I'll note it accordingly in case we want to change that in the future for internally hosted services too.

mimic/test/test_external_api.py
BenjamenMeyer: Correct, it should not be possible for it to return None unless Python failed to create the object. I could go either way on whether to keep the assertions - I typically try to check *everything* and assume *nothing*.

mimic/test/test_identity_objects.py
BenjamenMeyer: I'll add some checks; but normally if I'm running through exception checks then the lack of the exception is the assertion itself as an exception would cause the test to fail.

mimic/test/test_identity_objects.py
BenjamenMeyer: It's not in `mimic/test/test_core.py` b/c it is testing the `ExternalApiStore`'s version of `uri_for_service`. Instead of putting logic into `core`'s version of the function to differentiate, I made it part of the `ExternalApiStore` which makes it a little cleaner and obvious since `core` then has a call of `api.uri_for_service` vs `self.uri_for_service`.

I'll add a doc-string though.

mimic/test/test_core.py
BenjamenMeyer: Reformatted the lines where I could and added docstrings where missing.

mimic/test/test_identity_objects.py
derwolfe: Thanks for your explanation :).

Tests without assertions are a pattern I find difficult with which to work. I tend to view tests themselves as assertions about the system - the assertions in those tests are absolutely vital in communicating that the system will exhibit a certain behavior given the conditions that a test sets up. While it can be clear why an exception style test fails, it isn't immediately obvious what conditions should result in such a test passing.

mimic/test/test_core.py
```diff
@@ -92,3 +102,312 @@ def cleanup():
...
+            )
+        )
+
+    def test_service_with_region_internal(self):
+        """
+        Validate that an external service does not provide an internal service
+        resource.
+        """
+        iapi = ExampleAPI()
+        self.assertIsNotNone(iapi)
```
derwolfe: It looks like this assertion is repeated in several places, would it make sense to extract this into a single test or method that sets up the `ExampleApi` instance and asserts that it isn't `None`? Doing so would allow you to remove this from each of the tests.

BenjamenMeyer:
@glyph @derwolfe comments addressed.

mimic/test/test_core.py
```diff
@@ -92,3 +102,312 @@ def cleanup():
...
+    def test_service_with_region_external(self):
+        """
+        Validate that an external service does not provide an internal
+        service resource.
+        """
+        eeapi = make_example_external_api(
+            name=self.eeapi_name,
+            set_enabled=True
+        )
+        self.assertIsNotNone(eeapi)
```
derwolfe: It looks like this assertion is repeated several times - would you mind either extract it into a setup method and or creating a test showing that the object returned by `make_external_api` is not none?

mimic/test/test_core.py
BenjamenMeyer: If desired I could. But would that really add value? I'm not sure.

derwolfe: The value add that I see is for DRY-ness.

mimic/test/test_core.py
BenjamenMeyer: `make_external_api` is solely a test framework helper function; it's not used in actual Mimic. Perhaps an assert in the helper function itself?

glyph: Yeah; making it take a `TestCase` as an argument is a good way to make that clean.

glyph: (As a bonus, then it can add any clean-ups it needs to.)

mimic/test/dummy.py
```diff
@@ -93,36 +96,118 @@ def resource(self):
...
+        self.versionInfo = (versionInfoURL
+                            if versionInfoURL is not None
+                            else url + '/versionInfo')
+        self.versionList = (versionListURL
+                            if versionListURL is not None
+                            else url + '/versions')
+
+
+def make_example_internal_api(case, response_message="default message",
+                              regions_and_versions=[('ORD', 'v1')]):
```
derwolfe: `regions_and_versions` is mutable, I'd suggest setting it to `None`, then setting the default value in the body, e.g.:
```python
if regions_and_versions is None:
    regions_and_versions = [('ORD', 'v1')]
iapi = ExampleApi(...
```

mimic/test/dummy.py
```diff
@@ -93,36 +96,118 @@ def resource(self):
...
-    def uri_for_service(self, region, service_id):
-        """
-        Return the URI for the service in the given region.
-        """
-        for r, u in [(each[0], each[2]) for each in self.regions_and_versions]:
-            if r == region:
-                return u
-        return u""
+def make_example_external_api(case, name=u"example",
+                              endpoint_templates=[ExampleEndpointTemplate()],
```
derwolfe: I'd suggest the same pattern as https://github.com/rackerlabs/mimic/pull/614/files#r65996089 be used here with `endpoint_templates`, otherwise the behavior of the function might be unexpected.

mimic/test/test_catalog.py
```diff
@@ -0,0 +1,126 @@
...
+        self.assertEqual(endpoint.complete_url, self.external_api_url)
+
+        uri = endpoint.url_with_prefix(self.request_prefix)
+        self.assertEqual(uri, self.external_api_url)
+
+    def test_external_endpoint_invalid(self):
+        """
+        Validate that trying to creating an :obj:`Endpoint` for an external API
+        (setting `external` to `True` while *not* setting the `complete_url`
+        will raise a `ValueError` exception.
```
derwolfe: It looks like this docstring might be incorrect/incomplete. :-)

mimic/imimic.py
```diff
@@ -28,13 +28,37 @@ def resource_for_region(region, uri_prefix,
...
+    """
+    AN :obj:`IEndpointTemplate` provides an interface for the OS-KSCATALOG
+    Admin endpoint templates, defining the fields.
+    """
+    id_key = Attribute("Unique Service ID")
+    region_key = Attribute("Region the Endpoint services")
+    type_key = Attribute("Service type, example: object-store")
+    name_key = Attribute("Service Name, example: Cloud Files")
+    enabled_key = Attribute("Whether the Endpoint is globally available")
+    publicURL = Attribute("Internet Facing URL")
```
derwolfe: It looks like the attributes from here downward are using camel-casing. Please change these so that they use snake casing to match the other attributes. This will affect the implementation of the `ExampleEndpoint` as well.

mimic/test/dummy.py
BenjamenMeyer: @derwolfe thanks for the reminder; I've got that fixed in my next PR. I can back port it here.

mimic/test/dummy.py
BenjamenMeyer: Yes, I've got that fixed too. Again, can backport.

mimic/imimic.py
BenjamenMeyer: will update

mimic/test/test_catalog.py
BenjamenMeyer: Reworded, it was correct but I can see how it could be confusing so cleared it up.

BenjamenMeyer:
I'll update this branch after #618 and #639 are merged.

BenjamenMeyer:
@derwolfe should be all sync'd up with master and ready for review again.

glyph:
Looks to me like this has addressed all feedback in outstanding fashion.  Sorry for the long delays on review - vacation and conferences have kept me away from code review for several weeks.

Going to have one last look but I anticipate landing momentarily.

mimic/imimic.py
```diff
@@ -28,13 +28,37 @@ def resource_for_region(region, uri_prefix,
...
+    version_list = Attribute("URL to determine which version is hosted "
+                             "by the endpoint")
+
+
 class IExternalAPIMock(Interface):
     """
     An :obj:`IExternalAPIMock` provides an API via an entry for services
     external to Mimic in a tenant's service catalog.
     """
+    id_key = Attribute("Unique Service ID")
```
glyph: If `id_key` and `name_key` must both me unique, then what is the difference?

mimic/imimic.py
```diff
@@ -28,13 +28,37 @@ def resource_for_region(region, uri_prefix,
...
     name_key = Attribute("Name of the External API. Primarily used for "
                          "the internal mappings. Must be unique.")
+    type_key = Attribute("Type of the Templatized External API. Primarily "
+                         "used for the internal mappings. Must be unique.")
```
glyph: In general, one can't add attributes or methods to existing public interfaces that have been present in a release, because external code may already have been implementing these interfaces, and adding stuff like this makes their implementation into a lie.

However, I believe @derwolfe is the only one who might be using this interface yet in practice, so I'll have to defer to him as to whether this is acceptable.

mimic/imimic.py
```diff
@@ -46,7 +70,51 @@ def catalog_entries(tenant_id):  # pragma:nocover
...
+    def list_templates():  # pragma:nocover
+        """
+        List the available templates.
+        """
+
+    def add_template(endpoint_template):  # pragma:nocover
+        """
+        Add a new template for the external API.
+
+        :param unicode templates: a :obj:`IEndpointTemplate` to add to the
```
glyph: The parameter is called `endpoint_template` not `templates`.

(Clearly we should have set up doc linting as a travis job...)

glyph: Also: this should probably be `six.text_type` and not `unicode`, since `unicode` doesn't exist on python 3.

mimic/imimic.py
```diff
@@ -46,7 +70,51 @@ def catalog_entries(tenant_id):  # pragma:nocover
...
+            :obj:`IExternalAPIMock` instance
+        """
+
+    def update_template(endpoint_template):  # pragma:nocover
+        """
+        Update an existing template for the external API.
+
+        :param unicode templates: a :obj:`IEndpointTemplate` to add to the
```
glyph: again, the parameter is `endpoint_template`

mimic/imimic.py
glyph: It seems like this might want to be a different interface, with an adapter from `IExternalAPIMock` to `IExternalAPIMockWithTemplating`?

mimic/imimic.py
```diff
@@ -46,7 +70,51 @@ def catalog_entries(tenant_id):  # pragma:nocover
 
     def uri_for_service(region, service_id):  # pragma:nocover
         """
-        Get the URI for the service in a given region
+        Get the URI for the service in a given region.
+        """
+
+    def list_tenant_endpoints(tenant_id):  # pragma:nocover
```
glyph: The purpose of an `Interface` is clear, explicit documentation of a system boundary, so `:param:` and `:return:` documentation is particularly important here; please add `:param:` and `:return:` for every one of these.

mimic/imimic.py
```diff
@@ -46,7 +70,51 @@ def catalog_entries(tenant_id):  # pragma:nocover
...
+        """
+        List the tenant specific endpoints.
+        """
+
+    def enable_endpoint_for_tenant(tenant_id, template_id):  # pragma:nocover
+        """
+        Enable an endpoint for a specific tenant.
+        """
```
glyph: note for tracking `:param:`, `:return:` docstring

mimic/imimic.py
```diff
@@ -46,7 +70,51 @@ def catalog_entries(tenant_id):  # pragma:nocover
...
+
+    def disable_endpoint_for_tenant(tenant_id, template_id):  # pragma:nocover
+        """
+        Disable an endpoint for a specific tenant.
+        """
```
glyph: note for tracking `:param:`, `:return:` docstring

mimic/imimic.py
```diff
@@ -46,7 +70,51 @@ def catalog_entries(tenant_id):  # pragma:nocover
...
+
+    def list_templates():  # pragma:nocover
+        """
+        List the available templates.
+        """
```
glyph: note for tracking `:param:`, `:return:` docstring

glyph:
@BenjamenMeyer - OK, not quite close enough to just hit the green button, but pretty close; I don't have much feedback on any of the remaining code.  If @derwolfe can opine about the interface expansion and you can fill in a few more interface docstrings, we're good to go.

mimic/core.py
```diff
@@ -31,7 +35,7 @@ class MimicCore(object):
         that will show up in the Service Catalog
     """
 
-    def __init__(self, clock, apis, domains=()):
+    def __init__(self, clock, apis, domains=(), external_apis=()):
```
derwolfe: The `external_apis` parameter should be documented as part of `__init__`'s docstring.

derwolfe:
@glyph I may be misunderstanding your question regarding interface expansion. Do you mean the expansion of `MimicCore`'s public interface, or the addition of the new `IEndpointTemplate` and expansion of the `IExternalTemplateMock` interfaces? 

It didn't jump out to me that either would be problematic. Was there something you had in mind?

glyph:
@derwolfe - Specifically the expansion of `IExternalAPIMock` interface, which was the pre-existing one. Adding new interfaces is always fine :).  The question is: do you have existing `IExternalAPIMock` implementations you need to continue to support on new mimics?

mimic/imimic.py
BenjamenMeyer: It's part of the Keystone spec. ID is a UUID kind of thing, while Name is what the users call it (e.g Cloud Backup).

mimic/imimic.py
BenjamenMeyer: @glyph I should be the only one using these interfaces, `IExternalAPIMock` was something I introduced in a previous PR (#608) to differentiate the services this PR (and the next) will be managing versus the `IAPIMock` that has always been there. This was a refinement of the previous PR as the functionality matured. All of the `IExternalAPIMock` functionality will be done using the templating functionality (next PR will add the external RESTful API to support).

mimic/imimic.py
BenjamenMeyer: agreed will do.

mimic/imimic.py
BenjamenMeyer: Agreed; `templates` :arrow_right: `endpoint_template` was probably just missed in an update. Will correct.

mimic/core.py
BenjamenMeyer: it's actually not used, so removing it

BenjamenMeyer:
@glyph I believe I have everything addressed; most of it was just documented at a lower layer (the actual implementing object).

derwolfe:
@glyph - Thanks for clarifying. I had thought that expanding `IExternalAPIMock` would be ok, considering that @BenjamenMeyer had introduced it explicitly for this group of PRs. However, I hadn't thought about the possibility that someone, while this feature was being developed, may have based some work on the `IExternalApiMock` that was unofficially a work in progress. 

That said, I'm not yet using this interface, so it won't affect me :). I think we should be safe expanding the interface and :+1: for going ahead with out creating an adapter. I don't think that there are any other developers currently basing work off of the interface. But, if you disagree, I think that the adapter approach is the correct solution.

In the future, do you think that we should hide in-progress interfaces like this or document them in such a  way that they can be merged in, but treated as a work in progress?

BenjamenMeyer:
@glyph @derwolfe added some documentation to the relevant objects to mark as `work in progress`.

glyph:
OK.  I think we're in good shape to land now.

The thing about interface expansion is like a general "things you have to look out for in open source with public interfaces" problem; it's not something that we necessarily need to be super strict about on Mimic.  I think I temporarily got some wires crossed between `IAPIDomainMock` and `IExternalAPIMock` which is why I thought @derwolfe needed to opine :)

glyph:
In the future, let's try to keep changes a bit smaller and more tightly described - but I think this is now good to go.  Thanks for all the work on it.

BenjamenMeyer:
@glyph thanks...yes, generally I do try to keep PRs smaller, but with big new features like this...any how, just wanted to note I'll have one more big PR for the RESTful interfaces per Identity/Keysone Extension specs; after that it should be just small stuff. I'll try to file that one tomorrow. Thanks!

Issue #511, opened 2016-01-27
/identiy/v2.0 responds 404 and nova list command failed


nyasukun:
nova list command catch following error.
```
# nova list
No handlers could be found for logger "keystoneauth.identity.generic.base"
ERROR (DiscoveryFailure): Could not determine a suitable URL for the plugin
```

This is caused by there is no resource at /identiy/v2.0.
/identity/v2.0 should show version details.
Please see: http://developer.openstack.org/api-ref-identity-v2.html#showVersionInfo-v2.0

Benjamin91300:
This should be solved as my commit #618 and #619 have been merged.

BenjamenMeyer:
@Benjamin91300 would you please verify and post output from one of the openstack clients (e.g python-openstackclient, pyrax, etc).

PR #618, opened 2016-05-19
Implement keystone get_version

Benjamin91300:
Recent novaclient versions and openstackclient require keystone
resource /identity/v2.0. This commit provides it.

PR #619, opened 2016-05-19
Implement nova get_version

Benjamin91300:
Recent novaclient versions and openstackclient require resource
/compute/v2. This commit provides it.

### DIFF
--- a/mimic/test/test_nova.py
+++ b/mimic/test/test_nova.py
@@ -15,6 +15,7 @@ import treq
 
 from twisted.trial.unittest import SynchronousTestCase
 
+from mimic.canned_responses.nova import get_version_v2
 from mimic.test.helpers import json_request, request, request_with_content, validate_link_json
 from mimic.rest.nova_api import NovaApi, NovaControlApi
 from mimic.test.behavior_tests import (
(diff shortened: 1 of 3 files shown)


BenjamenMeyer:
Along with #618 this should resolve #511

mimic/test/test_nova.py
```diff
@@ -184,6 +193,23 @@ def setUp(self):
             create_server(self.helper, name=self.server_name))
         self.server_id = self.create_server_response_body['server']['id']
 
+    def test_get_version_nova(self):
+        """
+        Check response when listing available version of nova.
```
glyph: Could you add a citation for this that points at Nova's docs?

glyph:
Please just address the minor test doc issue and land.

mimic/test/test_nova.py
BenjamenMeyer: http://developer.openstack.org/api-ref-compute-v2.1.html

Benjamin91300:
@glyph, @BenjamenMeyer:
I added a reference to the nova doc in docstring.

I hope it is what you expected.

BenjamenMeyer:
@Benjamin91300 I would add it to the handler of the response itself, not simply the test (https://github.com/rackerlabs/mimic/pull/619/files#diff-623976324cdaaa34626dd686efc298b9R34).

Once that and the current merge conflict are resolved then I'll merge it as it otherwise LGTM.

BenjamenMeyer:
:+1:

PR #620, opened 2016-05-20
Enhancement: Add Service User for Cloud Backup

BenjamenMeyer:
Add a service user to the listing for Cloud Backup to use.


BenjamenMeyer:
@glyph I've added this and am trying to use it now. Is there anything I'm missing? If not, please merge.

glyph:
The lack of a test is a little concerning, but this is just additional data, and we should probably look at having some non-hard-coded way of dealing with this later.  Would you mind filing an issue for that?  In the meanwhile, merging seems fine :).

PR #638, opened 2016-06-07
Bug Fix: Validating existing tokens

BenjamenMeyer:
Validating an existing token assumed that the tenant-id was specified. If it was not (tenant_id = None) then a new session was being created. This caused problems for tools like Repose.

- Fixed the validate_token function
- Added an explicit test that operates like Repose does.

### DIFF
--- a/mimic/rest/identity_api.py
+++ b/mimic/rest/identity_api.py
@@ -294,8 +294,7 @@ class IdentityApi(object):
         """
         Creates a new session for the given tenant_id and token_id
         and always returns response code 200.
-        Docs: http://developer.openstack.org/api-ref-identity-admin-v2.html#
-              admin-validateToken
+        Docs: http://developer.openstack.org/api-ref-identity-admin-v2.html#admin-validateToken  # noqa
         """
         request.setResponseCode(200)
         session = None
--- a/mimic/rest/identity_api.py
+++ b/mimic/rest/identity_api.py
@@ -294,13 +294,26 @@ class IdentityApi(object):
         """
         Creates a new session for the given tenant_id and token_id
         and always returns response code 200.
-        Docs: http://developer.openstack.org/api-ref-identity-v2.html#admin-tokens
+        Docs: http://developer.openstack.org/api-ref-identity-admin-v2.html#
+              admin-validateToken
         """
         request.setResponseCode(200)
+        session = None
+
+        # Attempt to get the session based on tenant_id+token if the optional
+        # tenant_id is provided; if tenant_id is not provided, then just look
+        # it up based on the token.
         tenant_id = request.args.get(b'belongsTo')
         if tenant_id is not None:
             tenant_id = tenant_id[0].decode("utf-8")
-        session = self.core.sessions.session_for_tenant_id(tenant_id, token_id)
+            session = self.core.sessions.session_for_tenant_id(
+                tenant_id, token_id)
+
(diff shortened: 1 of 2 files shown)


BenjamenMeyer:
@glyph this allows Mimic to work with the keystone-v2 authentication filter in Repose 7.x/8.x.

mimic/rest/identity_api.py
```diff
@@ -294,13 +294,26 @@ def validate_token(self, request, token_id):
         """
         Creates a new session for the given tenant_id and token_id
         and always returns response code 200.
-        Docs: http://developer.openstack.org/api-ref-identity-v2.html#admin-tokens
+        Docs: http://developer.openstack.org/api-ref-identity-admin-v2.html#
```
glyph: Is it actually valid ReST to break a line this way?

mimic/test/test_identity_auth.py
```diff
@@ -826,6 +826,53 @@ def test_response_for_validate_token(self):
...
+            self, root, username=user["username"], password=user["password"])
+
+        (admin_response, admin_data) = authenticate_with_username_password(
+            self, root, username=admin["username"], password=admin["password"])
+
+        user_token = user_data['access']['token']['id']
+        user_tenant = user_data['access']['token']['tenant']['id']
+
+        admin_token = admin_data['access']['token']['id']
+        admin_tenant = admin_data['access']['token']['tenant']['id']
```
glyph: Minor stylistic nitpick: given that you're repeating the same 4 lines for user and admin, maybe refactoring into a short utility function could make the test shorter?  This could also eliminate the variables `user` and `admin`, leaving you with osmething like:

```python
user_token, user_tenant = tenant_and_token("user123456", "654321resu")
admin_token, admin_tenant = tenant_and_token("admin", "nimda")
```
which could take 20 lines (including blanks) down to 2; seems like a worthwhile refactoring :).

The main reason I propose this is that the more bound names there are in a particular scope, the more carefully one must read it, and the more stuff you have to remember as you work your way down.  It's not the pure line count that I'm concerned about so much as the fact that as I'm reading this test I have to keep asking myself "is `user_response` used again?  do I need to know about `admin_data`?"

glyph: (Sorry, given that ordering, obviously my sample function should have been called `token_and_tenant`...)

glyph:
@BenjamenMeyer - Once again, looks good modulo some minor nitpicking.

mimic/rest/identity_api.py
BenjamenMeyer: Not really, but there's no way to break a URI that I'm aware of and the linter will otherwise complain. Open to suggestions.

mimic/test/test_identity_auth.py
BenjamenMeyer: @glyph sounds reasonable; will update.

mimic/rest/identity_api.py
glyph: Hrm. The linter shouldn't complain; https://github.com/PyCQA/pycodestyle/issues/224 should have addressed that.  Did you actually see linter complaints or did you just assume it would bug you because of a hard-coded length limit? :)

BenjamenMeyer: I probably saw it; I did get it fixed though `# noqa` did the job.

PR #639, opened 2016-06-09
Enhancement osksadm credentials also

BenjamenMeyer:
Mimic supported looking up the API Key of a user via `/v2.0/users/<userid>/OS-KSADM/credentials/RAX-KSKEY:apiKeyCredentials`; however, tools like `pyrax` had support for doing the generic credentials lookup via `/v2.0/users/<userid>/OS-KSADM/credentials` which Mimic was not supporting. This adds the additional functionality so that either method can be used. It also includes a separate API key when doing so to aid with debugging, and also provides a similar password credentials.

### DIFF
--- a/mimic/rest/identity_api.py
+++ b/mimic/rest/identity_api.py
@@ -262,12 +262,17 @@ class IdentityApi(object):
         if user_id in self.core.sessions._userid_to_session:
             username = self.core.sessions._userid_to_session[user_id].username
             password = '9d5ed678fe57bcca610140957afab571'  # echo -n B | md5sum
+            apikey = '0d61f8370cad1d412f80b84d143e1257'  # echo -n C | md5sum
             return json.dumps(
                 {
                     'credentials': {
                         'passwordCredentials': {
                             'username': username,
                             'password': password
+                        },
+                        'RAX-KSKEY:apiKeyCredentials': {
+                            'username': username,
+                            'apikey': apikey
                         }
                     }
                 }
--- a/mimic/rest/identity_api.py
+++ b/mimic/rest/identity_api.py
@@ -5,10 +5,13 @@ Defines get token, impersonation
 
 from __future__ import absolute_import, division, unicode_literals
 
+import binascii
 import json
+import os
 import time
 
 import attr
+from six import text_type
 
 from twisted.python.urlpath import URLPath
 
@@ -186,11 +189,25 @@ class IdentityApi(object):
     :ivar core: an instance of :class:`mimic.core.MimicCore`
     :ivar registry_collection: an instance of
         :class:`mimic.model.behaviors.BehaviorRegistryCollection`
+    :ivar apikey_length: the string length of any generated apikeys
     """
     core = attr.ib(validator=attr.validators.instance_of(MimicCore))
     registry_collection = attr.ib(
         validator=attr.validators.instance_of(BehaviorRegistryCollection))
     app = MimicApp()
+
+    apikey_length = 32
+
+    @classmethod
+    def make_apikey(cls):
+        """
+        Generate an API key
+        """
+        # length of the final APIKey value
+        generation_length = int(cls.apikey_length / 2)
+        return text_type(
+            binascii.hexlify(os.urandom(generation_length)).decode('utf-8')
+        )
 
     @app.route('/v2.0/tokens', methods=['POST'])
     def get_token_and_service_catalog(self, request):
@@ -261,17 +278,10 @@ class IdentityApi(object):
         """
         if user_id in self.core.sessions._userid_to_session:
             username = self.core.sessions._userid_to_session[user_id].username
-            password = '9d5ed678fe57bcca610140957afab571'  # echo -n B | md5sum
-            apikey = '0d61f8370cad1d412f80b84d143e1257'  # echo -n C | md5sum
+            apikey = self.make_apikey()
             return json.dumps(
                 {
                     'credentials': [
-                        {
-                            'passwordCredentials': {
-                                'username': username,
-                                'password': password
-                            }
-                        },
                         {
                             'RAX-KSKEY:apiKeyCredentials': {
                                 'username': username,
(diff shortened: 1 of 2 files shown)


mimic/test/test_identity_auth.py
```diff
@@ -684,6 +684,58 @@ def test_rax_kskey_apikeycredentials(self):
...
+            "/identity/v2.0/users/1/OS-KSADM/credentials"
+        ))
+        self.assertEqual(response.code, 404)
+        self.assertEqual(
+            json_body['itemNotFound']['message'], 'User 1 not found')
+        creds = {
+            "auth": {
+                "passwordCredentials": {
+                    "username": "HedKandi",
+                    "password": "Ministry Of Sound UK"
```
glyph: These sample credentials look like an actual company and a concept related to that company.  Could you make this a bit more neutral?

mimic/rest/identity_api.py
```diff
@@ -250,6 +250,43 @@ def get_users_details(self, request):
...
+        """
+        if user_id in self.core.sessions._userid_to_session:
+            username = self.core.sessions._userid_to_session[user_id].username
+            password = '9d5ed678fe57bcca610140957afab571'  # echo -n B | md5sum
+            apikey = '0d61f8370cad1d412f80b84d143e1257'  # echo -n C | md5sum
+            return json.dumps(
+                {
+                    'credentials': [
+                        {
+                            'passwordCredentials': {
```
glyph: When I saw this part - `passwordCredentials` - I did a double-take.  ''Surely'' you must be mistaken and OpenStack would not actually have an extension for saving your plaintext credentials and relaying them back to you.  Alas, it seems that is accurate.  But at least, it seems the Rackspace cloud won't give them back.  Do you actually need `passwordCredentials` in here, or were you just leaving them there because the API docs said they'd be there?  Actually returning this value would require a badly misconfigured authentication backend that would write your credentials out in plain text, so I think we should default to not returning it.

mimic/rest/identity_api.py
```diff
@@ -250,6 +250,43 @@ def get_users_details(self, request):
...
+    def get_user_credentials_osksadm(self, request, user_id):
+        """
+        Support, such as it is, for the credentials call.
+
+        reference: http://developer.openstack.org/api-ref-identity-v2-ext.html
+            #listCredentials
+        """
+        if user_id in self.core.sessions._userid_to_session:
+            username = self.core.sessions._userid_to_session[user_id].username
+            password = '9d5ed678fe57bcca610140957afab571'  # echo -n B | md5sum
```
glyph: md5 is not a great hash algorithm.  Sometimes the presence of the string "md5sum" in a codebase is enough to raise security auditors' hackles.  Also, it's just 32 random hex digits, so it doesn't really matter whether they were the output of `md5sum` or not.

On that note, should they really even be the _same_ magic value every time?  As much as possible I'd like to avoid littering our implementation with magic constants.  You go to some lengths in the assertions below to avoid asserting about a specific value, so `binascii.hexlify(os.urandom(16))` ought to serve this purpose somewhat better than a constant...

glyph:
@BenjamenMeyer - this is mostly pretty good; if you can address the inline comments I made to your own satisfaction I think you can go ahead and land.

mimic/test/test_identity_auth.py
BenjamenMeyer: @glyph I'll update. I copied them from https://github.com/BenjamenMeyer/mimic/blob/0ccbeed5dd5e2cac98ddd23984a810f3d4e58fdb/mimic/test/test_identity_auth.py#L650, so I'll update that too.

mimic/rest/identity_api.py
BenjamenMeyer: @glyph I was primarily following the API docs which I think we should be accurate to. If `mimic` had a way of storing that for the user, then it would return what was used instead of some random string.

That said, I'm not sure off hand. I know Cloud Backup does some stuff where we have to actually authenticate as the user, but I'm not sure off hand if we use password or apikey when doing so. When they asked everyone to use impersonation we ended up knocking down Identity; so this bit is pretty important. I know v2 is using apikey; I'll have to check regarding v1.

mimic/rest/identity_api.py
BenjamenMeyer: @glyph sounds good; I just did the same thing that was already in place for the APIkey specific listing. I'll update them all.

mimic/rest/identity_api.py
glyph: Reading my comment back I guess it doesn't make a lot of sense.

The ability to give back a plain-text password means that the system must _store_ plain-text passwords.  A well-configured system should only ever be storing password hashes, making it impossible to give back a plain-text password.  That's what I'm hoping we're doing, and the fact that I can't find a way to get `passwordCredentials` back on my own accounts strongly suggests that we are.

Given the way that API keys work in OpenStack, it does actually have to store _them_ in plain text (which, IMHO, is still a design error: API keys should be generated, given back to the user, then hashes of them should be stored, and if you need a new API key you should have to request a new one, not look up the old one, but it's a less serious error because presumably randomly-generated credentials can't be shared with _other_ systems).

BenjamenMeyer: @glyph I removed the `passwordCredentials` as I didn't need them, and the API Key is now completely randomly generated. I thought about perhaps making it randomly generate at the class level so you could do the whole get/reset thing, but then just left it as it was. Check the latest commit and see what you think. I can move it around a bit if you like; it's at least easy enough to do now. :smile_cat:

glyph:
/me wearily checks https://codecov.statuspage.io to see what's going on *this* time, but sees nothing...

glyph:
Hey @codecov-io - any idea why this PR's coverage isn't getting reported?

BenjamenMeyer:
@glyph let's merge PR #618 before this one so @Benjamin91300 doesn't have to do another rebase.

BenjamenMeyer:
@glyph merged latest from trunk; ready to review again as soon as it passes gates.

BenjamenMeyer:
Given:
- https://github.com/rackerlabs/mimic/pull/639#issuecomment-225781324 and associated labels
- resolutions to above satisfied by https://github.com/rackerlabs/mimic/pull/639/commits/60dc50806cb3b5f4410d4644983e18c473bc1ef1
- only master merges since for sake of github's merge functionality

It seems like this branch is ready for merge.

BenjamenMeyer:
@glyph `codecov` seems stuck again.

@codecov-io

glyph:
Looks like codecov came unstuck?  Thanks for taking care of this :)

PR #640, opened 2016-06-14
upgrade pypy version

glyph:
requires.io is failing; hopefully this convinces it to succeed.


glyph:
Time to try this again.

PR #646, opened 2016-06-21
Bug Fix: Handle object names in swift properly

### LINKED ISSUE #645: Swift mock doesn't support sub-paths for objects
While testing Cloud Backup API I found that the container would be created but then a 404 would be returned when trying to upload the object. Turns out it was due to the URL formation.

Swift has a very greedy object naming rule. Essentially the URL is `/v1/<container>/<objectname>` and `<objectname>` may also contain additional `folders` to separate data.

In my case, `/v1/CloudBackupLogs` was being created, but then `/v1/CloudBackupLogs/<UUID>/<timestamp>.gz`, the extra layer of the `<UUID>` caused the Swift Mock to fail. My guess is that the `app.route` did not find a match against the URL.

BenjamenMeyer:
- Swift: object names are greedy and can contain additional
  directory names.
- Containers: Support for HEAD, DELETE
- Objects: Support for HEAD, DELETE
- corrected container meta-data so number of objects and byte-counts are properly retrieved
- refactored the Swift API tests to be more comprehensive
  - split Swift API tests into three groups: generic, container, object
  - refactored to minimize redundancy in code

Closes #645


BenjamenMeyer:
I've got another branch with some more functionality for Swift and improved test coverage for Swift API that I'm going to fold into this, removing Ready for Review for the time being, though comments on current work will be appreciated if available.

BenjamenMeyer:
@glyph There is admittedly a lot of redundancy in the tests; however, this should be temporary as the proper method of validating the Swift API functionality should be to look at the underlying infrastructure (model, etc) to validate the changes instead of using other calls (that might also be broken or incomplete) to retrieve the information. I would be glad to make this change as part of this PR if someone could tell me how to find the Swift API instances so that I can do the analysis, otherwise I'll file an Issue so this gets resolved in the future.

BenjamenMeyer:
@glyph sorry for the size...the majority of it is testing. It should also make the swift functionality mostly complete outside of Large Object support.

BenjamenMeyer:
FYI - I'm going to have a small follow-on to this one to add POST and HEAD at the Swift Account Level.

mimic/test/test_swift.py
```diff
@@ -119,74 +313,617 @@ def test_get_container(self):
...
         )
-        self.assertEqual(container_response.code, 404)
+        self.assertIsNone(
+            container_response.headers.getRawHeaders(
+                "X-Container-Bytes-Used")
+        )
+
+    def test_get_container_non_empty(self):
+        """
+        Test retrieving a non-empty container
```
glyph: For future reference, please avoid using terms like "test" and "validate" in test docstrings - https://jml.io/pages/test-docstrings.html has a good primer.

mimic/test/test_swift.py
```diff
@@ -119,74 +313,617 @@ def test_get_container(self):
...
+
+        # the container should have some information in it
+        # the names of the object's don't matter, but having some
+        # dynamic content will help prove it's all working correctly
+        # so generate a series of objects names to use
+        object_paths = [
+            self.object_path,
+        ]
+        object_path_prefix = b""
+        for i in range(10):
```
glyph: Control structures (loops, conditionals, etc) in tests are generally a bad idea, because if the test starts failing, this makes it much harder to quickly see what exact property was failing to verify.  Also, control structures are logic, and logic can have bugs, much more readily than a linear set of constants can have bugs.  However, I can see why you would want to have a bit of generative data here to ensure the various data structures you're populating are not interfering with each other, and so on.  Just something to keep in mind for the future.

mimic/test/test_swift.py
```diff
@@ -119,74 +313,617 @@ def test_get_container(self):
...
+            object_uri = self.uri + b"/" + object_path
+            self.put_object(
+                object_path=object_uri,
+                headers={b"content-type": [b"text/plain"]},
+            )
+
+            container_response, container_contents = self.get_container()
+
+            # Validate container meta-data
+            self.assertEqual(
```
glyph: Another good testing idiom you may want to consider in the future is "[one assert per test](http://www.mfranc.com/unit-testing/good-unit-test-one-assert/)".  When you make this many asserts in a test, in a loop, it can be very difficult to tease out what's going wrong when they start failing.  It's also harder to write the tests first this way, because you have to drop the whole implementation at once rather than incrementally one property at a time.

glyph:
Thanks @BenjamenMeyer - nice to have a more complete swift mock here.

This PR is a bit too big, and I pointed out a few quality issues, but, it seems to hold up to the quality standards we put forward and I really don't want to delay it (especially since its external-facing behavior is correct).  If you have any time for refactoring post-merge that would be great, but I know you've probably been bottlenecked on my code reviews for a while so I am going to try to clear that log-jam before getting picky about quality minutae :).

In general, I think we should be shooting for PRs that are approximately 1/4 this size.  There's pretty good research that shows - for effective code review - human cognition caps out at about 400 lines.  This doesn't mean you need to write 250 lines of code and then just stop, but any time you feel tempted to put a bulleted list into the description of a PR, that's a hint that each thing in the list really ought to be its own PR.  You can feel free to stack them, one on top of the next.  But for example, in this case, I believe that each of `PUT`, `DELETE`, `GET`, and `HEAD` could have been their own PRs; granted with some internal infrastructure to support meta-data stuff, but each one of those verbs splits reasonably cleanly into a separate piece of work, which would be much faster to review.

This does start getting into matters of taste though, and the primary issue we have here is "not enoguh reviewers"; the key stuff, test coverage, clear code and so on, is all met; but I hope we can get more trivial no-brainer PRs that I can merge by just pushing the button without thinking too much in the future :).

BenjamenMeyer:
@glyph thanks. I'll get those addressed and PRs submitted accordingly.

PR #647, opened 2016-07-13
Enhancement internal external urls

BenjamenMeyer:
- Add support for Internal URLs in addition to the Public URLs, by default they are the same; external APIs can make the different if desired.
- Use the Service ID as the unique identifier instead of the Service Name
- Add remaining infrastructure or correct existing infrastructure to support the implementation of the RESTful interfaces for the OSKSCATALOG and related OSKSADM extensions; corrections were due to better understandings of the extensions during RESTful implementation.
- Add RESTful interfaces for the OSKSCATALOG extension, and a couple OSKSADM extension interfaces related to the same support
- OSKSCATALOG/OSKSADM functionality is supported both globally (all users) and for specific tenants


BenjamenMeyer:
Note: 2194 of the new lines are for tests.

BenjamenMeyer:
@glyph sorry for the size - this should make the new functionality usable by all.

glyph:
@BenjamenMeyer - +3k? Jeez. Is there _any_ way to split this up?

BenjamenMeyer:
@glyph if y'all promise to do reviews faster then I'll push it in a series (PR #648 being the first in that series). The series will consist of 4 parts:
- More infrastructure (PR #648)
- OS-KSADM Service Registration
- OS-KSCATALOG Endpoint Template for All
- OS-KSCATALOG Endpoint Template for Tenant

glyph:
@BenjamenMeyer Review speed is largely a function of PR size :).  Small PRs that I can easily absorb in one sitting will definitely get reviewed faster. A 3K line PR is going to need to wait until I have an entire day set aside for Mimic only, which is likely to push it out weeks or more.  I think you noticed a big gap because I was at conferences, then on vacation, and the other primary maintainer is on maternity leave, so there was just nobody left to do anything.  I look at the PR list every day that I'm at work.

BenjamenMeyer:
@glyph understood; as I said - I'm happy to break it down to smaller PRs if it helps to make things go faster.

BenjamenMeyer:
#648 fully ready for review

BenjamenMeyer:
Note: Keeping this hear until it's completed in merges via other PRs.

Added the :umbrella: to denote it's coverage as such.

PR #648, opened 2016-07-19
Enhancement: Infrastructure in support of the External API support

BenjamenMeyer:
Infrastructure changes related to PR #647 
First breakdown PR for that massive PR
- removes all the new RESTful APIs and related tests, keeps all other internal changes and their tests.


.gitignore
```diff
@@ -7,6 +7,7 @@ dropin.cache
 twistd.pid
 docs/_build
 _tmp
-_trial_temp
-.tox
+_trial_temp*
+.tox*
 *.egg-info
+*.sw?
```
glyph: Could these configuration changes be split out as well?  They seem pretty independent.

.coveragerc
```diff
@@ -1,3 +1,6 @@
+[report]
+precision = 2
```
glyph: What is the purpose of this change?

mimic/model/identity_objects.py
```diff
@@ -62,17 +80,164 @@ def not_found(message, request):
...
+
+    def __init__(self, template_dict=None):
+        """
+        Create an :obj:`EndpointTemplateStore`
+
+        :param dict template_dict (optional): optional dictionary to load the
+            Endpoint Template data from. The dictionary is simply passed to the
+            deserializer.
+        """
+        self._template_data = template_dict
```
glyph: This could be made a lot shorter / simpler / safer / consistent by being an `attr.s`-ified class.

mimic/model/identity_objects.py
```diff
@@ -62,17 +80,164 @@ def not_found(message, request):
...
+            return data
+        else:
+            data['tenantId'] = tenant_id
+            for template_key, local_attribute in self.tenant_mapping:
+                value = getattr(self, local_attribute)
+                if value is not None:
+                    data[template_key] = value
+            return data
+
+    def deserialize(self, data):
```
glyph: It seems to me that this ought to be a classmethod that returns a new `EndpointTemplateStore` each time rather than one that has a side-effect of mutating an existing one.  Is there a use-case for multiple calls to `deserialize` on the same object?

mimic/model/identity_objects.py
```diff
@@ -62,17 +80,164 @@ def not_found(message, request):
...
+        :param text_type url: the URL to do the replacement on.
+        :param text_type tenant_id: the tenant-id to insert into the URL.
+        """
+        value_to_replace = self.tenant_alias
+        if value_to_replace is None:
+            value_to_replace = '%tenant_id%'
+
+        return url.replace(value_to_replace, tenant_id)
+
+    def serialize(self, tenant_id=None):
```
glyph: If this were an `attr.s`-ified class, it could use [`asdict`](https://attrs.readthedocs.io/en/stable/api.html#attr.asdict) to do serialization and skip most of this code, I believe.

glyph: Oops.  Maybe not completely.  I guess there's still the need to translate to decidedly Python-unfriendly names like `RAX-AUTH:tenantAlias`.

mimic/model/identity_objects.py
```diff
@@ -62,17 +80,164 @@ def not_found(message, request):
...
     :return: dictionary representing the error body.
     """
-    return _identity_error_message("forbidden", message, FORBIDDEN, request)
+    return _identity_error_message("unauthorized", message, UNAUTHORIZED, request)
+
+
+@implementer(IEndpointTemplate, IPlugin)
+class EndpointTemplateStore(object):
+    """
+    A :obj"`EndpointTemplateStore` is an object whiich provides the
```
glyph: nitpick: "whiich"

mimic/test/test_identity_objects.py
```diff
@@ -173,6 +477,82 @@ def test_update_endpoint_template_invalid(self):
...
+        )
+        eeapi.endpoint_templates[old_eeapi_template_id].id_key = \
+            new_eeapi_template_id
+
+        with self.assertRaises(ValueError):
+            eeapi.update_template(new_eeapi_template)
+
+    def test_has_endpoint_template_invalid(self):
+        """
+        Validate taht :obj:`ExternalApiStore` will return False if it
```
glyph: spelling: "taht"

mimic/model/identity_objects.py
glyph: "provides the functionality to integrate" is pretty maddeningly vague.  I realize that we're a couple of meta-levels out here from the actual functionality but can you say what it does more directly?

mimic/model/identity_objects.py
```diff
@@ -62,17 +80,164 @@ def not_found(message, request):
...
+    functionality to integrate an Endpoint Template instance. It is only
+    meant to work with a single Endpoint Template. Use :obj:`ExternalApiStore`
+    to manage multiple Endpoint Templates.
+
+    :ivar list required_mapping: list of entries in the template that are
```
glyph: Are these intended to be public `ivar`s, or are they really (internal) `cvar`s?

mimic/model/identity_objects.py
```diff
@@ -62,17 +80,164 @@ def not_found(message, request):
...
+    def __init__(self, template_dict=None):
+        """
+        Create an :obj:`EndpointTemplateStore`
+
+        :param dict template_dict (optional): optional dictionary to load the
+            Endpoint Template data from. The dictionary is simply passed to the
+            deserializer.
+        """
+        self._template_data = template_dict
+        self.id_key = None
```
glyph: I am very wary of exposing an API surface with this many `None`s in it.  If these are all values which might or might not be present, can these attributes be private?  If a caller wants to interact with an `EndpointTemplateStore`, they have to be a constant nervous wreck about testing every field for its existence before using it.

Sandi Metz has an article on the subject of nulls (albeit in Ruby, not Python) [here](http://www.sandimetz.com/blog/2014/12/19/suspicions-of-nil) which became a really great talk [here](https://www.youtube.com/watch?v=OMPfEXIlTVE); these echo my general opinion that whenever we see a `None` we should be asking: what does the caller really want out of this object?  Can we implement two versions, the "it's nothing" and "it's something" versions of that behavior, and just have two different classes, but one interface, so the caller gets uniform results?

In this case it's probably not as simple as one or the other because there are multiple optional values, but perhaps we could just hide everything but `serialize`, for example, which is uniformly a `dict{string:string}`?

mimic/test/dummy.py
```diff
@@ -183,6 +189,8 @@ def make_example_external_api(case, name=u"example",
     :param boolean or None set_enabled: If none, the endpoint templates
         are used AS-IS. If a boolean type, then it sets all the templates
         to have the same default accessibility for all tenants.
+    :param text_type service_type: type of the service. If noe, the type
```
glyph: "If noe" meaning "if unspecified"?

mimic/test/dummy.py
```diff
@@ -195,7 +203,17 @@ def make_example_external_api(case, name=u"example",
...
         endpoint_templates = [ExampleEndpointTemplate()]
 
-    service_type = endpoint_templates[0].type_key
+    if service_type is None:
+        # default parameter value, take the template type from the first
+        # endpoint template in the list
+        service_type = endpoint_templates[0].type_key
+
+    elif len(endpoint_templates) == 1:
+        # service_type was specified and most likely endpoint_templates
```
glyph: "most likely" seems a bit weird of an assumption to make, especially if you're going to be mutating the value passed in `endpoint_templates` in this specific case; if that's the intent, why not remember whether `endpoint_templates is None` earlier when you tested for it.  Was this written backwards?

mimic/test/helpers.py
```diff
@@ -244,3 +244,14 @@ def validate_link_json(testCase, json_containing_links):
                                   '"href" is not a string: {0}'.format(link))
         testCase.assertIsInstance(link['rel'], text_type,
                                   '"rel" is not a string: {0}'.format(link))
+
+
+def get_template_id(case, eeapi):
+    """
+    Retrieve the template id.
```
glyph: Needs more explanation - "retrieve the template ID from the given `EndpointApiTemplate`, failing the given test case if it's `None`.

mimic/test/helpers.py
```diff
@@ -244,3 +244,14 @@ def validate_link_json(testCase, json_containing_links):
...
+    """
+    id_key = None
+    for k in eeapi.endpoint_templates.keys():
+        id_key = k
```
glyph: This ordering seems random - is it intended to be?

mimic/test/test_core.py
```diff
@@ -141,6 +142,102 @@ class brokenApiMock(object):
         with self.assertRaises(TypeError):
             MimicCore(Clock(), [brokenApiMock()])
 
+    def test_load_duplicate_api_uuid(self):
```
glyph: Please write a test docstring a-la https://jml.io/pages/test-docstrings.html

mimic/test/test_core.py
```diff
@@ -141,6 +142,102 @@ class brokenApiMock(object):
...
+            eeapi.remove_template(template_id)
+
+        core = MimicCore(Clock(), [eeapi])
+        core.remove_external_api(
+            eeapi.uuid_key
+        )
+
+    def test_remove_api_invalid(self):
+        """
+        Validate API removal fails when it does not find the
```
glyph: again apropos of https://jml.io/pages/test-docstrings.html you can just delete the word "Validate" here :)

mimic/catalog.py
```diff
@@ -33,22 +33,40 @@ def __init__(self, tenant_id, region, endpoint_id, prefix=None,
...
+    def url_with_prefix(self, uri_prefix, internal_url=False):
         """
         Generate a URL to this endpoint, given the URI prefix for the service.
 
+        :param text_type uri_prefix: prefix to start the URL with, e.g the
+            Request's Base URL
+        :param boolean internal_url: whether or not to provide the Internal or
+            Public URL. If True, provide the internal URL.
+
+        Note: internal_url is only honored by External APIs, e.g `external` is
```
glyph: markup nitpick: [there's a `..note:` directive for this sort of thing in Sphinx](http://www.sphinx-doc.org/en/stable/markup/para.html?highlight=note#directive-note)

mimic/core.py
```diff
@@ -102,25 +114,51 @@ def add_api(self, api):
...
+
+        :param text_type api_id: the id of the API instance
+            e.g 3845-39583, cloudfiles-mimic
+        """
+        if api_id in self._uuid_to_api_external:
+            api = self._uuid_to_api_external[api_id]
+
+            if len(api.list_templates()) == 0:
+                del self._uuid_to_api_external[api_id]
+
```
glyph: style nitpick: this seems like a lot of vertical whitespace.  Especially given that these diffs are kinda big to begin with, would you mind compressing a little?  I don't mind a line or two separating different sections in a function, but after each "`else:`" is a bit much.

mimic/model/identity_objects.py
```diff
@@ -62,17 +80,164 @@ def not_found(message, request):
...
+    meant to work with a single Endpoint Template. Use :obj:`ExternalApiStore`
+    to manage multiple Endpoint Templates.
+
+    :ivar list required_mapping: list of entries in the template that are
+        required to be present for a valid template
+    :ivar list optional_mapping: list of entries in the template that are
+        optionally present in a valid template, along with a default value.
+
+
+    Note: The OpenStack documentation[0] does not specify any required
```
glyph: (see above about the note directive)

mimic/model/identity_objects.py
```diff
@@ -62,17 +80,164 @@ def not_found(message, request):
...
+        parameters. For this implementation, the `id`, `region`, `type`,
+        and `name` fields are required. The `name` and `type` fields
+        are used for creating an instance of the :obj:`ExternalApiStore`
+        that will enable the listing to be put into the Service Catalog;
+        the `id` and `region` fields are used by the :obj:`ExternalApiStore`
+        for managing the templates.
+
+    [0] http://developer.openstack.org/api-ref-identity-v2-ext.html
```
glyph: I think you want to use [ReST hyperlink syntax](http://www.sphinx-doc.org/en/stable/rest.html?highlight=url#hyperlinks) here, i.e. `` `The OpenStack documentation <http://developer.openstack.org/api-ref-identity-v2-ext.html>` ``

mimic/model/identity_objects.py
```diff
@@ -62,17 +80,164 @@ def not_found(message, request):
...
+    """
+
+    """
+    The following lists of tuples are for mapping the template defined by
```
glyph: There's nothing wrong with this per se, if it were to be a private attribute (so feel free to ignore this comment) but I vastly prefer a separate, named type that has its own documentation to look at (among other things, sphinx can't "see" this triple-quoted string here, so it's just a comment that will get ignored in generated documentation).

Instead, you can do something like:

```python
@attr.s
class MapInfo(object):
    spec_key = attr.ib()
    attr_name = attr.ib()
    default_value = attr.ib(optional=True)

"..."

required_mapping = [
    MapInfo(*value) for values in [
        ('id', 'id_key'), # et. al.
    ]
]
```

mimic/test/test_identity_objects.py
```diff
@@ -56,6 +297,46 @@ def test_listing_templates(self):
         eeapi.remove_template(eeapi_template_id)
         self.assertEqual(len(eeapi.list_templates()), 0)
 
+    def test_listing_templates_tenant(self):
+        tenant_id = 'some-tenant'
```
glyph: test docstring request for ease of tracking

mimic/test/test_identity_objects.py
```diff
@@ -56,6 +297,46 @@ def test_listing_templates(self):
...
+        )
+
+        listing = [ept for ept in eeapi.list_tenant_templates(tenant_id)]
+        self.assertEqual(len(listing), 1)
+        eeapi.remove_template(eeapi_template_id)
+        new_listing = [ept for ept in eeapi.list_tenant_templates(tenant_id)]
+        self.assertEqual(len(new_listing), 0)
+
+    def test_listing_templates_tenant_with_specific_enabled(self):
+        tenant_id = 'some-tenant'
```
glyph: test docstring request for ease of tracking

mimic/test/test_identity_objects.py
```diff
@@ -173,6 +477,82 @@ def test_update_endpoint_template_invalid(self):
...
+        eeapi = make_example_external_api(
+            self
+        )
+        self.assertFalse(
+            eeapi.has_template('some-template-id')
+        )
+
+    def test_has_endpoint_template(self):
+        """
+        Validate taht :obj:`ExternalApiStore` will return True if it
```
glyph: spelling: "taht"

glyph:
I think this is going to need another pass once some more of the docs have been filled out - hopefully there's enough meat on this review feedback bone to do something with for now ;).

One overall bit of feedback I have - can you give this a clearer title / scope than "break down #1"?  Each change should make some sense on its own even if it is part of a greater whole.

.coveragerc
BenjamenMeyer: Enabling the local code coverage to report the partial percentage so one can see the same as the codecov reports.

.gitignore
BenjamenMeyer: Done, see PR #649

mimic/model/identity_objects.py
BenjamenMeyer: Will do.

mimic/model/identity_objects.py
BenjamenMeyer: probably `cvars`, I also seemed to have missed the `tenant_mapping` one. Correcting.

mimic/test/dummy.py
BenjamenMeyer: fixed

mimic/test/dummy.py
BenjamenMeyer: The `most likely` wording is simply due to the fact that in most cases the default will be used. Still, after reviewing the function I think I've got a slight improvement that can remove the branch that wording falls under without having to add any additional variables to track the state.

The intent was not to remember anything, but to capture default values and ensure the initial template met the conditions of the docstring specification.

mimic/test/helpers.py
BenjamenMeyer: typically only 1 template exists in the `eeapi` where this is called, and it's primarily a helper for removing a template.

mimic/catalog.py
BenjamenMeyer: we're using the sphinx markup in the docstrings?

mimic/core.py
BenjamenMeyer: done

mimic/model/identity_objects.py
BenjamenMeyer: :+1: updating

mimic/test/test_identity_objects.py
BenjamenMeyer: I thought I had caught all of that...will review and update. Would be nice if the `tox -e docs` caught this since it does catch that on everything else.

mimic/model/identity_objects.py
BenjamenMeyer: hmm...having issues with the `optional=True` part. I don't see that kind of functionality documented https://attrs.readthedocs.io/en/stable/api.html#core either.

glyph: sorry, `default=None` should do the trick.

BenjamenMeyer:
@glyph implementation of `ddt` did in fact reduce code in this PR https://github.com/rackerlabs/mimic/pull/648/commits/2db0afc81612b7c02b54273c2498d811f572359f (+72, - 103); issue was that https://github.com/rackerlabs/mimic/pull/648/commits/1347c73fe852cdc10e0776623ef70a9b7528f0b7 added more than what was removed (+148, -69), so it was overall positive (+48). It'll probably have a bigger impact in other PRs.

mimic/model/identity_objects.py
BenjamenMeyer: Considering this; it'll probably be my next PR. I think most other things mentioned thus far have been addressed so I can focus on this one next.

mimic/model/identity_objects.py
BenjamenMeyer: Assuming the class was converted to one based on using `attr`, then the `__init__()` could probably go away and we're just left with the `serialize` and `deserialize`. That could very well work; but unfortunately due to the mimic's naming convention and that of OSKS-CATALOG and it's components - like `RAX-AUTH:tenantAlias` that's probably as close as we can get.

mimic/test/test_core.py
BenjamenMeyer: done

mimic/test/test_core.py
BenjamenMeyer: done

mimic/model/identity_objects.py
BenjamenMeyer: class was converted to using `attrs` in latest commit, and just those three functions defined, no more `__init__`.

BenjamenMeyer:
@glyph poke

mimic/model/identity_objects.py
```diff
@@ -151,11 +329,41 @@ def list_tenant_endpoints(self, tenant_id):
...
                     )
                 )
         return endpoints
 
+    def list_tenant_templates(self, tenant_id):
+        """
+        List the tenant specific endpoints.
+
+        :param text_type tenant_id: tenant id to operate on
+        :returns: an iterable of the endpoints available for the specified
```
glyph: Should this be "an iterable of the endpoint templates available"?

mimic/test/dummy.py
```diff
@@ -96,63 +101,61 @@ def resource(self):
         return example_resource
 
 
-@implementer(IEndpointTemplate)
-class ExampleEndpointTemplate(object):
+def ExampleEndpointTemplate(name=u"example", region="EXTERNAL", version="v1",
```
glyph: Coding-standard; since this is a function and not a type the first letter should be lower-case.  Thanks for exposing it as a function though, it's nice to not have to support inheritance as an API surface.

mimic/rest/identity_api.py
```diff
@@ -382,6 +385,7 @@ def validate_token(self, request, token_id):
                 impersonator_session.username)
 
         if token_id in get_presets["identity"]["token_fail_to_auth"]:
+            # weird mix between an unauthorized (401) and not_found (404)
```
glyph: I don't quite get this comment.

mimic/test/dummy.py
```diff
@@ -96,63 +101,61 @@ def resource(self):
...
-        self.tenant_alias = tenantid_alias
-        self.version_id = version
-        self.version_info = (version_info_url
-                             if version_info_url is not None
-                             else url + '/versionInfo')
-        self.version_list = (version_list_url
-                             if version_list_url is not None
-                             else url + '/versions')
+    epts = EndpointTemplateStore()
+    epts.id_key = uuid
```
glyph: Couldn't these all be supplied as arguments to `EndpointTemplateStore`'s constructor?  As I understand it, this is just here to provide lots of reasonable defaults.  Just style-wise, I strongly prefer expressions to statements, since that's the difference between "correct as created" vs. "numerous potentially broken side-effects".

mimic/test/dummy.py
```diff
@@ -96,63 +101,61 @@ def resource(self):
...
 
 
-@implementer(IEndpointTemplate)
-class ExampleEndpointTemplate(object):
+def ExampleEndpointTemplate(name=u"example", region="EXTERNAL", version="v1",
+                            url="https://api.external.example.com:8080",
+                            public_url=None, internal_url=None, admin_url=None,
+                            version_info_url=None, version_list_url=None,
+                            type_id=u"example", enabled=False,
+                            uuid=text_type(uuid.uuid4()),
```
glyph: This is fixed at import time.  If that's the idea, then having it fixed across all runs would be good; define it as something like `uuid.UUID('4397012d-4aab-4437-9ac5-b123744a3e80')`.  If the idea is that the unspecified default should generate a new, unique UUID each time, be sure to do that in the function.

mimic/rest/identity_api.py
BenjamenMeyer: The response structure being returned is that of a `404`, but the status code is that of a `401`. Why? I don't know; an old OpenStack bug? insight would be appreciated; I'm just highlighting what's is going on. I'll update the comment and if any insights provided I'll add them.

mimic/test/dummy.py
BenjamenMeyer: It should be different for each call.

mimic/model/identity_objects.py
BenjamenMeyer: fixed.

mimic/test/dummy.py
BenjamenMeyer: they probably could now; I originally didn't do that specifically b/c `deserialize` was being called in the `__init__`. Since we already changed how the `EndpointTemplateStore` is structured - using `attrs` and no longer having its own `__init__` - then this should be possible.

mimic/test/dummy.py
BenjamenMeyer: Fixed

mimic/rest/identity_api.py
```diff
@@ -382,6 +385,8 @@ def validate_token(self, request, token_id):
                 impersonator_session.username)
 
         if token_id in get_presets["identity"]["token_fail_to_auth"]:
+            # This is returning a 401 Unauthorized message but in a 404 not_found
+            # JSON data format. Is there a reason for this? An old OpenStack bug?
```
BenjamenMeyer: @glyph this is the update per https://github.com/rackerlabs/mimic/pull/648#discussion_r73444319; if there is any insight into the background I'd be happy to add it.

glyph:
OK!  It's big, but you've integrated tons of feedback here, so I think we can move forward.  Thanks for bearing with me.

PR #649, opened 2016-07-21
Enhancement: Config changes

BenjamenMeyer:
- add Vim swap files to gitignore
- enable local coverage report to generate fractional percentages
  regarding coverage so one can see locally what codecov generates

Per #648


glyph:
Thanks for splitting this out.  Looks straightforward enough.

PR #651, opened 2016-07-29
Bugfix account operations

BenjamenMeyer:
Update SwiftMock for support the HEAD and POST operations at the Account level.
These calls are inseparable without ability to access the underlying model/framework to validate the storage details as one adds data and the other retrieves it.

POST allows the ability to add custom metadata on the account.
HEAD allows retrieval of the custom metadata as well as counts of total containers, total objects (in all containers), and total bytes used (by all objects in all containers).

### DIFF
--- a/mimic/test/test_swift.py
+++ b/mimic/test/test_swift.py
@@ -303,7 +303,8 @@ class SwiftAccountTests(SwiftTestBase):
     @ddt.unpack
     def test_head_account(self, container_count, object_count):
         """
-        Validate that Heading an account reports the correct data.
+        HEADing an account correctly reports the container count,
+        object count, and number of bytes used.
         """
         # Add containers and objects per parameters
         for container_index in range(container_count):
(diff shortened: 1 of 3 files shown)


mimic/test/test_swift.py
```diff
@@ -243,6 +289,98 @@ def test_openstack_ids(self):
...
+                b"X-Account-Meta-MockTest": [b"HelloWorld"],
+                b"x-account-meta-field": [b"WorldWide"]
+            }
+
+        post_response, post_content = self.post_account(headers=headers)
+
+        # validate via a HEAD operation
+        head_response, head_content = self.head_account()
+
+        if headers is not None:
```
glyph: I'll repeat my same uneasiness with control structures in tests from previous reviews, but again, not blocking feedback, just something to think about :)

mimic/test/test_swift.py
```diff
@@ -243,6 +289,98 @@ def test_openstack_ids(self):
...
+                b"X-Account-Bytes-Used")[0],
+            "{0}".format(total_bytes).encode("utf-8"))
+
+    @ddt.data(
+        True,
+        False
+    )
+    def test_post_account_metadata(self, post_metadata):
+        """
+        POSTing an account correctly saves metadata to the account.
```
glyph: "correctly" is another one of those non-words from https://jml.io/pages/test-docstrings.html that we should avoid.

mimic/test/test_swift.py
```diff
@@ -243,6 +289,98 @@ def test_openstack_ids(self):
...
+
+    @ddt.data(
+        (0, 0),
+        (1, 0),
+        (2, 5),
+    )
+    @ddt.unpack
+    def test_head_account(self, container_count, object_count):
+        """
+        HEADing an account correctly reports the container count,
```
glyph: https://jml.io/pages/test-docstrings.html again :)

glyph:
If you could do a little bit of post-hoc cleanup based on the inline feedback that would be great but this looks good to merge to me.

BenjamenMeyer:
@glyph thanks...will address with follow up PR.

PR #653, opened 2016-08-10
update pull request 648

glyph:
Since github doesn't let me update a branch on someone else's remote that I can't merge because it's out of date, I need to create a new PR.


glyph:
The merge button was green so opening this was actually unnecessary. (I guess I must have seen some bad cached state or something.)

PR #654, opened 2016-08-10
fix the pypy builder to be able to build its own Cryptography wheels if necessary

glyph:
This should allow us to catch up with #636 .

### DIFF
--- a/.travis/install.sh
+++ b/.travis/install.sh
@@ -3,12 +3,23 @@
 set -e
 set -x
 
+UPDATED_BREW="false";
+
+function _brew_update () {
+    if "${UPDATED_BREW}"; then
+        return 0;
+    else
+        UPDATED_BREW=true;
+        brew update;
+    fi;
+}
+
 if [[ "${MACAPP_ENV}" == "system" ]]; then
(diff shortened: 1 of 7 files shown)


.travis/install.sh
```diff
@@ -19,19 +30,32 @@ else
     DARWIN="false";
 fi
 
+
 # Do we need to use a Python from pyenv, or will the system one suffice?
 PYENV_PYTHON="";
 
 if [[ "${TOXENV}" == "pypy" ]]; then
     PYENV_PYTHON="pypy-4.0.1";
```
derwolfe: (optional/discussion) Do you think we should bump the version of `pypy` as well?

derwolfe:
Assuming CI passes and modulo https://github.com/rackerlabs/mimic/pull/654/files#r74337131 I think this is good to land. :+1: :)

.travis/install.sh
glyph: Yes; in a separate PR.  My first attempt to fix this was to start by just bumping pypy to get a version that wheels were built against, but that was a mess.

PR #655, opened 2016-08-11
Enhancement default region

BenjamenMeyer:
The Service Catalog provided by Identity/Keystone provides the ability to specify a default region for a given user.


glyph:
Looks good. Can you file a follow-up for actually making this per-user settable, as per Keystone?  This confused me a bit because the description makes it sound like it's going to track the state, but really it just returns the already-hard-coded default region.

BenjamenMeyer:
Sorry about the confusion; I'll see what I can do.

PR #656, opened 2016-08-12

ryandub:


glyph:
Looks like this just hit a PyPI outage; re-ran the build.

glyph:
Might as well update the branch while we're at it though.

PR #665, opened 2016-08-23
Enhancement restore external api restful interface

BenjamenMeyer:
This restore the RESTful API that was originally part of #647, and fixed per changes in related PRs that resulted in #647 getting merged.

This PR consists of two things:
1) The RESTful interface
2) The related tests

Tests are broken into separate files for each division of the interface (f.e OSKS-CATALOG vs OSKS-ADM). All tests pass.

Note: As this is only 3 commits (no I didn't squash anything), I hate the size but I have no good way of breaking this down in a good time constraint. I don't want to repeat the process of #647 and #648 that lead to this one - which will just be a messy merge/delete/commit/merge/delete commit.

### DIFF
--- a/mimic/core.py
+++ b/mimic/core.py
@@ -24,6 +24,66 @@ from mimic.model.glance_objects import GlanceAdminImageStore
 from mimic.model.valkyrie_objects import ValkyrieStore
 
 
+class MimicCoreExceptions(Exception):
+    """
+    Parent for all Exceptions related to MimicCore
+    """
+    pass
+
+
+class ServiceBadInterface(MimicCoreExceptions):
+    """
+    Service does not implement the required interface.
+    """
+    pass
+
+
+class ServiceExistenceErrors(MimicCoreExceptions):
+    """
+    Exceptions related to Service Existence
+    """
+
+
+class ServiceDoesNotExist(ServiceExistenceErrors):
+    """
+    API does not exist
+    """
+    pass
+
+
+class ServiceExists(ServiceExistenceErrors):
+    """
+    API Already Exists
+    """
+
+
+class ServiceIdExists(ServiceExists):
+    """
+    API with the same ID already exists.
+    """
+    pass
+
+
+class ServiceNameExists(ServiceExists):
+    """
+    API with the same name already exists.
+    """
+    pass
+
+
+class ServiceStateError(MimicCoreExceptions):
+    """
+    Exceptions related to Service States
+    """
+    pass
+
+
+class ServiceHasTemplates(ServiceStateError):
+    """
+    Service still has assigned templates.
+    """
+
+
 class MimicCore(object):
     """
     A MimicCore contains a mapping from URI prefixes to particular service
@@ -90,13 +150,13 @@ class MimicCore(object):
             this_api_id = api.uuid_key
 
             if this_api_id in self._uuid_to_api_external:
-                raise ValueError(
+                raise ServiceIdExists(
                     'An Existing API already exists with the given UUID'
                 )
 
             for existing_api in self._uuid_to_api_external.values():
                 if existing_api.name_key == api.name_key:
-                    raise ValueError(
+                    raise ServiceNameExists(
                         'An Existing API with UUID ' + existing_api.uuid_key +
                         ' is already using that name'
                     )
@@ -108,7 +168,7 @@ class MimicCore(object):
                            random_hex_generator(3))
             self._uuid_to_api_internal[this_api_id] = api
         else:
-            raise TypeError(
+            raise ServiceBadInterface(
                 api.__class__.__module__ + '/' +
                 api.__class__.__name__ +
                 " does not implement IAPIMock or IExternalAPIMock"
@@ -127,9 +187,9 @@ class MimicCore(object):
             if len(api.list_templates()) == 0:
                 del self._uuid_to_api_external[api_id]
             else:
-                raise ValueError("API still has endpoint templates")
+                raise ServiceHasTemplates("API still has endpoint templates")
         else:
-            raise IndexError(api_id + " is not a valid external API")
+            raise ServiceDoesNotExist(api_id + " is not a valid external API")
 
     def get_external_apis(self):
         """
@@ -157,7 +217,7 @@ class MimicCore(object):
         if api_id in self._uuid_to_api_external:
             return self._uuid_to_api_external[api_id]
         else:
-            raise IndexError(
+            raise ServiceDoesNotExist(
                 "Unable to locate an API  the id" + str(api_id)
             )
 
--- a/mimic/core.py
+++ b/mimic/core.py
@@ -24,34 +24,31 @@ from mimic.model.glance_objects import GlanceAdminImageStore
 from mimic.model.valkyrie_objects import ValkyrieStore
 
 
-class MimicCoreExceptions(Exception):
+class MimicCoreException(Exception):
     """
     Parent for all Exceptions related to MimicCore
     """
-    pass
 
 
-class ServiceBadInterface(MimicCoreExceptions):
+class ServiceBadInterface(MimicCoreException):
     """
     Service does not implement the required interface.
     """
-    pass
 
 
-class ServiceExistenceErrors(MimicCoreExceptions):
+class ServiceExistenceError(MimicCoreException):
     """
     Exceptions related to Service Existence
     """
 
 
-class ServiceDoesNotExist(ServiceExistenceErrors):
+class ServiceDoesNotExist(ServiceExistenceError):
     """
     API does not exist
     """
-    pass
 
 
-class ServiceExists(ServiceExistenceErrors):
+class ServiceExists(ServiceExistenceError):
     """
     API Already Exists
     """
@@ -61,21 +58,18 @@ class ServiceIdExists(ServiceExists):
     """
     API with the same ID already exists.
     """
-    pass
 
 
 class ServiceNameExists(ServiceExists):
     """
     API with the same name already exists.
     """
-    pass
 
 
-class ServiceStateError(MimicCoreExceptions):
+class ServiceStateError(MimicCoreException):
     """
     Exceptions related to Service States
     """
-    pass
 
 
 class ServiceHasTemplates(ServiceStateError):
--- a/mimic/model/identity_errors.py
+++ b/mimic/model/identity_errors.py
@@ -7,35 +7,30 @@ class EndpointTemplateException(Exception):
     """
     Parent for all Identity Endpoint Template Exceptions
     """
-    pass
 
 
 class InvalidEndpointTemplateException(EndpointTemplateException):
     """
     Parent for all Endpoint Template Validation Exceptions.
     """
-    pass
 
 
 class InvalidEndpointTemplateInterface(InvalidEndpointTemplateException):
     """
     :obj: does not implement the required interface, :obj:IEndpointTemplate.
     """
-    pass
 
 
 class InvalidEndpointTemplateMissingKey(InvalidEndpointTemplateException):
     """
     :obj: is missing a required field.
     """
-    pass
 
 
 class InvalidEndpointTemplateId(InvalidEndpointTemplateException):
     """
     :obj: has an invalid endpoint template id value.
     """
-    pass
 
 
 class InvalidEndpointTemplateServiceType(InvalidEndpointTemplateException):
@@ -43,32 +38,27 @@ class InvalidEndpointTemplateServiceType(InvalidEndpointTemplateException):
     :obj: has an invalid service type or the service type does not match
     the API it is being submitted for.
     """
-    pass
 
 
 class EndpointTemplateDisabledForTenant(EndpointTemplateException):
     """
     Specified endpoint template is disabled for the tenant.
     """
-    pass
 
 
 class EndpointTemplateExistenceException(EndpointTemplateException):
     """
     Parent of template existence exceptions.
     """
-    pass
 
 
 class EndpointTemplateAlreadyExists(EndpointTemplateExistenceException):
     """
     Endpoint Template already exists.
     """
-    pass
 
 
 class EndpointTemplateDoesNotExist(EndpointTemplateExistenceException):
     """
     Endpoint Template does not exist.
     """
-    pass
--- a/mimic/model/identity_errors.py
+++ b/mimic/model/identity_errors.py
@@ -3,14 +3,14 @@ Errors for the Identity Models
 """
 
 
-class EndpointTemplateExceptions(Exception):
+class EndpointTemplateException(Exception):
     """
     Parent for all Identity Endpoint Template Exceptions
     """
     pass
 
 
-class InvalidEndpointTemplateException(EndpointTemplateExceptions):
+class InvalidEndpointTemplateException(EndpointTemplateException):
     """
     Parent for all Endpoint Template Validation Exceptions.
     """
@@ -46,28 +46,28 @@ class InvalidEndpointTemplateServiceType(InvalidEndpointTemplateException):
     pass
 
 
-class EndpointTemplateDisabledForTenant(EndpointTemplateExceptions):
+class EndpointTemplateDisabledForTenant(EndpointTemplateException):
     """
     Specified endpoint template is disabled for the tenant.
     """
     pass
 
 
-class EndpointTemplateExistenceExceptions(EndpointTemplateExceptions):
+class EndpointTemplateExistenceException(EndpointTemplateException):
     """
     Parent of template existence exceptions.
     """
     pass
 
 
-class EndpointTemplateAlreadyExists(EndpointTemplateExistenceExceptions):
+class EndpointTemplateAlreadyExists(EndpointTemplateExistenceException):
     """
     Endpoint Template already exists.
     """
     pass
 
 
-class EndpointTemplateDoesNotExist(EndpointTemplateExistenceExceptions):
+class EndpointTemplateDoesNotExist(EndpointTemplateExistenceException):
     """
     Endpoint Template does not exist.
     """
--- a/mimic/model/identity_errors.py
+++ b/mimic/model/identity_errors.py
@@ -1,0 +1,74 @@
+"""
+Errors for the Identity Models
+"""
+
+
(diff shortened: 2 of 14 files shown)


BenjamenMeyer:
Cleaning up and removing tests that are nearly identical and DDT can provide the same coverage.

glyph:
@BenjamenMeyer - can you address the linting failure and merge master? Thanks!

BenjamenMeyer:
@glyph that's in progress. I've already got the linting fixed locally; getting a couple other things cleaned up then I'll merge master and push another commit set up. When done I'll remove the `work in progress` label and let you know.

glyph:
Sounds good!

BenjamenMeyer:
@glyph everything is ready to go - ready for review.

mimic/rest/identity_api.py
```diff
@@ -669,6 +675,518 @@ def list_tenants(self, request):
...
+            return json.dumps(
+                not_found(
+                    "Service API for endoint template not found",
+                    request
+                )
+            )
+
+        try:
+            service.update_template(endpoint_template_instance)
+        except ValueError:
```
glyph: I think I should have caught this in an earlier review, but `ValueError` and `IndexError` are bad exceptions to use here.  You can get an `IndexError` or a `ValueError` from basic type issues with built-in Python data structures, and so it's possible to mask basic implementation bugs here and make them very hard to diagnose.  Please fix `update_template` to raise a more specific `EndpointAlreadyExists`, `ServiceTypeMismatch` and `ServiceIdMismatch`.

mimic/rest/identity_api.py
```diff
@@ -669,6 +675,518 @@ def list_tenants(self, request):
...
+            return json.dumps(
+                conflict(
+                    "Endpoint already exists and service id or service type "
+                    "does not match.",
+                    request
+                )
+            )
+        except IndexError:
```
glyph: Similarly, this should be `TemplateDoesntExist`

mimic/rest/identity_api.py
```diff
@@ -669,6 +675,518 @@ def list_tenants(self, request):
...
+            for api_id in self.core.get_external_apis():
+                api = self.core.get_external_api(api_id)
+                if api.has_template(template_id):
+                    service_id = api.uuid_key
+        else:
+            service_id = service_id.decode('utf-8')
+
+        try:
+            service = self.core.get_external_api(service_id)
+        except IndexError:
```
glyph: See point about generic exception types.

mimic/test/test_identity_osksadm.py
```diff
@@ -0,0 +1,366 @@
...
+        0, 1, 10
+    )
+    def test_listing(self, api_entry_count):
+        """
+        GET will list the registered services.
+        """
+        api_list = []
+
+        # create the desired number of services per test parameter
+        for _ in range(api_entry_count):
```
glyph: I tend to prefer "ignored" for ignored loop variables; `_` is used as a special variable by a few other systems like `gettext` and the interactive interpreter.  No serious problem, but as a habit, it helps avoid potential clashes.

glyph: Loops in tests are a code smell, and comprehensions can be nicer (when they work) for building up lists than manually calling `.append`.  Consider `api_list = [ExternalApiStore(...) for ignored in range(api_entry_count)]` instead.

mimic/test/test_identity_osksadm.py
```diff
@@ -0,0 +1,366 @@
...
+            api_list.append(
+                ExternalApiStore(
+                    text_type(uuid.uuid4()),
```
glyph: `uuid.uuid4` actually does I/O (gathering entropy) so the values under test here won't be stable.  You might want to _not_ ignore the loop variable and use a `uuid5` instead, which will allow you to write this without copy/pasting big obnoxious lists but will still give you stable output.

mimic/test/test_identity_osksadm.py
```diff
@@ -0,0 +1,366 @@
...
+        (response, json_body) = self.successResultOf(
+            json_request(self, self.root, self.verb,
+                         self.uri,
+                         headers=self.headers))
+
+        def validate_api(api_id, api_type, api_name):
+            """
+            Lookup the API in the test's set of APIs  and match the values
+            """
+            for api in api_list:
```
glyph: Another way to write this, again, avoiding loops in tests:

```python
matching_apis = [api for api in api_list if api.uuid_key == api_id]
self.assertEqual(len(matching_apis), 1)
[matching_api] = matching_apis
# ... assert assert assert ...
```

mimic/test/test_identity_osksadm.py
```diff
@@ -0,0 +1,366 @@
...
+            self.assertFalse(
+                True,
+                "Unknown service: {0} - {1} - {2}".format(
+                    api_id, api_type, api_name))
+
+        self.assertEqual(response.code, 200)
+        self.assertEqual(len(json_body["OS-KSADM:services"]), len(api_list))
+        # ensure all services in the response match one in the generated
+        # initially generated set
+        for entry in json_body["OS-KSADM:services"]:
```
glyph: Is it OK if this list is empty?

mimic/test/test_identity_osksadm.py
```diff
@@ -0,0 +1,366 @@
...
+                         self.uri))
+
+        self.assertEqual(response.code, 401)
+        self.assertEqual(json_body['unauthorized']['code'], 401)
+
+    def test_invalid_service(self):
+        """
+        DELETE an unknown service will generate a 404.
+        """
+        data = {
```
glyph: There's a lot of code duplication between these tests. Could this be factored out into a common assertion?  It seems like you'd save a lot of lines that way.

mimic/test/test_identity_oskscatalog.py
```diff
@@ -0,0 +1,688 @@
...
+        self.assertEqual(
+            len(json_body['OS-KSCATALOG:endpointsTemplates_links']),
+            0)
+
+    def test_multiple_external_apis(self):
+        """
+        GET can retrieve numerous External APIs that have External API Templates.
+        """
+        api_list = [self.eeapi]
+        for _ in range(10):
```
glyph: Another place where a list comprehension would be a better fit, I think.  `api_list = [self.eeapi] + [ ... ]`

glyph:
My main complaint is still that this is a lot of code to review at once - but rather than splitting it up, I think just attacking the code duplication in the tests could get it down to a much more manageable size, and make the tests more maintainable in the future, to boot.  Can you give that a shot?  I don't see any blockers here though, aside from raw change size (I'm really trying to get individual merges down to <400 lines); I think if I can review this at a smaller size I'll be happy to push the green button.

mimic/test/test_identity_osksadm.py
BenjamenMeyer: Yes, it means no APIs have been registered. I believe that is part of the test via `ddt`; though can't remember if it's in this specific test or not.

BenjamenMeyer:
@glyph understand the size issue; I do try to remove duplication where I can - and `ddt` helps a lot with that. I'll see what I can do once I address the issues you list with the various `Exceptions`.  The tests do make up a significant portion of the code additions (typically 75% from what I noticed in previous PRs).

BenjamenMeyer:
I think I fixed the issues related to the Exceptions - I changed the ones in `core.py` and `model/identity_objects.py`. Still have other comments to address but wanted to get this up here.

mimic/test/test_identity_osksadm.py
BenjamenMeyer: Actually, we really don't care about the stability of UUID4. It's perfectly okay for it to generate different values from one run to the next.

glyph: If you don't want to change it in this case, that's fine; this is more general test advice.

BenjamenMeyer:
@glyph I've gone through and cleaned up some of the duplication using `ddt` and mix-ins; net of -207 lines in the last 3 commits.

mimic/core.py
```diff
@@ -24,6 +24,66 @@
 from mimic.model.valkyrie_objects import ValkyrieStore
 
 
+class MimicCoreExceptions(Exception):
+    """
+    Parent for all Exceptions related to MimicCore
+    """
+    pass
```
glyph: The "pass" is unnecessary, since the docstring fills the syntactic niche that would require it to be here.  Can you remove these?

mimic/model/identity_errors.py
```diff
@@ -0,0 +1,74 @@
+"""
+Errors for the Identity Models
+"""
+
+
+class EndpointTemplateExceptions(Exception):
```
glyph: Style nitpick; class names should be singular; except in rare cases where they represent a collection, and this one doesn't.

mimic/rest/identity_api.py
```diff
@@ -669,6 +688,521 @@ def list_tenants(self, request):
...
+
+    @app.route('/v2.0/OS-KSCATALOG/endpointTemplates', methods=['POST'])
+    def add_endpoint_templates(self, request):
+        """
+        Add an API endpoint template to the system. By default the API
+        described by the template will disabled for all users.
+
+        Reference: http://developer.openstack.org/api-ref-identity-v2-ext.html
+
+        Note: Either the service-id must be specified in the header or
```
glyph: Not worth changing it in this branch, but the correct way to format this with a ReST docstring is `.. note::`, as documented [here](http://sphinx.readthedocs.io/en/stable/markup/para.html?highlight=note#directive-note).

BenjamenMeyer:
@glyph I had to update against master after merging #666 so I went ahead and address the few comments I saw as well (between `pass`, plurality naming, and RST format).

mimic/model/identity_errors.py
```diff
@@ -0,0 +1,74 @@
+"""
+Errors for the Identity Models
+"""
+
+
+class EndpointTemplateException(Exception):
+    """
+    Parent for all Identity Endpoint Template Exceptions
+    """
+    pass
```
glyph: Can you remove all these `pass` statements?

mimic/rest/identity_api.py
```diff
@@ -669,6 +688,521 @@ def list_tenants(self, request):
                 'message': ("No valid token provided. Please use the 'X-Auth-Token'"
                             " header with a valid token.")}})
 
+    @app.route('/v2.0/services', methods=['GET'])
+    def list_external_api_services(self, request):
+        """
+        List the available external services that endpoint templates
+        may be added to.
+
+        .. note:: Part of the OS-KSADM extension.
```
glyph: When adding a note like this, it's helpful to include a hyperlink that provides a direct reference so that future maintainers don't need to google around for a term.

mimic/rest/identity_api.py
```diff
@@ -669,6 +688,521 @@ def list_tenants(self, request):
...
+        .. note:: Does not implement the limits or markers.
```
glyph: Any time you have a reference in a docstring or a comment to desirable-but-unimplemented functionality, or a known bug, it should include a hyperlink referencing the place that the future work is tracked.

mimic/rest/identity_api.py
```diff
@@ -669,6 +688,521 @@ def list_tenants(self, request):
...
+        """
+        x_auth_token = request.getHeader(b"x-auth-token")
```
glyph: This `x_auth_token` check ought to be a function, or better yet, a decorator that declares authorization is required.  Although mimic is not exactly security sensitive currently, duplicating this kind of code is a very risky anti-pattern; since mimic *may* eventually be handling administrative credentials (and this change only makes it more likely that it will be) we should treat it as though it were.  Please de-duplicate this with the other auth-token checks so that if it does more than check for the existence of the header in the future we don't have to re-implement it.

mimic/rest/identity_api.py
```diff
@@ -669,6 +688,521 @@ def list_tenants(self, request):
...
+        Create a new external api service that endpoint templates
+        may be added to.
+
+        .. note:: Part of the OS-KSADM extensions.
+        .. note:: Only requires 'name' and 'type' fields in the JSON. If the 'id'
+            or 'description' fields are present, then they will be used;
+            otherwise a UUID4 will be assigned to the 'id' field and the
+            'description' will be given a generic value.
+        """
+        x_auth_token = request.getHeader(b"x-auth-token")
```
glyph: Reference earlier comment `x-auth-token` duplication.

mimic/rest/identity_api.py
```diff
@@ -669,6 +688,521 @@ def list_tenants(self, request):
...
+
+    @app.route('/v2.0/services/<string:service_id>', methods=['DELETE'])
+    def delete_external_api_service(self, request, service_id):
+        """
+        Delete/Remove an existing  external service api. It must not have
+        any endpoint templates assigned to it for success.
+
+        .. note:: Part of the OS-KSADM extension.
+        """
+        x_auth_token = request.getHeader(b"x-auth-token")
```
glyph: Reference earlier comment `x-auth-token` duplication.

mimic/rest/identity_api.py
```diff
@@ -669,6 +688,521 @@ def list_tenants(self, request):
...
+    @app.route('/v2.0/OS-KSCATALOG/endpointTemplates', methods=['GET'])
+    def list_endpoint_templates(self, request):
+        """
+        List the available endpoint templates.
+
+        Reference: http://developer.openstack.org/api-ref-identity-v2-ext.html
+
+        .. note:: Marker/Limit capability not implemented here.
+        """
+        x_auth_token = request.getHeader(b"x-auth-token")
```
glyph: Reference earlier comment `x-auth-token` duplication.

mimic/rest/identity_api.py
```diff
@@ -669,6 +688,521 @@ def list_tenants(self, request):
...
+        .. note:: Either the service-id must be specified in the header or
+            a Service Name by the same name must already exist. Otherwise
+            a Not Found (404) will be returned.
+
+        .. note:: A template has certain required parametes. For Mimic the
+            id, name, type, and region parameters are required. See
+            EndpointTemplateStore.required_mapping for details. Other
+            implementations may have different requirements.
+        """
+        x_auth_token = request.getHeader(b"x-auth-token")
```
glyph: Reference earlier comment `x-auth-token` duplication.

mimic/rest/identity_api.py
```diff
@@ -669,6 +688,521 @@ def list_tenants(self, request):
...
+
+        Reference: http://developer.openstack.org/api-ref-identity-v2-ext.html
+
+        .. note:: A template by the same id must already exist in the system.
+
+        .. note:: Either the service-id must be specified in the header or
+            a Service Name by the same name must already exist. Otherwise
+            a Not Found (404) will be returned.
+        """
+        x_auth_token = request.getHeader(b"x-auth-token")
```
glyph: Reference earlier comment `x-auth-token` duplication.

mimic/rest/identity_api.py
```diff
@@ -669,6 +688,521 @@ def list_tenants(self, request):
...
+               methods=['GET'])
+    def list_endpoints_for_tenant(self, request, tenant_id):
+        """
+        List the available endpoints for a given tenant-id.
+
+        Reference: http://developer.openstack.org/api-ref-identity-v2-ext.html
+
+        .. note:: Marker/Limit capability not implemented here.
+        """
+        x_auth_token = request.getHeader(b"x-auth-token")
```
glyph: Reference earlier comment `x-auth-token` duplication.

mimic/rest/identity_api.py
```diff
@@ -669,6 +688,521 @@ def list_tenants(self, request):
...
+
+    @app.route('/v2.0/tenants/<string:tenant_id>/OS-KSCATALOG/endpoints',
+               methods=['POST'])
+    def create_endpoint_for_tenant(self, request, tenant_id):
+        """
+        Enable a given endpoint template for a given tenantid.
+
+        Reference: http://developer.openstack.org/api-ref-identity-v2-ext.html
+        """
+        x_auth_token = request.getHeader(b"x-auth-token")
```
glyph: Reference earlier comment `x-auth-token` duplication.

mimic/rest/identity_api.py
```diff
@@ -669,6 +688,521 @@ def list_tenants(self, request):
...
+               '<string:template_id>', methods=['DELETE'])
+    def remove_endpoint_for_tenant(self, request, tenant_id, template_id):
+        """
+        Disable a given endpoint template for a given tenantid if it's been
+        enabled. This does not affect an endpoint template that has been
+        globally enabled.
+
+        Reference: http://developer.openstack.org/api-ref-identity-v2-ext.html
+        """
+        x_auth_token = request.getHeader(b"x-auth-token")
```
glyph: Reference earlier comment `x-auth-token` duplication.

glyph:
I made a separate comment for each duplicated x-auth-token check which will hopefully make it easy to track the review response.

Beyond that duplication I think we are OK to land, so if you can address that, please feel free to push the green button when you're done.  However, in the future, let's please try to keep the changes smaller ;).  (For example, the exception-translation stuff in this branch could have been split out; but, at this point that would just be a finger exercise, since I've gone through and done the whole 2k-line review at this point...)

mimic/model/identity_errors.py
BenjamenMeyer: Fixed for next push.

mimic/rest/identity_api.py
BenjamenMeyer: I've added URLs where I can still grab them...but it seems that OS-KSCATALOG is no longer being listed by OpenStack's documentation :frowning: so I used the URL I already had in there. It's the best I can do but also a reflection of the sad state of affairs regarding URLs and how easily they become non-existent.

glyph: Yeah :-/.  Links going dead is just an unfortunate fact of life with this stuff.

BenjamenMeyer:
@glyph please check the `rest/decorators.py` and `test/rest_decorators.py`. The rest should just be converting over to that. See https://github.com/rackerlabs/mimic/pull/665/commits/872dc382de8fcb9ae53ccfd3d5b1a70171e9f516 to keep it simple.

PR #666, opened 2016-08-25
Enhancement: missing identity token generation test

BenjamenMeyer:
- seems there is a branch that was not being tested or covered;
  this adds a specific test for the branching mechanism to cover
  both branch possibilities.

### DIFF
--- a/mimic/test/test_identity_model.py
+++ b/mimic/test/test_identity_model.py
@@ -1,0 +1,37 @@
+from __future__ import absolute_import, division, unicode_literals
+
+import ddt
+from twisted.trial.unittest import SynchronousTestCase
+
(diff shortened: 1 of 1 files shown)


glyph:
This looks great. But unfortunately I merged something else first, so I can't land this.  I'm starting to wonder if we should switch to a branching, rather than forking model, so that reviewers can hit the "update branch" button :-\

glyph:
Can you merge trunk and then land?

BenjamenMeyer:
@glyph probably a mix...things like my External API stuff probably would be good (better?) on a branch; but little stuff like this is fine in a fork.

BenjamenMeyer:
tests passed merged as requested

PR #669, opened 2016-08-30
Add Cinder catalog entries that match devstack

wallrj:
I've got some code which needs to work with Rackspace Cinder V1 and with OpenStack Cinder v1 and it attempts to detect whether Cinder v2 is available by performing a volume list using the v2 API and then the v1 API if the first results in an EndpointNotFound error.

In the absence of a permanent devstack testing environment, I'm testing this with Mimic, but mimic's cinder API doesn't use the same service catalog as devstack.

It *only* provides the default "cinder" endpoint which provides a Cinder v2 API.

Here I've added an additional v2 endpoint which seems to be in accordance with this specification:

* https://specs.openstack.org/openstack/openstack-specs/specs/service-catalog.html

For reference, the devstack mitaka service list looks like this:

```
stack@server-03:~$ keystone service-list
...
+----------------------------------+-------------+----------------+-----------------------------------+
|                id                |     name    |      type      |            description            |
+----------------------------------+-------------+----------------+-----------------------------------+
| 7e8eafced3ed4cb8a28a9e48ecf93ef2 |    cinder   |     volume     |       Cinder Volume Service       |
| 26461032e2584578a82a954a25626a37 |   cinderv2  |    volumev2    |      Cinder Volume Service V2     |
| 54bd719bbf1d406b904585c75827e6d7 |    glance   |     image      |        Glance Image Service       |
| eeb8b5d0ee044d74b3d0dcd7f7af9cd0 |   keystone  |    identity    |                                   |
| 05bdfbaf73054c4aa1c16e565ad09e60 |   neutron   |    network     |          Neutron Service          |
| 5772bcce97d0422d9506a4d935ca6597 |     nova    |    compute     |        Nova Compute Service       |
| df370717d4a644ac8d03b9a33259af13 | nova_legacy | compute_legacy | Nova Compute Service (Legacy 2.0) |
+----------------------------------+-------------+----------------+-----------------------------------+
```

### DIFF
--- a/mimic/rest/cinder_api.py
+++ b/mimic/rest/cinder_api.py
@@ -31,7 +31,14 @@ class CinderApi(object):
         """
         return [
             Entry(
-                tenant_id, "volume", "cloudBlockStorage",
+                tenant_id, "volume", "cinder",
+                [
+                    Endpoint(tenant_id, region, text_type(uuid4()), prefix="v2")
+                    for region in self._regions
+                ]
+            ),
+            Entry(
+                tenant_id, "volumev2", "cinderv2",
                 [
                     Endpoint(tenant_id, region, text_type(uuid4()), prefix="v2")
                     for region in self._regions
--- a/mimic/test/test_core.py
+++ b/mimic/test/test_core.py
@@ -63,7 +63,7 @@ class CoreBuildingPluginsTests(SynchronousTestCase):
             dns_plugin.dns,
             cinder_plugin.cinder
         ))
-        # all plugsin should be on the internal listing
+        # all plugins should be on the internal listing
         self.assertEqual(
             plugin_apis,
             set(core._uuid_to_api_internal.values()))
@@ -71,10 +71,6 @@ class CoreBuildingPluginsTests(SynchronousTestCase):
         self.assertEqual(
             set([]),
             set(core._uuid_to_api_external.values()))
-        self.assertEqual(
-            len(plugin_apis),
-            len(list(core.entries_for_tenant('any_tenant', {},
-                                             'http://mimic'))))
 
     def test_load_domain_plugin_includes_all_domain_plugins(self):
         """


mimic/test/test_core.py
```diff
@@ -63,18 +63,14 @@ def test_from_plugin_includes_all_plugins(self):
...
             plugin_apis,
             set(core._uuid_to_api_internal.values()))
         # the external listing should still be empty
         self.assertEqual(
             set([]),
             set(core._uuid_to_api_external.values()))
-        self.assertEqual(
-            len(plugin_apis),
-            len(list(core.entries_for_tenant('any_tenant', {},
-                                             'http://mimic'))))
```
wallrj: I removed this assertion because it failed when the Cinder plugin registered multiple endpoints and because `entries_for_tenant` seems to be well tested elsewhere in the test_core module.

mimic/rest/cinder_api.py
```diff
@@ -31,7 +31,14 @@ def catalog_entries(self, tenant_id):
         """
         return [
             Entry(
-                tenant_id, "volume", "cloudBlockStorage",
```
wallrj: Devstack Mitaka uses the name `cinder`
 * https://git.openstack.org/cgit/openstack-dev/devstack/tree/lib/cinder?h=stable%2Fmitaka#n364

mimic/test/test_core.py
BenjamenMeyer: Q: what was the error?

I'm curious as I put together a lot of the externally hosted API stuff, and am wondering what might be missing for similar support on the internally hosted APIs.

It is well tested, but I'm not familiar enough with the plugin functionality to say that that is equally well tested as the core IAPIMock and IExternalAPIMock functionality.

mimic/rest/cinder_api.py
BenjamenMeyer: I'm guessing we probably need both.

mimic/test/test_core.py
wallrj: The error was `15 != 16` (or something like) because the Cinder mimic plugin now has two catalog entries while that section of the test assumes that each plugin only has only one catalog entry.

Judging by the test docstring I'd say that this part of the test is probably unrelated.

```
         Using the :func:`MimicRoot.fromPlugin` creator for a
        :class:`MimicCore`, the nova and loadbalancer plugins are included.
```

mimic/rest/cinder_api.py
wallrj: Looks like Rackspace uses the name `cloudBlockStorage`

```
$ curl -H "Content-Type: application/json" -s -XPOST -d '{"auth":{"RAX-KSKEY:apiKeyCredentials":{"username":"joe.blogs.example.com","apiKey":"111222333"}}}' https://identity.api.rackspacecloud.com/v2.0/tokens | python -m json.tool
...
{
    "access": {
        "serviceCatalog": [
            ...
            {
                "endpoints": [
                    ...
                    {
                        "publicURL": "https://iad.blockstorage.api.rackspacecloud.com/v1/123456",
                        "region": "IAD",
                        "tenantId": "123456"
                    }
                ],
                "name": "cloudBlockStorage",
                "type": "volume"
            },
        ],
        ...
    }
}
```

I'll change this back for now because I don't think the name makes any difference in my use case but perhaps this needs to be configurable or perhaps Mimic could grow a ``--rackspace / --devstack`` compatibility option.

wallrj: Although, mimic's cinder fakery is already out of line with Rackspace since it only provides Cinder v2  API.

BenjamenMeyer: I don't think there's any reason we couldn't have it registered as both; then we can cover both use-cases.

As to API completeness...well, that's a whole other matter. It basically comes down to what people need and are willing to contribute. This project, AFAICT, is entirely volunteer based.

glyph: @wallrj @BenjamenMeyer - We should definitely register it as both.

However, I think @wallrj points out an important problem here: the endpoint type is "volume" but the presented API is cinder v2; doesn't that mean it should be "volumev2"?  Honestly I'm not sure what the _right_ way is for OpenStack to tell the user what API to expect; I thought it was "type" but the documentation does not appear to be explicit and the examples seem a little inconsistent on that point.

Regarding API completeness - I think calling it "volunteer based" might be slightly misleading, most Rackspace folks who work on it do so as part of their jobs - but it's not a supported product, it's just something we work on as time allows, in support of other projects :-).  Which is exactly why it's so important that we get outside help to make it more complete - so thanks for that!

BenjamenMeyer: True...

From working on Cloud Backup and now working through a migration of API versions, I know there's a field in the Service Catalog entry that provides for versioning capabilities - in fact, I use it with the Endpoint Templating functionality I put into Mimic. I don't know that `IAPIMock` supports it yet though.

glyph:
Hey @wallrj - it's out of date again.  Do you have https://mobile.twitter.com/glyph/status/773663288152367107 set?

wallrj:
Thanks for the tip @glyph I merged forward and ticked the box.

glyph:
@wallrj OK, I'm starting to think this Github feature doesn't work: 
screen shot 2016-09-22 at 10 14 28 pm

Why am I not allowed to push that button?  Am I not a "maintainer"?

glyph:
@wallrj - Out of date again. Can you merge again? I don't know why I can't :-\.

glyph:
All right.  LGTM, thanks for the contribution, and thanks as always for your patience with all the delays :)

wallrj:
Thanks @glyph !

PR #674, opened 2016-09-14
Nodes batch-delete is limited to 10

### LINKED ISSUE #673: CLB bulk-delete should limit to 10 nodes
CLB bulk-delete nodes [doc](https://developer.rackspace.com/docs/cloud-load-balancers/v1/api-reference/nodes/#bulk-delete-nodes) mention that only 10 nodes are allowed in a single call. This API returns 400 if > 10 nodes are provided. Mimic should provide such behavior.

manishtomar:
Fixes #673

### DIFF
--- a/mimic/model/clb_errors.py
+++ b/mimic/model/clb_errors.py
@@ -128,6 +128,52 @@ def loadbalancer_not_found():
     )
 
 
+def invalid_node_ids_error(node_ids):
+    """
+    Return validation error string given when node_ids are not part of CLB
+
+    :param ``list`` node_ids: List of int node_id
+    :rtype: ``str``
+    """
+    nodes = ','.join(map(str, sorted(node_ids)))
+    return "Node ids {0} are not a part of your loadbalancer".format(nodes)
+
+
+def batch_delete_limit_error(nodes_requested, max_limit):
+    """
+    Return validation error string given when nodes > limit are deleted
+    in a single call.
+
+    :param int nodes_requested: Number of nodes requested to be deleted
+    :param int max_limit: Maximum number of nodes allowed to be deleted
+    :rtype: ``str``
+    """
+    fmt = (
+        "Request to delete {} nodes exceeds the account limit BATCH_DELETE_LIMIT "
+        "of {} please attempt to delete fewer then 10 nodes")
+    return fmt.format(nodes_requested, max_limit)
+
+
+def validation_errors(errors):
+    """
+    Verified 2016-09-13 when trying to bulk-delete > 10 nodes
+
+    :param list errors: List of error strings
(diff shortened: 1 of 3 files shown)


mimic/model/clb_errors.py
```diff
@@ -128,6 +128,52 @@ def loadbalancer_not_found():
     )
 
 
+def invalid_node_ids_error(node_ids):
+    """
+    Return validation error string given when node_ids are not part of CLB
+
+    :param ``list`` node_ids: List of int node_id
```
glyph: I am pretty sure that putting markup like that into the middle of a `:param:` is a ReST syntax error.  It should just be `:param list node_ids:`.

mimic/model/clb_errors.py
```diff
@@ -128,6 +128,52 @@ def loadbalancer_not_found():
...
+    """
+    fmt = (
+        "Request to delete {} nodes exceeds the account limit BATCH_DELETE_LIMIT "
+        "of {} please attempt to delete fewer then 10 nodes")
+    return fmt.format(nodes_requested, max_limit)
+
+
+def validation_errors(errors):
+    """
+    Verified 2016-09-13 when trying to bulk-delete > 10 nodes
```
glyph: There's no documentation of this anywhere?

mimic/model/clb_objects.py
```diff
@@ -212,6 +215,9 @@ class RegionalCLBCollection(object):
                          validator=attr.validators.instance_of(int))
     lbs = attr.ib(default=attr.Factory(dict))
     meta = attr.ib(default=attr.Factory(dict))
+    # Current default limit for batch delete is 10 as per
+    # https://developer.rackspace.com/docs/cloud-load-balancers/v1/api-reference/nodes/#bulk-delete-nodes
```
glyph: Aah. Here we go.  This link should be in a docstring rather than a comment, ideally; perhaps an `:ivar:` on the class?

mimic/model/clb_objects.py
```diff
@@ -490,6 +496,10 @@ def delete_nodes(self, lb_id, node_ids):
         if lb_id not in self.lbs:
             return not_found_response("loadbalancer"), 404
 
+        if len(node_ids) > self.batch_delete_limit:
```
glyph: When you verified this, were you trying to delete 10 _valid_ node IDs or were you just deleting 10 random IDs?  My suspicion is that it validates the IDs before checking the length, but I just wanted to check :).

manishtomar: You are absolutely right. It validates the IDs before checking for limit. I'll do the same thing here.

manishtomar:
@glyph I've addressed your review. Not sure how to mark it such.

PR #675, opened 2016-09-15
Create node atom feed

manishtomar:
An atom feed is added when CLB node is created. Unfortunately, there is no documentation about the format of the feed apart from an email from CLB team.

### DIFF
--- a/mimic/model/clb_objects.py
+++ b/mimic/model/clb_objects.py
@@ -64,7 +64,7 @@ class Node(object):
     status = attr.ib(validator=one_of_validator("ONLINE", "OFFLINE"),
                      default="ONLINE")
     # List of (xml text, updated seconds since epoch) tuple
-    feed_events = attr.ib(default=[])
+    feed_events = attr.ib(default=attr.Factory(lambda: []))
 
     @classmethod
     def from_json(cls, json_blob):
@@ -572,7 +572,7 @@ class RegionalCLBCollection(object):
                 for node in nodes:
                     node.status = "OFFLINE"
 
-            self._add_node_created_feeds(self.lbs[lb_id], nodes)
+            self._add_node_created_feeds(nodes)
 
             self._verify_and_update_lb_state(
                 lb_id, current_timestamp=current_timestamp)
@@ -581,6 +581,9 @@ class RegionalCLBCollection(object):
         return not_found_response("loadbalancer"), 404
 
     def _add_node_created_feeds(self, nodes):
+        """
+        Add "Node created..." feed for each given nodes
+        """
         created_feed = (
             "Node successfully created with address: '{address}', port: '{port}', "
             "condition: '{condition}', weight: '{weight}'")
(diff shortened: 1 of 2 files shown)


mimic/test/test_loadbalancer.py
```diff
@@ -812,6 +812,32 @@ def test_add_multiple_nodes(self):
...
+            ("<feed xmlns=\"http://www.w3.org/2005/Atom\"><entry>"
+             "<summary>Node successfully created with address: '1.1.1.1', "
+             "port: '80', condition: 'ENABLED', weight: '10'</summary>"
+             "<updated>1970-01-01T00:00:00.000000Z</updated></entry></feed>"),
+            ("<feed xmlns=\"http://www.w3.org/2005/Atom\"><entry>"
+             "<summary>Node successfully created with address: '1.1.1.2', "
+             "port: '80', condition: 'ENABLED', weight: '10'</summary>"
+             "<updated>1970-01-01T00:00:00.000000Z</updated></entry></feed>")
+        ]
+        for node_id, exp_node_feed in zip(node_ids, node_feeds):
```
glyph: I'd really like to get away from using loops in tests.  Could you express this as a few repeated calls to a function, instead, so that the traceback encodes information about what case is failing, by the line number that has failed?

mimic/model/clb_objects.py
```diff
@@ -66,7 +66,8 @@ class Node(object):
                  default=attr.Factory(lambda: randrange(999999)))
     status = attr.ib(validator=one_of_validator("ONLINE", "OFFLINE"),
                      default="ONLINE")
-    feed_events = attr.ib(default=[])
+    # List of (xml text, updated seconds since epoch) tuple
```
glyph: Can you make this an `:ivar:` docstring field rather than a comment?

mimic/model/clb_objects.py
```diff
@@ -66,7 +66,8 @@ class Node(object):
...
+    feed_events = attr.ib(default=attr.Factory(lambda: []))
```
glyph: You can just do `Factory(list)` instead of writing your own `lambda` - a few fewer characters :).

mimic/model/clb_objects.py
```diff
@@ -578,12 +579,26 @@ def add_node(self, node_list, lb_id):
...
+
             self._verify_and_update_lb_state(
                 lb_id, current_timestamp=current_timestamp)
             return {"nodes": [node.as_json() for node in nodes]}, 202
 
         return not_found_response("loadbalancer"), 404
 
+    def _add_node_created_feeds(self, nodes):
+        """
+        Add "Node created..." feed for each given nodes
```
glyph: Should this document the fact that there was a private email explaining this format, perhaps, since there's no public doc for it?

Also, `nodes` could use a `:param:` :).

mimic/model/clb_objects.py
manishtomar: Thanks! I realized after putting up the PR :)

manishtomar:
@glyph Addressed comments. Ok to merge this?

glyph:
Looks good!

PR #676, opened 2016-11-15
fix dependencies from requires.io PR

glyph:


glyph:
I'm going to unilaterally merge this without pre-commit review since we are really _already_ in a broken state (`pip install` doesn't work with recent versions of everything), and this doesn't change any implementation code, just build configuration.

PR #678, opened 2016-11-16
don't depend on implicit behavior, removed in recent twisted

glyph:
Fix the Dockerfile to actually *install* the software.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,10 +3,8 @@ FROM python:2.7
 RUN mkdir -p /usr/src/app
 WORKDIR /usr/src/app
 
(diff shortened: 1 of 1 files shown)


Kuwagata:
👍  Confirmed that the container starts up successfully.

glyph:
Thanks for reporting and testing @Kuwagata !  I'll hit the green button as soon as Travis is done.

everett-toews:
👍

glyph:
Thanks for the second look @everett-toews !  I was just waiting for CI to pass (Travis has been having a hard week) but it's always nice to get more reviewers :).

PR #700, opened 2016-12-13

ryandub:


glyph:
I've reproduced the failure locally; it seems modulegraph just can't see zope.interface 4.3.3 for some reason.

glyph:
Reported against zope here https://github.com/zopefoundation/zope.interface/issues/67 and py2app here https://bitbucket.org/ronaldoussoren/py2app/issues/206/py2app-doesnt-work-with-release-433-of

PR #721, opened 2017-01-13
Mimic specific vimrc file

manishtomar:
which comes into effect when opening any file in this directory. Currently, it only colors 105 column.

### DIFF
--- a/_vimrc_local.vim
+++ b/_vimrc_local.vim
@@ -1,0 +1,3 @@
+" Directory specific vim config file. Requires https://github.com/LucHermitte/local_vimrc plugin installed
+" Mark column at 105 as that is limit set in tox.ini
+set colorcolumn=105  


glyph:
I understand the desire for automatic project-specific configuration, but I really do not want to encourage the use of a _horribly_ insecure plugin to achieve this.  (`vim` should not execute code from the current directory).

glyph:
I feel a little bad closing this PR, since I agree the metadata is useful, but unless there's some way to make this type of execution safe, I feel like it would be irresponsible to add it to the project.  So for now at least, closing.

manishtomar:
Hmm. Since this file is only executed by the plugin wouldn't the developer have made the decision about its security before this file is actually executed?

glyph:
@manishtomar My point isn't that Mimic might somehow subvert their security, but rather that I don't want to encourage this plugin's use, because it is doubtful that the developer has thoughtfully evaluated it.  For example, would you expect `cd Downloads/some-weird-thing; vi textfile.txt` to execute arbitrary code?

manishtomar:
> For example, would you expect cd Downloads/some-weird-thing; vi textfile.txt to execute arbitrary code?

Most definitely not. After [browsing](http://usevim.com/2012/03/28/modelines/) [about](http://stackoverflow.com/a/3164830/499808) this I am now concerned about the plugins I've installed 😕  They aren't much but I'll check them. Thanks for the perspective.

PR #734, opened 2017-02-11

stavxyz:


glyph:
This PR is failing, and will continue to fail, until someone can switch the mimic testing stuff over to use `treq.testing`, which has since been updated to work with new versions of Twisted.  Since treq.testing grew out of this work, the API ought to be mostly compatible.

